PR #58, opened 2020-06-03
Download restart

liuzheng-arctic:
I like the idea to save orderIDs more frequently. Everything looks fine. I see that the recent change here brought back the email "bombing" again. Having the emails providing all the download and processing information is nice, at least for people with less experience with IS2 data. I am getting used to the no-email ordering experience, especially because icepyx took care of all the data access workflow so fluently, the emails do not help too much any more. Is it possible to control the number of emails sent or just an on/off switch? For a 5 order request, I am already getting 10 emails (5 for processing 5 for complete/fail). For a larger request, it can be annoying. Maybe consider adding an option to reduce/remove email notifications?


asteiker:
@liuzheng-arctic Yes there is an option in the NSIDC API to disable email. If the email keyword is not set, no email will be returned. The email can also be set without specifying the email address itself - this is automatically determined on the back end from Earthdata Login. These are all acceptable key-value-pairs:

`email=true`
`email=yes`
`email=name@domain.com`

JessicaS11:
I've merged this PR and will see what I can do to make the email optional in a new branch (I've got a few other pre-HW small fixes to make too).

liuzheng-arctic:
Thanks for the clarification. Can we use email=False or email='no' to turn
it off?
If such switch is supported by NSIDC API, then we should take advantage of
that in icepyx as well.

Zheng

JessicaS11:
@asteiker Does NSIDC have a preference as far as the default behavior (to send emails or not)? I'll use whatever your preference is as the default and have added code to make it easier to change this setting.

Issue #35, opened 2020-03-11
functions to pass data to other formats


JessicaS11:
One of icepyx's primary goals is to make it easier to work with ICESat-2 data. The present class object data structure aims to facilitate switching between multiple data structures to capitalize on other resources for data analysis. Thus, we need a set of functions for the class (e.g. to_dataframe, to_geodataframe, to_dict, to_xarray, to_netcdf, to_hdf5) that enable the user to easily put their data in these formats for further analysis. How this proceeds is part of ongoing conversations about development directions, so please contribute to the conversation here or on Discourse even if you don't think you'll actually work on this issue directly.

weiji14:
Hi @JessicaS11, pretty sure that most of the [`to_*`](http://xarray.pydata.org/en/v0.15.1/api.html#dataset-methods) functions you mentioned are available if you load the HDF5 files into an [xarray.Dataset](http://xarray.pydata.org/en/v0.15.1/data-structures.html#dataset), except for `to_geodataframe`. Was that what you had in mind? Plus you get all of the data variables loaded nicely, see the example code/screenshots below:

```python
# ...
order_id: str = region_a.granules.orderIDs[0]
download_path: str = f"icepyx_data/{order_id}"
region_a.download_granules(path=download_path)
```

```python
ds: xr.Dataset = xr.open_mfdataset(
    paths=f"{download_path}/**/processed_*h5",
    engine="h5netcdf",
    group="gt2l/land_ice_segments",
    combine="by_coords",
)
ds
```

![xarray loading ICESat-2 HDF5 file](https://user-images.githubusercontent.com/23487320/83350596-3a2f8680-a391-11ea-8b6c-3684fa5fd08d.png)

![xarray.Dataset to_ methods](https://user-images.githubusercontent.com/23487320/83350658-b32ede00-a391-11ea-8769-e82b0dbf9fef.png)

I'd be keen to make this part of `icepyx`, just need to know which part of the library I should put it in :smile:

JessicaS11:
Excellent @weiji14! We'd be happy to have you join the team and help make some of these things happen. I agree that leveraging existing libraries is the way to go wherever possible. The purpose of icepyx is to try and make it easier to go from data acquisition to final data products in a single workspace (versus, e.g., a download API, data processing API, then separate figure making API) and for scientists who are familiar with one data format but not necessarily transitioning between multiple formats (or using those that are cloud optimized).

I think I see conversions to other data formats happening as part of a two step process:
[...]

PR #59, opened 2020-06-04
Get filepaths to downloaded hdf5 granule files

weiji14:
Before [h5py](https://github.com/h5py/h5py), [h5netcdf](https://github.com/shoyer/h5netcdf), [xarray](https://github.com/pydata/xarray), and other readers can open the HDF5 files are, we'll need to figure out where the filepaths to actually are! This is the first step in resolving #35. Users can access a `filepaths` property in the `Icesat2Data.granules` class after they've downloaded the granules (or they can set the `Icesat2data.granules.filepaths` property if they've downloaded the files already).

```python
region_a = icepyx.icesat2data.Icesat2Data(
    dataset="ATL06",
    spatial_extent=[-55, 68, -48, 71],
    date_range=["2019-02-20", "2019-02-28"],
    version="003",
)
region_a.earthdata_login(uid=uid, email=email)

region_a.download_granules(path="icepyx_data/")
filepaths = region_a.granules.filepaths
print(filepaths)
```

gives:

```python
['icepyx_data/5000000123456/processed_ATL06_20190221121851_08410203_003_01.h5',
 'icepyx_data/5000000123456/processed_ATL06_20190225121032_09020203_003_01.h5',
 'icepyx_data/5000000123456/processed_ATL06_20190222010344_08490205_003_01.h5',
 'icepyx_data/5000000123456/processed_ATL06_20190226005526_09100205_003_01.h5']
```

Users can then load the data files from the filepaths, e.g. as follows:

```python
# h5py
import h5py

for file in filepaths:
    h5file = h5py.File(file, mode="r")
    gt2l = h5file["gt2l"]["land_ice_segments"]

# xarray
import xarray as xr

gt2l: xr.Dataset = xr.open_mfdataset(
    paths=filepaths,
    engine="h5netcdf",
    group="gt2l/land_ice_segments",
    combine="by_coords",
)
```

I can write unit tests for this ~~if someone finds a way to securely authenticate to the NSIDC servers without putting my uid/password up :laughing:~~ Done: Unit tests written!


weiji14:
> I think it would be better to have `filepaths` be a property of the granules class that is filled in during the download process (if a download occurs) 

Ok done, but I'll probably need to refactor it later to account for the items below :point_down:

>..., but I have concerns about how modular/flexible the filepath list would be given that it is explicitly tied to the download process (and thus cannot be used for locally available files that have already been downloaded).
> ...
> This also removes the need for a login/authentication in order to write a unit test (unit tests for new functions are encouraged and would be super appreciated - thanks for offering!

I'll need to have a rethink about how to do this. I see you've added some new tests for `granules.py` (which is great!), and there was a line `obs_grans = [gran['producer_granule_id'] for gran in reg_a.granules]` that seems to get a list of would-be downloaded files (though I don't see how reg_a.granules is iterable). Would that be the way to go? I.e. get the filepaths before download starts, and test for that. Will need set a `reg_a.granules.download_path` property perhaps.

> As an aside, thinking about how to unit test things has done wonders for my code-writing, given the authentication issue you note. Our current approach is to have as many functions as possible be independent of a login, and then we test the login independently using a secure key in Travis CI).

Wait until you start writing integration tests :laughing: But yes, good to have unit tests be as independent as possible.

weiji14:
Sorry, must have missed that branch! Probably should refactor and port the unit tests here over to there, but I'll take a closer look first and probably come to the next meeting (if my timezone works out) to see where things are at for icepyx.

JessicaS11:
@weiji14 Browsing through some of these older/open PRs (I was remembering that this one was waiting for read-in functionality consideration, which #222 does). Updated thoughts on this (and if we want to pursue any of it, could you update the base so we can see where the differences are more easily)?

weiji14:
Closing this as it's not really relevant anymore, especially with the new `icepyx.Read` module at https://github.com/icesat2py/icepyx/blob/v0.6.1/icepyx/core/read.py

Issue #38, opened 2020-03-13
speed up Travis builds


JessicaS11:
How can we speed up our test builds, e.g. through [parallelization](https://docs.travis-ci.com/user/speeding-up-the-build/) and [caching](https://docs.travis-ci.com/user/caching/)?

PR #61, opened 2020-06-04
Fix test failures in test_granules.py

weiji14:
Pytest automatically tests any files starting with the name `test_`, in folders under `tests`, see https://docs.pytest.org/en/latest/example/pythoncollection.html. This means we don't need to explicitly add new tests to the `.travis.yml` file everytime.

TODO:
- [x] Let travis test everything
- [ ] Fix failing tests in test_granule.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ stages:
 jobs:
   include:
     - stage: basic tests
-      script: pytest icepyx/
+      script: pytest icepyx/ --ignore icepyx/tests/behind_NSIDC_API_login.py
 
     # - stage: behind Earthdata
     #   script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,9 @@ stages:
 jobs:
   include:
     - stage: basic tests
-      script: pytest icepyx/tests/test_APIformatting.py
-              icepyx/tests/test_geospatial.py
-              icepyx/tests/test_is2ref.py
-              icepyx/tests/test_validate_inputs.py
-        
+      script: pytest icepyx/
+
     # - stage: behind Earthdata
-    #   script: 
+    #   script:
     #     - export NSIDC_LOGIN=$NSIDC_LOGIN
     #     - pytest icepyx/tests/behind_NSIDC_API_login.py
--- a/icepyx/tests/test_granules.py
+++ b/icepyx/tests/test_granules.py
@@ -22,7 +22,7 @@ from icepyx.core.granules import Granules as Granules
 def test_granules_info():
     # reg_a = ipd.Icesat2Data('ATL06', [-55, 68, -48, 71], ['2019-02-20','2019-02-24'], version='3')
     # granules = reg_a.granules.avail
-    granules = [{'producer_granule_id': 'ATL06_20190221121851_08410203_003_01.h5',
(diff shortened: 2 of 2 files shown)


weiji14:
This is the remaining test failure:

```python-traceback
=================================== FAILURES ===================================
______________________ test_correct_granule_list_returned ______________________

    def test_correct_granule_list_returned():
        reg_a = ipd.Icesat2Data('ATL06',[-55, 68, -48, 71],['2019-02-20','2019-02-28'], version='2')
        reg_a.avail_granules()
>       obs_grans = [gran['producer_granule_id'] for gran in reg_a.granules]
E       TypeError: 'Granules' object is not iterable

icepyx/tests/test_granules.py:407: TypeError```

weiji14:
Ok, found out that the `test_correct_granule_list_returned` should be using `reg_a.granules.avail` instead of `reg_a.granules`. Should be all good now!

weiji14:
Ok, I've reverted the change (and changed the title too). But **before** you merge this, what I can do is to set travis up so that the first stage tests everything except `tests/behind_NSIDC_API_login.py` using the `--ignore` option in pytest as documented at https://docs.pytest.org/en/latest/example/pythoncollection.html#ignore-paths-during-test-collection. If that's what all the second stage is about (testing functions that require a login)?

Alternatively, we could put all the 'second stage' tests in a folder like `tests/login_tests`, and have `pytest` ignore those ones in the first stage, and test those ones in the second stage. Or you could set up a test matrix so that both of the stages are run concurrently (which might help with #38).

As an aside, I found https://automationpanda.com/2017/03/06/python-testing-101-introduction/ to be a good resource for learning about testing when I first started (maybe check it out after the Hackweek), the `pytest` specific one at https://automationpanda.com/2017/03/06/python-testing-101-introduction/ might be a good one to focus on :smile:

JessicaS11:
@weiji14 I am fine with using the pytest `--ignore` option for the first suite of tests (basic tests stage). One of the issues I think we will keep running into in the short term is that the test suite is not yet completed (and has only been partially brought over from a previous version of the code, which is why not all of the test files start with "test" - that's the indicator that they still need updating!). My thought was that once we have a more complete set of tests it will be easier to determine which of them should be put into test matrices (as per #38 --> I intended for this to apply to the basic tests stage, which is currently what takes the bulk of the test time).

Thanks for those resources - I look forward to checking them out when this crazy week is over!

weiji14:
Cool, I've used `--ignore` to skip the `behind_NSIDC_API_login.py`. I've only just noticed it didn't start with `test_`, which means they wouldn't actually be tested :astonished:! It should be fine with the short term (not testing everything), but once we get into refactoring code, then it would be best to have a proper test suite.

Great job on the Hackweek by the way, I've just been going over a few of the youtube videos and I can only imagine all the work going behind them to make everything go so smoothly! Definitely get some rest after, it's well deserved :grin:

JessicaS11:
Yup - taking "test" out of the filename was very intentional until I got to updating it! Thanks for these changes and the kudos. The Hackweek went quite well, but I'm still catching up on everything that happened during it (or I missed because of it). : )

JessicaS11:
@weiji14 Don't forget to add yourself as a contributor on your next PR!

PR #63, opened 2020-06-07
Unzip hdf5 files into one directory instead of many

### LINKED ISSUE #33: downloaded file unzipping and sorting
Data files downloaded from NSIDC come in a series of directories and subdirectories with long names. Amy's tutorial included some code for managing this structure and simplifying it to put all of the data files in one place. This code has been copied into the icepyx class object, but is not yet implemented (currently commented out). It would be great if this functionality were available, either as a step in the download process (where it currently is) or as a standalone method for the data class (probably the more flexible option).

weiji14:
The downloaded zip file stores each HDF5 file in separate subfolders. This Pull Request fixes #33 by modifying the unzipping code so that the files are all in a single folder (the download folder set by the user). I.e. instead of having this:

- download_folder
  - 123456789
    - processed_ATL06_2019*.h5
  - 234567891
    - processed_ATL06_2019*.h5
  - 345678912
    - processed_ATL06_2019*.h5
  - 456789123
    - processed_ATL06_2019*.h5

The unzip will now do this:

- download_folder
    - processed_ATL06_2019*.h5
    - processed_ATL06_2019*.h5
    - processed_ATL06_2019*.h5
    - processed_ATL06_2019*.h5

There seems to be an `extract` argument (which takes a boolean True/False) that is commented out in the code, should I enable it?

### DIFF
--- a/icepyx/core/granules.py
+++ b/icepyx/core/granules.py
@@ -379,7 +379,10 @@ class Granules():
         #         #Note: extract the dataset to save it locally
         # if extract is True:    
             with zipfile.ZipFile(io.BytesIO(zip_response.content)) as z:
-                z.extractall(path)
+                for zfile in z.filelist:
+                    # Remove the subfolder name from the filepath
+                    zfile.filename = os.path.basename(zfile.filename)
+                    z.extract(member=zfile, path=path)
             
             # update the current finished order id and save to file
             with open(downid_fn,'w') as fid:


JessicaS11:
@weiji14 The extract keyword is commented out because it was not yet fully implemented. It would be great to have that functionality for those who do not want to automatically unzip their downloaded files (especially for very large datasets). I think it would be best to keep/set the default behavior as `extract=True` though.

weiji14:
Yes, especially since there are ways to read the data inside the zipfile without explicitly unzipping them. Leave that for a separate Pull Request.

PR #71, opened 2020-06-15
Add authentication with netrc

### LINKED ISSUE #69: Enhancement: Add netrc for authentication
Currently, authentication is handled via manual user input. 

https://github.com/icesat2py/icepyx/blob/master/icepyx/core/Earthdata.py#L88 which uses https://docs.python.org/2/library/getpass.html

Suggest enhancement to use `netrc` library to read and pass credentials. This will enable programmatic authentication.

friedrichknuth:
This PR adds authentication via `netrc`.

To use this functionality, create a .netrc file in your $HOME directory with the following lines:
```
machine urs.earthdata.nasa.gov
login your_Earthdata_user_ID
password your_password
```
Then change the permissions of that file to 600 with `chmod 600 ~/.netrc`.

The function checks for this file to retrieve credentials, prior to prompting for manual input. Authentication must still be instantiated with `earthdata_login(earthdata_uid, email)`, but the password is pulled programmatically from the `.netrc` file.

closes #69

### DIFF
--- a/icepyx/core/Earthdata.py
+++ b/icepyx/core/Earthdata.py
@@ -1,6 +1,7 @@
 import requests
 import getpass
 import socket
+import netrc
 import re
 import json
 
@@ -25,7 +26,7 @@ class Earthdata():
     -------
     Earthdata session object after a successful login
     """
-        
+
     def __init__(
         self,
         uid,
@@ -33,10 +34,11 @@ class Earthdata():
         capability_url,
         pswd=None,
     ):
-        
+
         assert isinstance(uid, str), "Enter your login user id as a string"
         assert re.match(r'[^@]+@[^@]+\.[^@]+',email), "Enter a properly formatted email address"
-        
+
+        self.netrc = None
         self.uid = uid
         self.email = email
         self.capability_url = capability_url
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -6,3 +6,4 @@ matplotlib
 numpy
 requests
 shapely
+netrc
```
wallinb, line 9: This isn't necessary is it? 'netrc' is in the standard library since Python 1.5

tsutterley: @wallinb you're right i removed the netrc requirement.  i also rebased into a single commit and am using urs.earthdata.nasa.gov following the NSIDC programmatic guide

tsutterley:
looks good to me.  thanks @friedrichknuth !

JessicaS11:
Thanks @friedrichknuth, @wallinb, and @tsutterley! This looks great. @tsutterley, are you able to officially review and select approve? If you select "files changed", there should be a drop-down in the top right that gives you the option to.

JessicaS11:
I'm having trouble getting the Travis-CI check to complete to merge this PR. I resolved the changes requested, but I'm wondering if @wallinb needs to approve the changes too (since you're listed as a reviewer?) to trigger the build.

JessicaS11:
Thanks @wallinb. It still does not seem to have triggered a test run. My only other idea (if it's not a side effect of the force-push) is that the master branch is protected, so you can only push to it from a non-protected branch within the same repo. Thus, it's not triggering the merge build because it doesn't allow a direct push from a fork to master. This is the expected behavior (given how the repo is set up), but I haven't actually seen any PRs from a forked repo before.

wallinb:
@JessicaS11 Strange, I would expect the CI to run regardless of protections, or at least not with this outcome. You have probably seen this?: https://travis-ci.community/t/on-a-pr-some-checks-havent-completed-yet-waiting-for-status-to-be-reported/7951

Issue #22, opened 2020-02-19
interact with NSIDC API without paging


JessicaS11:
Paging is currently used to parse through granule search results, but it is not the recommended way to interact with the API. The Icesat2Data object and its methods should be modified to not use paging.

wallinb:
@JessicaS11 I think you are referring to using 'scrolling' rather than paging with CMR correct? I'd be happy to take this on.

JessicaS11:
@wallinb - yes! That sounds right, and it would be great if you could make that happen. I'll add you to the organization so that you can create a branch for the work and push changes directly to the repo. By way of orientation: I suspect most of the changes will be in the granules module. We're also encouraging the addition of tests for any new code. Please keep in touch if you have any questions or want a more thorough intro. As always, suggestions and insights are welcome.

JessicaS11:
@wallinb left NSIDC in 2021 and this issue has languished. I picked up the open PR recently and in the process of updating it learned that the [current recommended approach](https://cmr.earthdata.nasa.gov/search/site/docs/search/api.html#search-after) is to use the `CMR-Search-After` header rather than `scrolling` (which was preferred over `paging`). Thus, #87 has been updated to the latest and greatest.

PR #87, opened 2020-06-18
Cmr scrolling search

wallinb:
Unfortunately EGI does not support scrolling yet, so this only updates granule search. 

Some thoughts: Ordering and querying could be less coupled than they are now e.g. `place_order` calls `get_avail` to know how many granules will be ordered, but the sub-setting service will do the similar search itself to fulfill the order. In some edge cases these two searches can become inconsistent with each other, especially now that one uses scrolling and one uses paging. At the moment it is the only way to know/estimate the number of pages/orders necessary so I left it as is. Maybe long term we want to separate 'querying' functionality from 'ordering' into separate classes or something?

### DIFF
--- a/icepyx/core/granules.py
+++ b/icepyx/core/granules.py
@@ -96,9 +96,7 @@ class Granules():
         icesat2data.Icesat2Data.avail_granules
         """
 
-        assert (
-            CMRparams is not None and reqparams is not None
-        ), "Missing required input parameter dictionaries"
+        assert (CMRparams is not None and reqparams is not None), "Missing required input parameter dictionaries"
 
         # if not hasattr(self, 'avail'):
         self.avail = []
@@ -106,21 +104,21 @@ class Granules():
         granule_search_url = 'https://cmr.earthdata.nasa.gov/search/granules'
 
         headers = {'Accept': 'application/json'}
+        #DevGoal: check the below request/response for errors and show them if they're there; then gather the results
+        #note we should also do this whenever we ping NSIDC-API - make a function to check for errors
         params = apifmt.combine_params(
             CMRparams, {k: reqparams[k] for k in ['page_size']}
         )
         params['scroll'] = 'true'
 
-        # DevGoal: check the below request/response for errors and show them if they're there; then gather the results
-        # note we should also do this whenever we ping NSIDC-API - make a function to check for errors
         cmr_scroll_id = None
         while True:
-            if cmr_scroll_id:
+            if cmr_scroll_id is not None:
                 headers['CMR-Scroll-Id'] = cmr_scroll_id
 
             response = requests.get(granule_search_url, headers=headers, params=params)
 
-            if not cmr_scroll_id:
+            if cmr_scroll_id is None:
                 hits = int(response.headers['CMR-Hits'])
 
             cmr_scroll_id = response.headers['CMR-Scroll-Id']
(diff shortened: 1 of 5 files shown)


icepyx/tests/test_granules.py
```diff
@@ -22,7 +22,7 @@
 def test_granules_info():
     # reg_a = ipd.Icesat2Data('ATL06', [-55, 68, -48, 71], ['2019-02-20','2019-02-24'], version='3')
     # granules = reg_a.granules.avail
-    granules = [{'producer_granule_id': 'ATL06_20190221121851_08410203_003_01.h5',
+    entries = [{'producer_granule_id': 'ATL06_20190221121851_08410203_003_01.h5',
```
wallinb, line 25: Just some fixes to these existing tests.

requirements.txt
```diff
@@ -1,4 +1,3 @@
-datetime
```
wallinb, line 1: datetime is in python stdlib, so no need to have in requirements.txt

icepyx/tests/test_is2class_query.py
```diff
@@ -16,17 +15,14 @@ def test_reqconfig_params():
...
     exp_keys_all = ['page_size','page_num']    
     assert all(keys in obs_keys for keys in exp_keys_all)
 
     #test for download params
-    reg_a.reqparams=None
-    reg_a.build_reqconfig_params('download')
     reg_a.reqparams.update({'token':'','email':''})
     obs_keys = reg_a.reqparams.keys()
-    exp_keys_all = ['page_size','page_num','request_mode','token','email','include_meta']
+    exp_keys_all = ['page_size','page_num','token','email']
```
JessicaS11, line 25: Does this mean that the 'request_mode' and 'include_meta' parameters are never needed? If scrolling can only be applied to data querying, wouldn't these parameters still be needed for ordering/downloading?

icepyx/core/granules.py
```diff
@@ -182,29 +192,26 @@ def place_order(self, CMRparams, reqparams, subsetparams, verbose,
...
-
-            #DevNote: earlier versions of the code used a file upload+post rather than putting the geometries
-            #into the parameter dictionaries. However, this wasn't working with shapefiles, but this more general
-            #solution does, so the geospatial parameters are included in the parameter dictionaries.
+        total_pages = int(np.ceil(len(self.avail) / reqparams['page_size']))
+        print('Total number of data order requests is ',total_pages, ' for ',len(self.avail), ' granules.')
+        for page_num in range(1, total_pages+1):
+  
+            print('Data request ',page_num,' of ',total_pages,' is submitting to NSIDC',)
+            request_params = apifmt.combine_params(CMRparams, reqparams, subsetparams)
```
JessicaS11, line 209: I think this line is redundant with line 200 and will result in unintentional subsetting if subset=False. Suggest removing.

JessicaS11: Remove line (and help @JessicaS11 learn how to use the GitHub PR review tools!)

JessicaS11 (dismissed):

icepyx/tests/test_is2class_query.py
wallinb: I am not sure actually -- these tests were failing and excluded in CI so I  made these changes that they would pass but didn't delve much deeper. A better thing to have done is take the time to understand the correct behavior and update the tests from that :blush: I'll take a closer look

icepyx/core/granules.py
wallinb: We'll have to learn together -- I predominantly use bitbucket at work so github is much less familiar to me!

icepyx/tests/test_is2class_query.py
JessicaS11: They were likely excluded because I hadn't gotten to updating them since the major code refactor, which would've affected how they were called moreso than what the expected values should be. Thanks for noticing that the function wasn't being tested!

JessicaS11:
@wallinb I'm trying to get through some of these older PRs. Any thoughts on reviving this effort to switch from paging to scrolling?

wallinb:
@JessicaS11 😬 Sorry this has languished! I'll sync it up and see what else is needed.

JessicaS11:
See https://github.com/icesat2py/icepyx/issues/22#issuecomment-1028392271 for an update on why the switch from cmr scrolling to cmr-search-after.

Note: the PR-linked RTDs build is failing because of a [recent pip release](https://github.com/readthedocs/readthedocs.org/issues/6554) combined with the fact that you can't wipe the build environment for PRs. A [branch build](https://icepyx.readthedocs.io/en/cmr-scrolling-search/) shows they pass with no issues.

JessicaS11:
@betolink Do you think you could review this PR (or recommend someone who can)?

betolink:
Hi @JessicaS11 I can take a look this week, maybe Andy would be another person to ask as well. One thing, `Search-After` is great for performance on the CMR side but it is not compatible with deep pagination.

i.e. this would be an anti-pattern for an initial search:
```python
reqparams = {'page_size': 2000, 'page_num': 3}
```
Are the pagination parameters exposed to the users or it's an internal Icepyx interface? As a user I don't really need to access say page 565, usually we just want to know how many granules we get back from CMR and then download them all. The PR Looks good!

JessicaS11:
> Are the pagination parameters exposed to the users or it's an internal Icepyx interface?

Primarily the latter. With this PR, `page_num` is not submitted during the CMR query step, instead using `Search_After` to collect available results. `page_num`IS used by icepyx during ordering (and download) to iterate through the orders, since to my understanding `Search_After` is not available outside of CMR queries. `reqparams`, and thus `page_num`, is not hidden to the user, but my guess is nobody really pays it any mind (and if they don't view `reqparams` after placing an order, they'll probably never see `page_num` since it's not a required parameter for the CMR query step).

In general, I tried to design icepyx so that users had 100% control over (and visibility of) any request/parameters being sent to an endpoint *if they want it* (and to help it map clearly to the CMR docs for programmatic data access), but don't actually have to be bothered with the parameter dictionaries and request syntax/parsing otherwise.

icepyx/tests/test_granules.py
```diff
@@ -628,5 +628,5 @@ def test_avail_granule_CMR_error():
     ermsg = "An error was returned from NSIDC in regards to your query: temporal start datetime is invalid: [badinput] is not a valid datetime."
     with pytest.raises(NsidcQueryError, match=re.escape(ermsg)):
         CMRparams = {"version": "003", "temporal": "badinput", "short_name": "ATL08"}
-        reqparams = {"page_size": 1, "page_num": 1}
+        reqparams = {"page_size": 1}
```
betolink, line 631: Is this correct? `page_size = 1`?

betolink (commented):
It's looking great, there won't be a noticeable performance difference for Icepyx but CMR will handle these big queries better than just using pagination. I'll wait for Andy to take a look before approving it. Great work!

icepyx/tests/test_granules.py
JessicaS11: Yeah - it could be any number, since it's testing for the bad temporal input message and otherwise just needs to be a valid request. Would you recommend using a different number instead?

betolink: Got it, @JessicaS11 I think you can merge the PR. Have you heard anything from Andy?

betolink (approved):
I think this PR is ready to be merged and tested in the dev branch. Looks like regular pagination and search-after will coexist and will be transparent to the users so that's great.

PR #90, opened 2020-06-18
Update codecov badge to point to 'development' branch

wallinb:


JessicaS11:
Do you know if there's a way to make it point to the "current" (or development if not master) branch? It would be ideal to have it be accurate for both master and development for times in between releases, but if it's a lot of trouble we can just point it at development.

wallinb:
I don't think there is a way to do it dynamically (at least not easily). 

Some ideas here, such as a pre-commit hook -- seems a bit ugly and maybe not developer friendly: https://stackoverflow.com/questions/18673694/referencing-current-branch-in-github-readme-md

Or the readme could to have both badges (main and development branches). But IMO this badge is mostly of interest to developers, so it makes sense to reference 'development'.

JessicaS11:
@wallinb Thanks for the link. I suspected that might be the case, so I think you're right that simply pointing it to the dev branch makes the most sense.

Issue #72, opened 2020-06-16
local file operations for subsetting and file conversion


tsutterley:
adding another layer of in-memory bytesIO objects for performing local operations on subsetted files from icepyx

- subsetting to valid data points using provided or calculated quality flags  
- converting to different file formats not available from NSIDC (such as [zarr](https://zarr.readthedocs.io/en/stable/))  

basic addition to https://github.com/icesat2py/icepyx/blob/master/icepyx/core/granules.py#L390 will be like this (with the file operations coming after):  
```
for zfile in z.filelist:
    # Remove the subfolder name from the filepath
    zfile.filename = os.path.basename(zfile.filename)
    fileID = io.BytesIO(z.read(zfile))
    fileID.seek(0)
    # open in-memory HDF5 file and perform operations
    with h5py.File(fileID,'r') as source:
```

fspaolo:
@tsutterley I haven't used bytesIO before. What would be the advantage over simply working on downloaded HDF5s (in the standard way)?

tsutterley:
@fspaolo mostly that the end user would only see the end result (the data they want in the form they want).  I think it would make using ICESat-2 data a little bit more approachable (if outputting the "valid" data in a simple form).  Could even split into 6 beam level datasets to work with captoolkit.

fspaolo:
Another use case we might want to think about is "what happens if the user finds out it needs another variable"? Will it have to download the data again (since we have not persisted the original files)?

tsutterley:
that is a good point.  With the slimmed down "valid" versions that certainly would be an issue.  But I guess if the users want to "level up" and get more variables that'd be their next step.  Having these slimmed down forms might also allow for some more exploratory uses of the data.  This certainly could be interesting (I am thinking of an animation from @joemacgregor that used an early NSIDC subsetting to look at glacier termini with ATM).  On the other end of the spectrum for cloud computing purposes, if outputting as zarr might not want to store 2 full copies of the data.

fspaolo:
I think this can also be mitigated by providing some guidance on the "most common variables" for standard use cases, and perhaps also suggesting potential additional variables for more complex use cases.

joemacgregor:
Here's a couple of the videos Tyler mentioned for Greenland termini:
All Greenland 1993-2016: https://www.youtube.com/watch?v=8o4DnXLIhxc
Just Helheim: https://www.youtube.com/watch?v=BZR4czv2Kag
[...]

PR #91, opened 2020-06-18
initial version of "seam-less" conversion routines to help address #72

tsutterley:


requirements.txt
```diff
@@ -2,6 +2,7 @@ datetime
 fiona
 geopandas
 h5py
+zarr
```
JessicaS11, line 5: Please put new libraries in alphabetical order!

icepyx/core/icesat2data.py
```diff
@@ -702,7 +702,7 @@ def download_granules(self, path, verbose=False, subset=True, restart=False, **k
             Spatial subsetting returns all data that are within the area of interest (but not complete
             granules. This eliminates false-positive granules returned by the metadata-level search)
         restart: boolean, default false
-            If previous download was terminated unexpectedly. Run again with restart set to True to continue. 
+            If previous download was terminated unexpectedly. Run again with restart set to True to continue.
         **kwargs : key-value pairs
```
JessicaS11, line 706: We need to add any new kwargs to the docs.

icepyx/core/icesat2data.py
```diff
@@ -727,21 +727,21 @@ def download_granules(self, path, verbose=False, subset=True, restart=False, **k
...
         if restart == True:
             pass
         else:
             if not hasattr(self._granules, 'orderIDs') or len(self._granules.orderIDs)==0: self.order_granules(verbose=verbose, subset=subset, **kwargs)
-    
-        self._granules.download(verbose, path, session=self._session, restart=restart)
-  
-   
+
+        self._granules.download(verbose, path, session=self._session, restart=restart, **kwargs)
```
JessicaS11, line 742: This implementation for `convert` will cause issues given the way the code is set up right now (which may point to a weakness in the existing code rather than this being a bad way to implement this functionality - it will need some more thought). Specifically, acceptable kwargs here (as noted in the docstring) are ONLY those that are accepted by the NSIDC subsetter. The APIformatting module relies on the proper set of passed kwargs to format the request properly.

icepyx/core/convert.py
```diff
@@ -0,0 +1,209 @@
+import os
+import re
+import h5py
+import zarr
+import itertools
+import numpy as np
+
+# PURPOSE: Wrapper function for converting HDF5 files to different formats
+def file_converter(hdf5_file, path, reformat):
```
JessicaS11, line 9: What is the benefit of having these as a set of standalone functions rather than turning them into a conversion class?

icepyx/core/convert.py
```diff
@@ -0,0 +1,209 @@
...
+    elif (PRD == 'ATL07'):
+        VARIABLE_PATH = ['sea_ice_segments','height_segment_id']
+    elif (PRD == 'ATL08'):
+        VARIABLE_PATH = ['land_segments','segment_id_beg']
+    elif (PRD == 'ATL10'):
+        VARIABLE_PATH = ['freeboard_beam_segments','delta_time']
+    elif (PRD == 'ATL12'):
+        VARIABLE_PATH = ['ssh_segments']['delta_time']
+    # create list of valid beams within the HDF5 file
+    beams = []
```
JessicaS11, line 113: Can we simplify this by using some of the existing variable module functionality? That might also make it easier to generalize.

JessicaS11:
Sorry for the delay in getting to this - I've been having trouble keeping up with the post-hackweek flurry of activity.

I think these additions are off to a great start, but I think we can do a lot to more seamlessly integrate them into the existing codebase. I think I've noted most of the key points for this in the review. I'm happy to help work on these, since they're pretty big changes that relate to some of the ways we're restructuring things beyond the scope of this PR.

tsutterley:
I closed the PR as you're right a lot more would need to be done before any of these changes are added.  Kick the can down the road.

fspaolo:
@tsutterley this is a very good addition. As @JessicaS11 said, we should think how to implement this key functionality to integrate well with the different (planed) modules within `icepyx`. For example, it should be equally accessible from `query` and from `data` (formerly `icesat2data`). See a prototype here:

PR #94, opened 2020-06-30
add OpenAltimetry quick plotting

### LINKED ISSUE #92: Add quick plots from OpenAltimetry
During 2020 ICESat-2 Hackweek we explored the potential of reading unfiltered ICESat-2 directly from OpenAltimetry into memory, without downloading files onto local disk. This can be used for quickly reading and visualizing data prior to download from NSIDC. We think it would be useful to embed the OpenAltimetry API functions developed during the hackweek into Icepyx for built-in visualization and quick data query.

icetianli:
**Enhancement to enable quick plotting via data requests from OpenAltimetry.** 
- Adds `OA_request()` and `generate_OA_request_params()` functions to `Icesat2Data()` class. 
- Can be called by setting `region.visualize_spatial_extent(OA_quickplot=True)` 
- Default is `OA_quickplot=False`, so this should not impact existing workflows.
- Returned geopandas GeoDataFrame can be accessed via `region.OA_data`
- Test added under icepyx/icepyx/test/test_OpenAltimetry.py
**Notes:** 
- OpenAltimetry API will not return data that exceeds bounds of 5 degrees in the lat or lon dimension.
- Setting a broad request, such as `date_range = ['2018-10-14','2020-04-04']` and `spatial_extent = [-75, -76.5, -74.5, -76]` will return a large amount of data, which may take 1-2 minutes to complete. See icepyx/icepyx/test/test_OpenAltimetry.py for smaller request example.
closes #92

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ matplotlib
 numpy
 requests
 shapely
+descartes
--- a/icepyx/core/icesat2data.py
+++ b/icepyx/core/icesat2data.py
@@ -5,8 +5,15 @@ import json
 import warnings
 import pprint
 import time
+import re
+from itertools import compress
+import pandas as pd
+import numpy as np
 import geopandas as gpd
+from shapely.geometry import Point
 import matplotlib.pyplot as plt
+import concurrent.futures
+from pyproj import CRS
 
 from icepyx.core.Earthdata import Earthdata
 import icepyx.core.APIformatting as apifmt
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -6,3 +6,4 @@ matplotlib
 numpy
 requests
 shapely
+descartes
```
JessicaS11, line 9: for future reference, please try to add new libraries in alphabetical order. It makes it easier to see if the one you want is here as the list grows.

weiji14 (commented):
Hi @icetianli, thanks again for showing us your OpenAltimetry demo during the [icepyx meeting](https://discourse.pangeo.io/t/icepyx-team-meetings/722)! Just have a couple of suggestions that might help to simplify the code. I think the static quickplot here is definitely a good start, and we can work together on adding interactive [hvplot](https://hvplot.holoviz.org) maps later on in a separate PR.

icetianli:
Thanks @JessicaS11  and @weiji14 for the advice on this PR! I am closing it for now and will resubmit changes after the branch transition from master to main is completed!

PR #95, opened 2020-07-02
Docs updates and add pypi build+upload to travis

JessicaS11:


weiji14:
Hi Jessica, can't help but notice that you added two Tandem X DEM files in https://github.com/icesat2py/icepyx/pull/95/commits/aa6b5fb49ef5635e0428290ed5ba2cc943fabab8 to the repository which are used for the 'ICESat-2_DEM_comparison_Colombia_working.ipynb' tutorial. However, they're rather big files (25.7MB and 8.65MB), and git isn't usually meant for storing data files like this, so I just wanted to make sure (before you make any new commits) that this was intentional?

PR #96, opened 2020-07-02
Black and flake8

JessicaS11:
Formatted code using black (and added a pre-commit hook to run black). Added a github action to run flake8 and annotate where code doesn't meet standards


weiji14:
> Unfortunately Black doesn't format docstrings and comments. At some point we might want to conform comments to the 88 line-width limit as well.

There's [`blackdoc`](https://github.com/keewis/blackdoc) which applies black to code in documentation.

JessicaS11:
After discovering the GitHub action to run flake8 was failing, I did some playing around and it should be working now.

@fspaolo I think you raise a few great points. There's a longstanding debate about [whether or not black should format docstrings ](https://github.com/psf/black/issues/144) in any way, and if so, how. I think for now that using the numpydoc format gives us some guidance, if not specifics on spaces and newlines...

@weiji14 Thanks for the info about `blackdoc` - definitely a good tool for us to consider adding!

fspaolo:
@JessicaS11 right, formatting docstrings usually comes with a lot of personal preference... this is one reason why Black is not imposing this (yet). I would say (as a personal rule) when writing docstrings just keep in mind Black's default 88 line length (since we are imposing that for the rest of the code).

Issue #97, opened 2020-07-06
nodename nor servname provided, or not known?


bradlipovsky:
Hello!

The following code works on the Hackweek Pangeo but not on my personal computer.

```
import os
import shutil
import h5py
import matplotlib.pyplot as plt
from icepyx import icesat2data as ipd

short_name = 'ATL06'
spatial_extent = [31.5, -70.56, 33.73, -69.29]
date_range = ['2020-03-30','2020-04-1']
region_a = ipd.Icesat2Data(short_name, spatial_extent, date_range)

earthdata_uid = 'therealbradlipovsky'
email = 'brad_lipovsky@fas.harvard.edu'
region_a.earthdata_login(earthdata_uid, email)
region_a.download_granules('/home/jovyan/data') 
```

When run on my personal computer, I get the following error.  Am I doing something incorrectly?

Thanks!

Brad

```
---------------------------------------------------------------------------
gaierror                                  Traceback (most recent call last)
~/Google Drive/Science/icepyx/icepyx/core/Earthdata.py in login(self)
    106             self.uid,_,self.pswd = netrc.netrc(self.netrc).authenticators(url)
--> 107             session = self._start_session()
    108 

~/Google Drive/Science/icepyx/icepyx/core/Earthdata.py in _start_session(self)
     50         hostname = socket.gethostname()
---> 51         ip = socket.gethostbyname(hostname)
     52 

gaierror: [Errno 8] nodename nor servname provided, or not known

During handling of the above exception, another exception occurred:

gaierror                                  Traceback (most recent call last)
<ipython-input-13-0400ac4c047b> in <module>
     12 earthdata_uid = 'therealbradlipovsky'
     13 email = 'brad_lipovsky@fas.harvard.edu'
---> 14 region_a.earthdata_login(earthdata_uid, email)
     15 region_a.download_granules('/home/jovyan/data')

~/Google Drive/Science/icepyx/icepyx/core/icesat2data.py in earthdata_login(self, uid, email)
    572 
    573         capability_url = f'https://n5eil02u.ecs.nsidc.org/egi/capabilities/{self.dataset}.{self._version}.xml'
--> 574         self._session = Earthdata(uid,email,capability_url).login()
    575         self._email = email
    576 

~/Google Drive/Science/icepyx/icepyx/core/Earthdata.py in login(self)
    111             for i in range(5):
    112                 try:
--> 113                     session = self._start_session()
    114                     break
    115                 except KeyError:

[...]

PR #98, opened 2020-07-06
add try/except for gaierror to address #97

tsutterley:
mac OS upgrades can lead to machines unable to get IP from hostname.  this is a fix to revert to using 'localhost' in case of a gaierror from socket.gethostbyname


JessicaS11:
fixes #97

PR #102, opened 2020-07-10
change Icesat2data class to Query class

JessicaS11:
As per discussion during the 2020 Hackweek, I've updated the primary class name from Icesat2Data to Query. The Icesat2Data class temporarily remains with a deprication warning. I modified the docs, uml diagram, and examples to use the new syntax.


icepyx/core/icesat2data.py
```diff
@@ -1,843 +1,11 @@
...
 
-        world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
-        f, ax = plt.subplots(1, figsize=(12, 6))
-        world.plot(ax=ax, facecolor="lightgray", edgecolor="gray")
-        geospatial.geodataframe(self.extent_type, self._spat_extent).plot(
-            ax=ax, color="#FF8C00", alpha=0.7
+        warnings.filterwarnings("always")
+        warnings.warn(
+            "Depricated. Please use icepyx.Query to create a download data object (all other functionality is the same)",
+            DepricationWarning,
```
fspaolo, line 10: Do you mean `DeprecationWarning`?

https://docs.python.org/3/library/exceptions.html#DeprecationWarning

Also, I've seen these warnings usually in all caps:

 "DEPRECATED. Please use ..."

JessicaS11: Yes, I do mean DeprecationWarning... apparently spelling is hard!

JessicaS11:
Not sure why black didn't catch that. I removed it manually anyway.

Good point - I've added a note to the meeting planning docs to discuss abbreviations with the dev team at our next meeting.

Issue #101, opened 2020-07-09
Feature request: Subset and download data by rgt


barcheck:
For some processing, it would be nice to be able to download all repeats of a single or a few rgts, without having to download anything else.

JessicaS11:
@asteiker @wallinb @nicholas-kotlinski @tsutterley Do you know if this is something that's possible within the NSIDC or CMR APIs? I know it's been a topic of discussion whether or not it could be implemented on the user (icepyx) end.

fspaolo:
Wouldn't this be a matter of just filtering the list of files by selecting the RGT number from the file name? If the user specifies an area, `icepyx` could query every granule that intersects that area and select only the specific RGTs (and download just that area or the complete orbits).

wallinb:
I think this is possible in the CMR/EGI queries too, though it might require using related orbit parameters instead of the RGT number directly, e.g. equator crossing longitude: [https://cmr.earthdata.nasa.gov/search/granules?short_name=ATL03&page_size=2000&page_num=1&equator_crossing_longitude=103.4,103.5](https://cmr.earthdata.nasa.gov/search/granules?short_name=ATL03&page_size=2000&page_num=1&equator_crossing_longitude=103.4,103.5)

tsutterley:
I think we could also use the orbit_number parameter to map from RGT.  That way we don't have to download a larger zip file of unwanted RGTs or figure out the mapping between RGT and crossing longitude.
```
cmr_orbit_number = RGT + np.arange(ncycles)*1387 + 201
```

For example:  
- Track 314 Cycle 1: https://cmr.earthdata.nasa.gov/search/granules?short_name=ATL03&page_size=2000&page_num=1&orbit_number=515&version=003
- Track 314 Cycle 2: https://cmr.earthdata.nasa.gov/search/granules?short_name=ATL03&page_size=2000&page_num=1&orbit_number=1902&version=003
- Track 314 Cycle 3: https://cmr.earthdata.nasa.gov/search/granules?short_name=ATL03&page_size=2000&page_num=1&orbit_number=3289&version=003

tsutterley:
[...]

PR #103, opened 2020-07-13
Add cycles and tracks keywords to address #101

tsutterley:
Added keywords `cycles` and `tracks` to be able to reduce a search query based on 91-day orbital cycle or RGT.  Uses the CMR `orbit_number` parameter to reduce the query.  Updated Jupyter notebook to show functionality and add notes.


tsutterley:
this did work locally :disappointed:

JessicaS11:
@tsutterley These are great! Are you suggesting that this no longer works though (I think the functions it uses are largely untested so far, though it does pass Travis - maybe we could work on adding a few tests with this PR? I'd be happy to help with that.)? I still need to check out the notebook changes and test the functionality on my local system...

tsutterley:
It didn't pass Travis initially.  I edited one of the tests to add the optional orbit_number to the CMR parameters.  Should be good to go at this point @JessicaS11

JessicaS11:
Sorry - apparently I forgot to press go on this review before!

icepyx/core/icesat2data.py
```diff
@@ -666,8 +745,40 @@ def avail_granules(self, ids=False):
...
+                # TRK: Reference Ground Track (RGT)
+                # CYCL: Orbital Cycle
+                # GRAN: Granule region (1-14)
+                # RL: Data Release
+                # VERS: Product Version
+                # AUX: Auxiliary flags
+                # SFX: Suffix (h5)
+                PRD,HEM,YY,MM,DD,HH,MN,SS,TRK,CYCL,GRAN,RL,VERS,AUX,SFX=rx.findall(i).pop()
+                cycle_list.append(CYCL)
+                track_list.append(TRK)
```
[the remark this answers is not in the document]
tsutterley: Great point.  These have been moved into the granules module.

icepyx/core/icesat2data.py
```diff
@@ -632,9 +708,9 @@ def earthdata_login(self, uid, email):
         self._email = email
 
     # DevGoal: check to make sure the see also bits of the docstrings work properly in RTD
-    def avail_granules(self, ids=False):
+    def avail_granules(self, ids=False, cycles=False, tracks=False):
```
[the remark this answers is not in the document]
tsutterley: keywords have been added to the docstring

icepyx/core/granules.py
```diff
@@ -43,6 +44,55 @@ def gran_IDs(grans):
...
+        # CYCL: Orbital Cycle
+        # GRAN: Granule region (1-14)
+        # RL: Data Release
+        # VERS: Product Version
+        # AUX: Auxiliary flags
+        # SFX: Suffix (h5)
+        PRD,HEM,YY,MM,DD,HH,MN,SS,TRK,CYCL,GRAN,RL,VERS,AUX,SFX = \
+            rx.findall(gran["producer_granule_id"]).pop()
+        gran_tracks.append(TRK)
+    return gran_tracks
```
JessicaS11, line 95: Best practices question (I have no idea what the correct answer is, if there even is one): it seems like `gran_IDs()`, `gran_cycles()`, and `gran_RGTs()` all do almost the same thing. Should we be thinking about turning them into one method? I'm just thinking that if you're querying enough of the archive you could potentially see performance improvements if you only iterate through the granule list once.

tsutterley: They do almost exactly the same thing.  A few commits ago they were a single function with kwargs.  I'll put it back that way :+1:

icepyx/core/query.py
```diff
@@ -252,11 +281,50 @@ def end_time(self):
...
+        >>> reg_a = icepyx.query.Query('ATL06',[-55, 68, -48, 71],['2019-02-20','2019-02-28'])
+        >>> reg_a.tracks
+        ['0841', '0849', '0902', '0910']
+        """
+        return sorted(set(self._tracks))
+
+    @property
+    def orbit_number(self):
+        """
+        Return the ICESat-2 CMR orbit number
```
JessicaS11, line 313: for the test case where all tracks/cycles are used, no orbit number(s) are returned. Is this the expected behavior?

icepyx/core/granules.py
JessicaS11: Cool - thanks for being patient with my learning curve!

icepyx/core/query.py
tsutterley: yes that's the expected behavior as it'd be an unnecessary CMR parameter (basically would run as icepyx is now)

PR #116, opened 2020-08-12
relative imports and import icepyx as ipx

### LINKED ISSUE #107: standardize import statements
set up icepyx to use ipx abbreviation and alias imports for things like query.Query

JessicaS11:
At our last dev meeting we agreed to standardize imports as `import icepyx as ipx` combined with the use of relative imports to avoid things like `ipx.query.Query()`. I've modified __init__.py and the examples to that effect. Closes #107 and should be merged after PR #115 (since it was branched from that PR).


icepyx/__init__.py
```diff
@@ -1 +1 @@
-from .core import query
+from icepyx.core.query import Query
```
weiji14, line 1: Wow, this was a lot easier than I thought! I still see some `from icepyx import is2class as ipd` code under the 'icepyx/tests' folder though, could you please change them to use the new 'import icepyx as ipx' syntax?

PR #143, opened 2020-09-14
pre-commit check for files over 5MB

JessicaS11:


lheagy (commented):
This is a good idea @JessicaS11! 

A couple thoughts and a question:
- is there a Contributor guide file? In many projects this would be in the root directory. (I didn't see it there). In either case, just making sure that the instructions for setting up the hooks are clear is important as if the developers don't have `pre-commit` installed, and if they haven't run `pre-commit install` in the repo directory, then this will not automatically be applied
- is there an extra step for the large file hook? I tested this with a 16.2MB file and was able to commit it on a temp branch (which is not the desired behaviour...)

JessicaS11:
@lheagy Better late than never - I added some basic instructions on using `pre-commit` to the contribution guidelines. In general the documentation needs a bit of an overhaul/update, including more details on this, but I'd like to get the pre-commit hooks into the package in the meantime. I'm thinking a good path forward will be to provide these pre-commit hooks for those who want to use them, but ultimately add some more actions to PRs to do some of this checking and provide the code to run them manually if need be.

There shouldn't be any extra steps required for the large file hook (it stopped me from committing a too-large file), so I'm not sure what the issue might be. I did see some posts ([e.g.](https://stackoverflow.com/questions/49912695/git-pre-and-post-commit-hooks-not-running)) suggesting that there can be path issues if you've set `core.hooksPath` in your `git config`. That could cause it to skip the `$GIT_DIR/hooks` default, which is what is relied upon here.

JessicaS11:
@RomiP @zachghiaccio Are either of you able to give this a quick look? I'd like to make sure the explanatory text I added to the contributor docs makes sense.

doc/source/contributing/contribution_guidelines.rst
```diff
@@ -73,7 +75,8 @@ General Guidelines
 
 Basic Steps to Contribute
 ^^^^^^^^^^^^^^^^^^^^^^^^^
-We encourage users to follow the `git pull request workflow <https://www.asmeurer.com/git-workflow/>`_. For more detailed steps, please see :ref:`How to Contribute<dev_guide_label>`.
+We encourage users to follow the `git pull request workflow <https://www.asmeurer.com/git-workflow/>`_.
+For more detailed steps, please see :ref:`How to Contribute<dev_guide_label>`.
```
zachghiaccio, line 79: For ReadtheDocs pages, is it best practice to place sentences on separate lines like this?

zachghiaccio (approved):
Other than a clarification question, looks good to me!

doc/source/contributing/contribution_guidelines.rst
JessicaS11: As with most things, putting each sentence on a new line for rendered text (One Sentence Per Line, or OSPL) as "best practice" is up for debate. It requires that whatever system you are ultimately using to render your text treats a single newline as a sentence break and has a separate new paragraph indicator (for markdown and rst, it's simply two consecutive newlines). Personally, I've found that for text in shared version controlled workflows (e.g. Jupyterbooks, documentation), it reduces the cognitive load of reviewing text changes if OSPL is followed. For instance, if one contributor automatically reformats text to follow PEP's 79 character limit and another reformats using the (also widely accepted) 99 character limit, you end up with a diff that - instead of highlighting actual text changes - is effectively only highlighting where the hard returns are (vs where they used to be). Or you end up with edited text that has really uneven line lengths because it's not being reformatted. Having spent time reviewing text in these situations where it becomes really unclear what is an actual text edit, I think starting from OSPL is a good practice to help minimize this, since at least some proportion of the sentences will be shorter than these character widths. icepyx hasn't adopted any approach explicitly, but I try to use natural sentence breaks (e.g. commas) and new sentences on newlines as a starting point. This was (obviously) not my original approach, so as I make edits to text, I'm trying to adopt the OSPL practice.

PR #144, opened 2020-09-14
Data visualization with OpenAltimetry API

### LINKED ISSUE #92: Add quick plots from OpenAltimetry

icetianli:
Follow the original [RP](https://github.com/icesat2py/icepyx/pull/94) and the discussion with @JessicaS11 and @friedrichknuth, further improvements have been made regarding visualization by data requests from OpenAltimetry API:

- Generalization of the code for different datasets available from OA API (https://openaltimetry.org/data/swagger-ui/). It now supports products ['ATL06','ATL07','ATL08','ATL10','ATL12','ATL13’]. ATL03 is not supported beacause OA API only supports ATL03 data request at single date with a spatial limitation of 1 degree at either lat/lon.

- Create a visualization module to host all relevant methods related to OA request and data visualisation. Three methods of visualization are available, which are: 1) interactive plot in holoview and datashader; 2) traditional matplotlib plot for geodataframe; 3) default spatial bounding box plot. This is still at the exploration stage and can be enhanced later. 

- Solve the issue mentioned in [RP](https://github.com/icesat2py/icepyx/pull/94) on a 5-degree spatial limit in the latitude or longitude dimension set by the OpenAltimetry API. OA API can now return data with any bounding box. OA request is the most time-consuming part, presumably related to the way how parameter lists generated in generate_OA_paras function.

- Test added under icepyx/icepyx/test/test_OpenAltimetry.py

closes #92

### DIFF
--- a/icepyx/core/visualization.py
+++ b/icepyx/core/visualization.py
@@ -1,0 +1,284 @@
+# utility modules
+import glob
+import os
+import sys
+import re
+import intake
+import requests
+import numpy as np
+import pandas as pd
+import geopandas as gpd
+from pyproj import CRS
+from shapely.geometry import Point
+from itertools import compress
+import concurrent.futures
+import matplotlib.pyplot as plt
+import holoviews as hv
+import dask.array as da
+import dask.dataframe as dd
+import hvplot.dask  # adds hvplot method to dask objects
+import datashader as ds
+from holoviews import opts
+from holoviews.operation.datashader import datashade, rasterize
+#from holoviews.element.tiles import EsriImager
+hv.extension('bokeh', "matplotlib") 
+
+import icepyx.core.geospatial as geospatial
+
+def filelist_latestcycle(filelist, cycle_list):
+    """
+    Only get filelist for lastest cycle if multiple cycles are requested
+    """
+    if len(cycle_list) >1:
+    # for multiple repeat cycles, only request data for lastest cycle
+        latest_cycle = max(cycle_list)
+        new_filelist = [f for f in filelist if int(f[-14:-12])==latest_cycle]
+        return new_filelist
+
+    else: # only one cycle
+        return filelist
+
+class Visualize:
+    
+    """
+    Visualize data extent and the elevation distribution by requesting OpenAltimetry API before downloading data from NSIDC
+    """
+    
+    def __init__(self, filelist, bbox, product):
+        
+        self.filelist = filelist # filelist requested from icepyx Query object
+        self.bbox = bbox # Input bounding box 
+        self.product = product # Input ICESat-2 product
+
+    def file_meta(self,bbox_i):
+        """
+        Obtain metadata from filename list with bounding box satistifying OA 5 degree threshold
+        Return:
+            para_lists
+            
+        ### NEED CHANGE HERE ###
+        """
+
+        para_lists = [] # list of parameters for API query
+
+        for fname in self.filelist:
+
+            if self.product in ['ATL07', 'ATL10']:
+                temp = intake.source.utils.reverse_format(
+                        format_string="{product:5d}-{HH:02d}_{datetime:%Y%m%d%H%M%S}_{rgt:04d}{cycle:02d}{region:02d}_{release:03d}_{version:02d}.h5",
+                        resolved_string=fname,
+                    )
+
+            else: 
+                temp = intake.source.utils.reverse_format(
+                        format_string="{product:5d}_{datetime:%Y%m%d%H%M%S}_{rgt:04d}{cycle:02d}{region:02d}_{release:03d}_{version:02d}.h5",
+                        resolved_string=fname,
+                    )
+
+            rgt = temp['rgt']
+            cycle = temp['cycle']
+            f_date = str(temp['datetime'].date())
+
+            paras = [rgt, f_date, cycle, bbox_i, self.product]
+            para_lists.append(paras)
+
+        return para_lists
+
+    def grid_bbox(self, binsize = 5):
+        """Split bounding box into smaller grids if latitude/longitude range exceed the default 5 degree limit of OpenAltimetry """
+
+        lonmin = self.bbox[0]
+        latmin = self.bbox[1]
+        lonmax = self.bbox[2]
+        latmax = self.bbox[3]
+
+        lonx = np.arange(lonmin, lonmax + binsize, binsize)
+        laty = np.arange(latmin, latmax + binsize, binsize) 
+        lonxv, latyv = np.meshgrid(lonx, laty)
+
+        ydim, xdim = lonxv.shape 
+
+        bbox_list = []
+        for i in np.arange(0,ydim-1): 
+            for j in np.arange(0,xdim-1):
+                # iterate grid for bounding box
+                bbox_ij = [lonxv[i,j], latyv[i,j], lonxv[i+1,j+1], latyv[i+1,j+1]]
+
+                if bbox_ij[2] > lonmax:
+                    bbox_ij[2] = lonmax
+
+                if bbox_ij[3] > latmax:
+                    bbox_ij[3] = latmax
+
+                bbox_list.append(bbox_ij)
+
+        return bbox_list
+
+    def generate_OA_paras(self):
+        """Generate parameter lists for OpenAltimetry API"""
+
+        # spatial extent check and grid bbox
+        lonmax = self.bbox[2]
+        lonmin = self.bbox[0]
+        latmax = self.bbox[3]
+        latmin = self.bbox[1]
+
+        # check if bounding box satisfies extent thresholds: |maxx - minx| < 5 and maxy - miny < 5 ###
+        extent_flag = ((lonmax - lonmin) > 5 ) or ((latmax - latmin) > 5 ) 
+
+        if extent_flag:
+            #print ('spatial extent exceeds OA 5 degree threshold, spliting bbox...')
+            bbox_list = self.grid_bbox()
+            paras = [self.file_meta(bbox_i) for bbox_i in bbox_list]
+            para_lists = [para_list for sublists in paras for para_list in sublists]
+        else:
+            #print ('spatial extent satifies OA 5 degree threshold, generating data...')
+            bbox_list = self.bbox
+            para_lists = self.file_meta(bbox_list)
+
+        return para_lists
+
+    def request_OA_elev(self, paralist):
+        """
+        Request data from OpenAltimetry based on API: https://openaltimetry.org/data/swagger-ui/#/
+        Return:
+            OA_array: numpy array for all beams of one RGT in paralist
+        """
+        base_url = 'https://openaltimetry.org/data/api/icesat2/level3a'
+
+        points = [] # store all beam data for one RGT
+
+        trackId,Date,cycle,bbox,product = paralist[0], paralist[1], paralist[2], paralist[3], paralist[4]
+
+        # iterate all six beams 
+        # for beam in beamlist:
+        # Generate API
+        payload =  {'product':product.lower(),
+                    'endDate': Date,
+                    'minx':str(bbox[0]),
+                    'miny':str(bbox[1]),
+                    'maxx':str(bbox[2]),
+                    'maxy':str(bbox[3]),
+                    'trackId': trackId,
+                    'outputFormat':'json'} # default return all six beams
+
+        # request OpenAltimetry
+        r = requests.get(base_url, params=payload)
+
+        # get elevation data
+        elevation_data = r.json()
+
+        # length of file list
+        file_len = len(elevation_data['data'])
+
+        # file index satifies aqusition time from data file
+        idx = [elevation_data['data'][i]['date'] == Date for i in np.arange(file_len)]
+
+        # get data we need
+        beam_data = list(compress(elevation_data['data'], idx))
+
+        if not beam_data:
+            return 
+
+        data_name = 'lat_lon_elev'
+
+        if product == 'ATL08':
+            data_name = 'lat_lon_elev_canopy'
+
+        # elevation array by iterating all six beams
+        beam_elev = [beam_data[0]['beams'][i][data_name] for i in np.arange(6) if beam_data[0]['beams'][i][data_name] != []]
+
+        if not beam_elev: # check if no data 
+            return  
+
+        OA_array = np.vstack(beam_elev)
+
+        if OA_array.size > 0:
+            return OA_array
+
+
+    def parallel_request_OA(self):
+        """ 
+        Requesting elevation data from OpenAltimetry API in parallel
+        Only supports OA_Products ['ATL06','ATL07','ATL08','ATL10','ATL12','ATL13']
+        For ATL03 Photon Data, OA only supports single date request according to: https://openaltimetry.org/data/swagger-ui/#/Public/getATL08DataByDate, 
+        and the geospatial limitation is 1 degree in both lat/lon
+        
+        Return: 
+            OA_data: dataframe containing all requested OA datasets
+        """
+
+        print ('Requesting data for elevation visualization, please wait...')
+
+        # generate parameter lists for OA requesting
+        para_lists = self.generate_OA_paras()
+
+        ### Parallel processing ###
+        pool = concurrent.futures.ThreadPoolExecutor(max_workers=30)
+        parallel_OA_data = {pool.submit(self.request_OA_elev,paralist): paralist for paralist in para_lists}
+
+        requested_OA_data=[]
+
+        for future in concurrent.futures.as_completed(parallel_OA_data):
+            r = future.result()
+            if r is not None:
+                requested_OA_data.append(r)
+
+        try:
+            OA_data = np.vstack(requested_OA_data) 
+            return OA_data
+
+        except:
+            pass
+
+    def visualize_elevation_matplotlib(self, OA_data):
+        """Visualize elevation data requested from OpenAltimetry API"""
+
+        # convert into geodataframe
+        geometry = gpd.points_from_xy(x=OA_data[:,1], y=OA_data[:,0])     
+        gdf_OA = gpd.GeoDataFrame({'lat':OA_data[:,0],
+                                   'lon':OA_data[:,1],
+                                   'h':OA_data[:,2]}, 
+                                  geometry=geometry, 
+                                  crs=CRS('EPSG:4326'))
+
+        # plot elevation
+        fig= plt.figure(figsize=(8, 8))
+        ax = fig.add_subplot(111)
+        cs = plt.scatter(gdf_OA['geometry'].x, gdf_OA['geometry'].y, c=gdf_OA['h'], vmin=gdf_OA['h'].min(),
+                         vmax=gdf_OA['h'].max(), cmap='RdYlBu_r', s=15, label='OpenAltimetry')
+
+        colorbar = fig.colorbar(cs, ax=ax)
+        colorbar.set_label('Elevation(m)', fontsize=15)
+        colorbar.ax.tick_params(labelsize=15)
+
+        plt.tight_layout()
+        plt.show()
+
+    def visualize_elevations_holoview(self, OA_data):
+        """
+        Visualize elevation data requested from OpenAltimetry API
+        https://holoviz.org/tutorial/Large_Data.html
+        https://datashader.org/getting_started/Interactivity.html
+        https://towardsdatascience.com/data-visualization-with-python-holoviz-plotting-4848e905f2c0
+        """
+        # convert array into dask dataframe
+        oa_array = da.from_array(OA_data, chunks=(5000, 3))
+        oa_df = dd.from_dask_array(oa_array, columns=['lat', 'lon', 'h'])
+        x, y = ds.utils.lnglat_to_meters(oa_df.lon, oa_df.lat)
+        ddf = oa_df.assign(x=x, y=y).persist()
+        points = hv.Points(ddf, ['x', 'y'])
+        tiles = hv.element.tiles.EsriImagery().opts(xaxis=None, yaxis=None, width=700, height=500)
+        return tiles * datashade(points)
+
+    def visualize_spatial_extent(self, extent_type, spat_extent):
+        """Default visualization of spatial extent"""
+
+        world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
+        f, ax = plt.subplots(1, figsize=(12, 6))
+        world.plot(ax=ax, facecolor="lightgray", edgecolor="gray")
+        geospatial.geodataframe(extent_type, spat_extent).plot(
+            ax=ax, color="#FF8C00", alpha=0.7
+        )
+        plt.show()
+    
--- a/icepyx/tests/test_OpenAltimetry.py
+++ b/icepyx/tests/test_OpenAltimetry.py
@@ -1,0 +1,22 @@
+import icepyx as ipx
+from icepyx.core.visualization import Visualize, filelist_latestcycle
+
+
+def test_OA_request():
+    
+    short_name = 'ATL06'
+    date_range = ['2019-4-14','2019-7-1']
+    bbox = [2, -73, 10, -70]
+    region = ipx.Query(short_name, bbox, date_range)
+    filelist_nsidc = region.avail_granules(ids=True, cycles=True, tracks=True)[0]
+    product = region.dataset
+    cycle_list  = list(map(int, set(region.avail_granules(ids=True, cycles=True, tracks=True)[1])))
+    filelist = filelist_latestcycle(filelist_nsidc, cycle_list)
+    OA_viz = Visualize(filelist, bbox, product)
+    oa_data = OA_viz.parallel_request_OA()
+    OA_viz.visualize_elevations_holoview(oa_data)
+    
+    assert oa_data.empty == False
+
+
+
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,17 @@
+datashader
 datetime
 fiona
 geopandas
 h5py
+holoviews
+hvplot
+intake
 matplotlib
 numpy
 pre-commit
 requests
 shapely
+
+
+
+


weiji14 (commented):
Hi @icetianli, good job on this, looks like you've picked up some `dask` and `datashader` skills too! Just a quick suggestion on first user experience, which is to add a progress bar while downloading. Also one minor suggestion to reduce the bounding box size for the test area.

I'll be happy to take a closer look later and can help with adding more tests. What will be helpful is if you can run [`black`](https://github.com/psf/black) on your code (cc @JessicaS11, can you remind us how the black pre-commit hook from https://github.com/icesat2py/icepyx/pull/96 works?)

icepyx/tests/test_OpenAltimetry.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_OA_request():
+    
+    short_name = 'ATL06'
+    date_range = ['2019-4-14','2019-7-1']
+    bbox = [2, -73, 10, -70]
+    region = ipx.Query(short_name, bbox, date_range)
+    filelist_nsidc = region.avail_granules(ids=True, cycles=True, tracks=True)[0]
+    product = region.dataset
```
JessicaS11, line 12: One of the items on my long dev to-do list is cleaning up some of the inconsistencies in variable names. It seems that to stay consistent with common DAAC/community syntax, dataset should refer to ICESat-2 and product should refer to ATLXX. Does that jibe with your experience? If so, I will work on making this change throughout the icepyx library, allowing us to ultimately get rid of this line of code (clearly a minor point here, but good motivation).

JessicaS11 (commented):
This looks great, @icetianli! I've got a few overarching comments for some next steps before we merge this PR:
1. As @weiji14 suggested, it would be great to run black on this to format it. If you install pre-commit (`pip install pre-commit`) in your local working environment, black should be run automatically when you make a commit, reformatting the files to give the library consistent spacing and use of " vs ' throughout. Let me know if you have any issues with this.
2. We should expand the docstrings to include input parameters and outputs.
3. To get the new functionality to show up in our documentation, we have to explicitly add it to the docs files.
4. Some examples to showcase this amazing functionality! This could be an addition to an existing how-to notebook or part of a standalone simple visualization, or maybe even both. My guess is you have either a script or notebook that you've been using this code in, which will make a great starting point for sharing it.

I'm happy to take on some of this work - it looks like a lot, but it's a critical piece of our efforts to make sure that your code is used by the broader community!

icepyx/core/visualization.py
```diff
@@ -0,0 +1,284 @@
...
+        # generate parameter lists for OA requesting
+        para_lists = self.generate_OA_paras()
+
+        ### Parallel processing ###
+        pool = concurrent.futures.ThreadPoolExecutor(max_workers=30)
+        parallel_OA_data = {pool.submit(self.request_OA_elev,paralist): paralist for paralist in para_lists}
+
+        requested_OA_data=[]
+
+        for future in concurrent.futures.as_completed(parallel_OA_data):
```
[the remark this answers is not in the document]
icetianli: Great advice, thanks @weiji14! Will check out tqdm to add a progress bar! Yes, the test region is Fimbul Ice Shelf and this bounding box is to test the 5-degree limit set by OA, but I agree that a large region with high-density data is not suitable for testing, will find another test site with a large bounding box but low-density data.

icepyx/tests/test_OpenAltimetry.py
icetianli: Yes - I think it’s better to use ‘product’ to describe different ATLXX rather than ‘datasets’, see OA API also uses [‘product’](https://openaltimetry.org/data/swagger-ui/#/Public/getLevel3Data) to be consistent with the [NSIDC syntax](https://icesat-2.gsfc.nasa.gov/science/data-products).

icepyx/core/visualization.py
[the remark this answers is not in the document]
weiji14: How about somewhere further north like on the Antarctic Peninsula? I saw your work on mapping the Larsen C Ice Shelf grounding zone at https://doi.org/10.5194/tc-2020-105, and that might be a suitable region, or we could just use an Equatorial area :smile:

P.S. Great paper too by the way, might ask you more about it at some point as I've got some office mates who were quite keen on a newer grounding line over the Amundsen Sea Embayment and Siple Coast.

icetianli:
Thanks @JessicaS11  and @weiji14  for the advice! Sorry it took me so long to update! I have now improved the default spatial extent plot using geoviews, I made a [pull request](https://github.com/icesat2py/icepyx/pull/176) to implement this so it can be merged separately. For OpenAltimetry data visualization, I deprecated the functionality that only requests the lastest cycle demonstrated in my last commit, as now what I think the most valuable function of OA API is to provide a low-cost way of playing ICESat-2 data, **e.g. compare geolocations and elevations between different ground tracks and cycles for a large region**, without going through the trouble of downloading and plotting data separately. 

Now the script can show elevation distributions from different cycles, see [notebook](https://github.com/icesat2py/icepyx/blob/openaltimetry/examples/ICESat-2_Data_Visualization_Example.ipynb). In terms of future development, I think it can add a `lat-elevation` plot for individual RGT/ cycle in a similar way as OpenAltimetry since the data is now stored in xarray so it becomes quite handy to achieve this. 

The improvements to speed up data request include:

* Avoid sending empty requests
* Append requested data for each 5*5 grid in a multi-dimensional  xarray dataset 
* Use cycles as a control parameter to visualize elevations
* Add colorbar to datashade plot 

I only test the functions in Antarctica Peninsula where the data density is relatively low. For a large region with a long time span, OA request may take several minutes, but I think this can be further enhanced by improving the parallel processing or just setting a limit for the spatial extent and date range to warn the users that it may take a long time to process.

weiji14:
Hi @icetianli, welcome back, hope your research is going well! Just had a quick scan of the changes here and it looks like you've got made some really nice improvements, especially with that ICESat-2 cycle slider! I'll give this a full review tomorrow as it's getting late in my timezone.

In the meantime, could you merge in the changes from the development into this branch using `git merge development`? There's a small conflict on the `requirements.txt` file you'll need to resolve. Or you can do it via the Github UI by clicking on the 'Resolve conflicts' button:

image

JessicaS11:
This looks awesome! Thanks for all these updates, @icetianli, and the example is great. I haven't had a chance to do a line-by-line review, but I'm thinking it would make sense to better integrate the capabilities from #176 into this new module (we can always merge that PR into development, then use this one to migrate it all to the visualization module). Then, we can call visualize directly from the query object with options for the different methods of visualization. We'll also need to add in some checks for valid datasets specifically for this step, because as I remember OA doesn't host all datasets, so we'll need to let users know if they can't quick-view the dataset they're requesting. We can talk in more detail about some of this at [Tuesday's icepyx meeting](https://discourse.pangeo.io/t/icepyx-team-meetings/722/2?u=jessicas11)!

icepyx/core/visualization.py
```diff
@@ -0,0 +1,312 @@
...
+                   'maxx': str(bbox[2]),
+                   'maxy': str(bbox[3]),
+                   'trackId': trackId,
+                   'outputFormat': 'json'}  # default return all six beams
+
+        # request OpenAltimetry
+        r = requests.get(base_url, params=payload)
+
+        # get elevation data
+        elevation_data = r.json()
```
weiji14, line 181: I feel like there might be a way to simplify data loading from JSON using [`pandas.json_normalize`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.json_normalize.html) (did some work a while ago at https://github.com/weiji14/deepicedrain/blob/v0.4.0/atl06_play.py#L507), but will need to play around a bit to ensure it works. Stay tuned on this!

weiji14 (commented):
Just a few more comments for now, mostly just minor typos and code style suggestions. I'll still need to walk myself through the logic a bit more and play around with your 'examples/ICESat-2_Data_Visualization_Example.ipynb' notebook (which should be listed under the README.md file by the way under https://github.com/icesat2py/icepyx/tree/v0.3.2#examples-jupyter-notebooks).

> I'm thinking it would make sense to better integrate the capabilities from #176 into this new module (we can always merge that PR into development, then use this one to migrate it all to the visualization module). Then, we can call visualize directly from the query object with options for the different methods of visualization.

Agree. So the workflow would be something like `Merge #176 into 'development'` -> `Merge 'development' into this PR #144` -> `Migrate the 'visualize_spatial_extent' to 'visualization.py'`.

> We'll also need to add in some checks for valid datasets specifically for this step, because as I remember OA doesn't host all datasets, so we'll need to let users know if they can't quick-view the dataset they're requesting. We can talk in more detail about some of this at [Tuesday's icepyx meeting](https://discourse.pangeo.io/t/icepyx-team-meetings/722/2?u=jessicas11)!

Yep, something like an allowlist to check if the product is in `['ATL06','ATL07','ATL08','ATL10','ATL12','ATL13']` should work. I'll *try* and listen in for the meeting tomorrow but probably won't talk much cause it'll be 6am for me :laughing:

icepyx/core/visualization.py
```diff
@@ -0,0 +1,340 @@
...
+
+        if self.product in ['ATL06', 'ATL07', 'ATL08', 'ATL10', 'ATL12', 'ATL13']:
+
+            OA_ds = self.parallel_request_OA()
+
+            if OA_ds is None:
+                print('No data')
+
+            else:
+                ddf = OA_ds.to_dask_dataframe().astype({'lat': 'float', 'lon': 'float','elevation': 'float'})
```
weiji14, line 317: Actually, I'm wondering if it's better to put everything in a dask.DataFrame directly, instead of going through an xarray.Dataset (though I imagine the xarray.Dataset format would be useful for storing attributes). What do you think?

```python
dask.dataframe.from_array(x=OA_array, columns=["lat", "lon", "elevation"])
```

```
Dask DataFrame Structure:
	lat	lon	elevation
npartitions=1			
0	float64	float64	float64
2831	...	...	...
Dask Name: from_array, 1 tasks
```

weiji14 (commented):
Just restarting the review process again :smile: Sorry if it looks like a lot (got a bit carried away), but hopefully you can just batch commit them, or I can push the commits to this branch directly if you're ok with it. Feel free to disagree or discuss about any of the recommended changes too, you've worked on this code a lot more than I have and tbh, it's been a while since I worked on ATL06!

To continue with the rest of the review though, could you please:

1. Merge in the latest changes from the 'development' branch into this 'openaltimetry' branch (click on the 'Resolve conflicts' button in the GitHub UI and remove the extra blank line).
2. Clear the output of the first cell (the one with `from icepyx.core.visualization import Visualize`) in your ICESat-2_Data_Visualization_Example.ipynb notebook and re-commit it to GitHub. This will result in a smaller diff (the bokeh svg icon is very large for some reason), and hopefully we can comment/review on the *.ipynb file easier later.

**P.S.** I'm doing some work at #196 to enable showing jupyter notebooks directly as a 'Tutorial' on the icepyx ReadTheDocs page. If that work gets merged in, we'll have to move your ICESat-2_Data_Visualization_Example.ipynb to a new home under `doc/`, and users will be able to see your walkthrough tutorial even better! Here's a sneak peek of how it would look like:

image

Keep up the good work :raised_hands:

icepyx/core/visualization.py
```diff
@@ -0,0 +1,340 @@
...
+
+        for bbox_i in bbox_list:
+            ### NEED Change for Track & Cycle Implementation
+            region = ipx.Query(self.product, bbox_i, self.date_range)
+            icesat2_files = region.avail_granules(ids=True, cycles=True, tracks=True)[0]
+
+            if not icesat2_files:
+                continue
+            else:
+                icesat2_files_latest_cycle = files_in_latest_n_cycles(icesat2_files, [int(c) for c in region.cycles])
```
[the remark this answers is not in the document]
icetianli: Yes you're right...somehow `region.cycles` & `region.tracks` won't show the exact cycle/ track numbers unless you input these two parameters, maybe @JessicaS11 has a better idea on this?

icepyx/core/visualization.py
```diff
@@ -0,0 +1,340 @@
...
+        file_len = len(elevation_data['data'])
+
+        # file index satisfies acquisition time from data file
+        idx = [elevation_data['data'][i]['date'] == Date for i in np.arange(file_len)]
+
+        # get data we need
+        beam_data = list(compress(elevation_data['data'], idx))
+
+        if not beam_data:
+            return
```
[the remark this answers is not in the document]
icetianli: Thanks @weiji14 for the great advice, they're very helpful!! I'll check this one and the following comments in next couple of days

icepyx/core/visualization.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+class Visualize:
+
+    def __init__(self, product, bbox, date_range=None, cycles=None, tracks=None):
+        self.product = product
+        self.bbox = bbox
+        self.date_range = date_range
+        self.cycles = cycles
+        self.tracks = tracks
```
JessicaS11, line 53: I remember we talked at an icepyx meeting about making it so that the user could call `Visualize()` or `Query.visualize()`. With that in mind, I'm thinking that we should use the `__init__` to allow both inputs and end up with a Visualize object that has the required info that's been validated (so if the user enters the search criteria directly, `query_icesat2_filelist` is called and we thus know that if a Visualize object has been created it has a valid file list to do the visualizations). I'm happy to work on this cross-object synchronization, @icetianli.

icepyx/core/visualization.py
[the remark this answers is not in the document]
icetianli: Yes,  `pd.json_normalize` is much more clearer, have updated this part in the latest commit!

icepyx/core/visualization.py
```diff
@@ -0,0 +1,340 @@
...
+        beam_data = list(compress(elevation_data['data'], idx))
+
+        if not beam_data:
+            return
+
+        data_name = "lat_lon_elev_canopy" if product == "ATL08" else "lat_lon_elev"
+
+        # iterate six beams, sampling rate: 1/20
+        beam_elev = [beam_data[0]['beams'][i][data_name][::20] for i in np.arange(6) if
+                     beam_data[0]['beams'][i][data_name] != []]
```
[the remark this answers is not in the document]
icetianli: Now I added a `sample_rate` parameter, for ATL06 it's 1/50, for other products (segment length normally about 100 m) I use 1/20 to avoid over-sampling for low latitude regions.

icepyx/core/visualization.py
icetianli: Agree, `dask.dataframe` is better as in the `visualization_elevation` function I converted `xarray.dataset` into a `dask dataframe` anyway so really no need to generate xarray.dataset here.

icepyx/core/visualization.py
icetianli: Thanks @JessicaS11 for checking on this! Yes that's what we discussed in our last meeting, I agree the ultimate goal should be directly calling `Visualize` in `Query` class, it's a good idea to use `query_icesat2_filelist` to validate if the searched region has data or not so we can inform users of the results. The only problem is `Visualize` class needs to create an individual Query class object for each 5*5 bounding box to query OA API, so I am not very sure what's the best way to connect `Query` and `Visualize` classes, I'll check later.

JessicaS11: Ah - that makes sense. I'll give it more thought too (I think I was misunderstanding exactly what `query_icesat2_filelist` does), but at the moment I don't think creating multiple query objects within the code (since the user never sees them) will be an issue.

I'm thinking the mechanisms for access could be something like:
1. `reg_a_vis = ipx.Query(product, bbox, date/cycles/tracks).visualize()`
2. `reg_a_vis = ipx.Visualize([inputs])` where inputs is either a) an existing query object; b) product, bbox, date/cycles/tracks. In the former case, if the user inputs a shapefile or polygon, we could turn it into a bbox using `.bounds()`.

icetianli: Now `Query` should be able to call `Visualize` directly via `visualize_elevation` in `Query` class, see https://github.com/icesat2py/icepyx/pull/144/commits/0911ea45976490464af759136388d87402750827, I also converted the polygon/ shapefile input into bounding box by using `[shapely.minimum_rotated_rectangle](https://shapely.readthedocs.io/en/stable/manual.html)`, so non-bounding box input can also be used for elevation check now.

icepyx/tests/test_visualization.py
```diff
@@ -0,0 +1,16 @@
+from icepyx.core.visualization import Visualize
```
weiji14, line 1: Need to import pytest to fix the broken Travis CI test

```suggestion
import pytest

from icepyx.core.visualization import Visualize
```

icepyx/core/visualization.py
```diff
@@ -0,0 +1,408 @@
...
+
+        except:
+            beam_data = None
+
+        if not beam_data:
+            return
+
+        data_name = "lat_lon_elev_canopy" if product == "ATL08" else "lat_lon_elev"
+
+        sample_rate = 50 if product == "ATL06" else 20
```
weiji14, line 267: > Now I added a sample_rate variable, for ATL06 it's 1/50, for other products (segment length normally about 100 m) I use 1/20 to avoid over-sampling for low latitude regions.

This is great!

icepyx/core/query.py
```diff
@@ -939,3 +940,34 @@ def visualize_spatial_extent(
...
+            bbox = [lonmin, latmin, lonmax, latmax]
+
+        date_range = [self._start.strftime('%Y-%m-%d'),
+                      self._end.strftime('%Y-%m-%d')] if hasattr(self, '_start') else None
+
+        cycles = self._cycles if hasattr(self, '_cycles') else None
+
+        tracks = self._tracks if hasattr(self, '_tracks') else None
+
+        print(product, bbox, date_range, cycles, tracks)
```
JessicaS11, line 967: Is there a purpose for showing this to the user, or can we remove it once we're done development?

icetianli: Yeah we can remove it when we finishing this code, sorry it's for testing yesterday! I also need to update the test script, previously it just tested whether the returned `xarray.dataset` has data or not.

requirements.txt
```diff
@@ -1,8 +1,13 @@
+backoff
+datashader
 fiona
 geopandas
 h5py
+holoviews
```
JessicaS11, line 6: is geoviews a dependency of holoviews? If so, we'll need to do the same gymnastics we did before for those who haven't/can't install geoviews. Relatedly (though not to these particular lines of code), we should add a few checks/messages for users trying to use/import the visualization module who don't have geoviews. Otherwise they'll get top level import errors (we might be able to simplify this by just doing it once in the `__init__.py`, though it might make sense to have it explicitly in the places it will become an issue).

icepyx/core/query.py
JessicaS11: No worries! My code always has a lot of print statements while I'm writing it. : )

requirements.txt
icetianli: Looks like `[holoviews](https://github.com/holoviz/holoviews/blob/master/setup.py)` doesn't require `cartopy` or `geoviews`, I removed `geoviews` from `Visualize` as it's only required to plot spatial extent, which was moved to `Query` and `geoviews` dependency has been solved, just tested in a fresh environment using `pip install -e.` and everything is working fine.

icetianli: Yeah it's a good idea to add checks for `geoviews import` error, should we just check this import error then tell them different ways of installing `geoviews` depending on whether they `conda install icepyx` or `pip install icepyx`?

JessicaS11: Awesome! At some point I think we discussed moving `Query.visualize_spatial_extent()` to the `Visualize` module (obviously, in either case we'll keep it so that `geoviews` is only an optional dependency and this method can be called directly from the Query object). Thoughts? I haven't fully argued this out with myself yet.

icetianli: Just watched the webinar, great talk! Yeah we discussed it before about migrating `visualize_spatial_extent()` from https://github.com/icesat2py/icepyx/pull/176 back to the `Visualize` module, but I am not sure if it's necessary now...since everything all ended up in `Query` module, we could just call `Query.visualize_spatial_extent()` for spatial extent and `region.visualize_elevation()` for elevation visualization directly.

icepyx/core/visualization.py
```diff
@@ -0,0 +1,431 @@
...
+                if r is not None:
+                    requested_OA_data.append(r)
+
+        if not requested_OA_data:
+            return
+        else:
+            OA_data_da = da.concatenate(requested_OA_data, axis=0)
+            return OA_data_da
+
+    def viz_elevation(self):
```
weiji14, line 358:
```suggestion
    def viz_elevation(self) -> (hv.DynamicMap, hv.Layout):
```

requirements.txt
weiji14: Agree with @icetianli to keep `visualize_elevation()` where it is now in the `Query` object. I'm on the fence on whether to make `holoviews` a required dependency of `icepyx` because it has a lot of sub-dependencies (though not as bad as geoviews), but it seems inevitable if we're moving into exploratory data analysis work (people like fancy GUIs) so I'm ok with adding it.

weiji14 (commented):
This is coming along real nicely @icetianli, good job on moving from `xarray.Dataset` to `dask.dataframe` in e9f88b7bc8171bb88a2518e444a63bbf3463dfe5, and adding more unit tests! I was so happy to see Jessica demo-ing your `visualize_spatial_extent()` function, and can't wait to get this elevation visualization PR in for other people to use too!

Anyways, mostly minor documentation related comments/suggestions this time, and two recommendations on the ICESat-2_Data_Visualization_Example.ipynb notebook:

1. Clear the output of the cells containing the holoviews icon (holoviews icon) because they're large. This notebook is currently 8.3MB, and clearing those cells should bring it down by a few MBs.
2. Re-run the cells so the cell numbering goes as 1, 2, 3 instead of 70+

If it helps, I've added a Binder button to your top post at https://github.com/icesat2py/icepyx/pull/144#issue-486910161 so you can have a fresh cloud environment to run this branch. Caveat is that `geoviews` isn't installed, so you won't get an interactive `visualize_spatial_extent`.

Other than that, check with @JessicaS11 on what else might be needed. I think she mentioned finalizing this in the next week or two, so hopefully :crossed_fingers: we don't have to wait too much longer!

requirements.txt
JessicaS11: Sounds good on adding holoviews as a dependency. To clarify, either way we'd have `visualize_elevation()` and `visualize_spatial_extent()` methods for Query objects; the bigger question is where the implementation of those lives (in the query or visualization modules). My thinking was that this is analogous to the other modules (e.g. granules), which very early on were part of a giant query module and ultimately separated out into submodules with their own class objects to break the code into more reasonably sized units. If, as @weiji14 notes, we get more into including GUIs for data analysis work, then I suspect having all of the visualization components stored in one `Visualization` object is going to be smoother. I will work on some of these changes as I was envisioning them and we can just revert the commit if it makes things worse.

icetianli: Sounds like a plan!

icepyx/core/visualization.py
```diff
@@ -0,0 +1,431 @@
...
+from holoviews.operation.datashader import rasterize
+import intake.source.utils
+import numpy as np
+import pandas as pd
+import requests
+from tqdm import tqdm
+
+import icepyx as ipx
+
+hv.extension("bokeh")
```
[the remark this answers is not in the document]
icetianli: Thanks for the comments! The Binder test is great! I think `hv.extension("bokeh")` is necessary here, `matplotlib` backend is only for static map and we want interactive plot right? Yes we can put `holoviews` backend in the notebook, then we also need to do the same for `geoviews`. Not all users will have geoviews installed, so it may throw an import error if they don't have it, but this error will remind users to install geoviews themselves instead of just plot the static extent map silently in `visualize_spatial_extent()` as before, not sure about this though...any idea?

requirements.txt
JessicaS11: I just pushed my version of this, including examples of each access method at the end of your example notebook. This approach has the added benefit of using the checks built into `Query` to validate the user's inputs.

icepyx/core/visualization.py
```diff
@@ -0,0 +1,503 @@
...
+
+        df = pd.json_normalize(data=elevation_data, record_path=["data"])
+
+        # get data we need (with the correct date)
+        try:
+
+            df_series = df.query(expr="date == @Date").iloc[0]
+            beam_data = df_series.beams
+
+        except:
```
JessicaS11, line 347: Are there multiple exceptions we're capturing here, or can we explicitly list the one we're targeting?

icepyx/tests/test_visualization.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@pytest.mark.parametrize("product, date_range, bbox, expect", [
+    ('ATL06',['2019-6-15', '2019-7-1'], [-64.5, -66, -63.5, -65], 3240),
+    ('ATL07',['2019-7-1', '2019-8-1'], [-65, -66, -64.5, -65], 7160),
+    ('ATL08',['2019-6-15', '2019-7-1'], [-18, 63, -17, 64], 852),
+    ('ATL10',['2019-8-1', '2019-9-1'], [-64, -67, -60, -60], 7375),
+    ('ATL12',['2019-7-1', '2019-10-1'], [-65.5, -65.5, -64.5, -65], 95),
+    ('ATL13',['2019-6-1', '2019-12-1'], [-75, -51, -74, -50], 20),
```
JessicaS11, line 12: Do you know how OA handles new version releases? I'm concerned these tests could break if a new version of a product is released, but I don't know if that's a thing we can control for or will just have to fix manually.

JessicaS11:
This is looking so great - thanks for all your hard work, @icetianli! I managed to break your tests, which I will work to fix pending our decision about how we initialize the Visualization object. I'd also like to try and add a few more tests to target some of the other functions you've added.

Also, since I made the PR a mess by running black on some of the files, over the next couple of days I will put in a separate PR to do the reformatting, and once we merge it into development we can bring it in here to clean things up again. Sorry about that - I didn't realize how many modules had fallen behind on formatting!

weiji14:
> This is looking so great - thanks for all your hard work, @icetianli! I managed to break your tests, which I will work to fix pending our decision about how we initialize the Visualization object. I'd also like to try and add a few more tests to target some of the other functions you've added.

@JessicaS11, could you describe your idea a bit more on how the Visualization object should be initialized? I read your note on https://github.com/icesat2py/icepyx/pull/144#discussion_r623106393 and saw commit b862ecafc20b0eb2e85081958c31f8dba7d6b35c (which looks good by the way), so is this pretty much done? Or do you mean having an API like `Query.visualize.elevation()` and `Query.visualize.spatial_extent()` to replace the current `Query.visualize_elevation()` and `Query.visualize_spatial_extent()` calls (probably best left to a separate PR tbh).

> Also, since I made the PR a mess by running black on some of the files, over the next couple of days I will put in a separate PR to do the reformatting, and once we merge it into development we can bring it in here to clean things up again. Sorry about that - I didn't realize how many modules had fallen behind on formatting!

Just on this, you can also revert commits 53b55c42a97e681dc5ba954acee6e418a525e350 and 8c1cd052b77e1a8570efdcd739db3d107d14469d to remove unrelated changes and reduce the git diff on this PR. Following that, do cherry-pick (using -n/--no-commit) on 8c1cd052b77e1a8570efdcd739db3d107d14469d to get only the relevant changes (see https://stackoverflow.com/questions/5717026/how-to-git-cherry-pick-only-changes-to-certain-files).

```bash
git revert 53b55c42a97e681dc5ba954acee6e418a525e350
git revert 8c1cd052b77e1a8570efdcd739db3d107d14469d

git cherry-pick -n 8c1cd052b77e1a8570efdcd739db3d107d14469d
git add doc/source/user_guide/documentation/query.rst examples/examples.rst icepyx/core/visualization.py
git commit -m "add docstrings and visualization to docs; reformat with black"
```

icepyx/core/visualization.py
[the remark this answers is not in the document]
weiji14: Yes we prefer to have interactivity out of the box, but I think Bokeh is the default backend already right? I.e. users would have to explicitly set `hv.extension("matplotlib")` to get an static plot for Bokeh >= 2.0, otherwise an interactive plot should be returned by default. Maybe try removing this line and re-run your example notebook to see if you get an interactive or static map.

icepyx/tests/test_visualization.py
weiji14: There doesn't seem to be a version option listed at https://openaltimetry.org/data/swagger-ui/#/Public/getATL06DataByDate (or the other ATLAS products). Might just have to update this every so often, There has only been 4 version releases for ATL06 over the past 2 years so shouldn't be too much work :crossed_fingers:.

icepyx/core/visualization.py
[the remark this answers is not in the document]
icetianli: If not set `hv.extension`, it will return: 

> ValueError: No plotting extension is currently loaded. Ensure you load an plotting extension with hv.extension or import it explicitly from holoviews.plotting before applying any options.

If set `hv.extension("matplotlib")`, it will have `ValueError: Tiles type not found, could not apply options.`, looks like map tiles are only supported in Bokeh, see https://holoviews.org/reference/elements/bokeh/Tiles.html. 

Also to my understanding, 

> The default plotting extension for HoloViews until a 2.0 release is Matplotlib when HoloViews will start defaulting to Bokeh (see the Plotting with Bokeh user guide).
 
2.0 release is the version requirement for Holoviews (current version is 1.14.3) rather than Bokeh, so the default is a static map from `hv.extension("matplotlib")`, Bokeh will be the default when Holoviews >= 2.0. 

```
import numpy as np
import holoviews as hv
from holoviews import opts
img = hv.Image(np.random.rand(10, 10))
fig = hv.render(img)
fig
```
This example returns a static map. 

Since we only need interactive plots, I think we could just set Bokeh as default in backend for geoviews and holoviews.

icepyx/tests/test_visualization.py
icetianli: Agree with @weiji14, we could update this when a new version is released.

icepyx/core/visualization.py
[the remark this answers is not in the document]
weiji14: Ah ok, thanks for the clarification! Then it's fine to keep the `hv.extension("bokeh")` line intact. I got confused with the bokeh and holoviews versions, I know bokeh is at v2 but hadn't realize holoviews is still at v1 :grimacing: Also we need the map tiles so Bokeh backend it is then!

icepyx/core/visualization.py
icetianli: Yes, sorry I forgot to specify them, they are `NameError` and `KeyError` caused by empty dataframes return from `pd.json_normalize(data=elevation_data, record_path=["data"])`, could you please add them in your end? I am afraid it may cause chaos if I push changes from my side atm......thanks!

```
try:

    df_series = df.query(expr="date == @Date").iloc[0]
    beam_data = df_series.beams

except (NameError, KeyError) as error:
    beam_data = None
```

weiji14: It should be safe to push the `except (NameError, KeyError)` change @icetianli (as long as you do a local `git pull` from the `openaltimetry` branch first). Jessica's recent changes didn't modify L347 so the `git revert` commands I mentioned at https://github.com/icesat2py/icepyx/pull/144#issuecomment-829686206 won't cause any problems.

icetianli:
> I managed to break your tests, which I will work to fix pending our decision about how we initialize the Visualization object. I'd also like to try and add a few more tests to target some of the other functions you've added.

This change looks good to me @JessicaS11 , it's nice to be able to pass `Query` object directly to `Visualize`. So currently it's `ipx.Visualize.viz_elevation()` for elevation, then why still keep `visualize_elevation()` function in `Query` class? I am a bit confused about this.

icepyx/core/visualization.py
icetianli: OK, done

JessicaS11:
>  So currently it's `ipx.Visualize.viz_elevation()` for elevation, then why still keep `visualize_elevation()` function in `Query` class? I am a bit confused about this.

> Or do you mean having an API like Query.visualize.elevation() and Query.visualize.spatial_extent() to replace the current Query.visualize_elevation() and Query.visualize_spatial_extent() calls (probably best left to a separate PR tbh).

Good points, @icetianli and @weiji14. After some more thought, I think that it makes the most sense to encourage users to access the visualizations as methods on the Query object (`reg_a.visualize_spatial_extent()` and `reg_a.visualize_elevation()`), since that's when they're most likely to be using the functions. As with all the other submodules (`granules.py`, `variables.py`, etc.), these can be accessed independently, if the user so chooses, by importing the module directly `from icepyx.core.visualization import Visualize`). I thus took the Visualize object importation out of the `__init__.py` but did leave a note that you could access the module directly at the end of the visualization example notebook. This also will decrease confusion if we ultimately add any other types of visualizations for working with the data.

Thanks for asking all these great clarifying questions! Let me know if more changes are still needed.

JessicaS11:
Travis is failing on the ATL07 `test_visualization_date_range`. I can reproduce the error with:
```
from icepyx.core.visualization import Visualize
vis = Visualize(product='ATL07', spatial_extent=[-65, -66, -64.5, -65], date_range=['2019-7-1', '2019-8-1'])
```

And the error traceback is:
```
IndexError                                Traceback (most recent call last)
<ipython-input-10-19eb5e1c5abb> in <module>
----> 1 vis.parallel_request_OA().size

~/computing/icepyx/github_repo/icepyx/icepyx/core/visualization.py in parallel_request_OA(self)
    428                 total=len(parallel_OA_data),
    429             ):
--> 430                 r = future.result()
    431                 if r is not None:
    432                     requested_OA_data.append(r)

/opt/anaconda3/envs/icepyx/lib/python3.9/concurrent/futures/_base.py in result(self, timeout)
    431                 raise CancelledError()
    432             elif self._state == FINISHED:
--> 433                 return self.__get_result()
    434 
    435             self._condition.wait(timeout)

/opt/anaconda3/envs/icepyx/lib/python3.9/concurrent/futures/_base.py in __get_result(self)
    387     def __get_result(self):
    388         if self._exception:
--> 389             raise self._exception
    390         else:
    391             return self._result

/opt/anaconda3/envs/icepyx/lib/python3.9/concurrent/futures/thread.py in run(self)
     50 
     51         try:
---> 52             result = self.fn(*self.args, **self.kwargs)
     53         except BaseException as exc:
     54             self.future.set_exception(exc)

~/computing/icepyx/github_repo/icepyx/icepyx/core/visualization.py in request_OA_data(self, paras)
    342         try:
    343 
--> 344             df_series = df.query(expr="date == @Date").iloc[0]
    345             beam_data = df_series.beams
    346 

/opt/anaconda3/envs/icepyx/lib/python3.9/site-packages/pandas/core/indexing.py in __getitem__(self, key)
    893 
    894             maybe_callable = com.apply_if_callable(key, self.obj)
--> 895             return self._getitem_axis(maybe_callable, axis=axis)
    896 
    897     def _is_scalar_access(self, key: Tuple):

/opt/anaconda3/envs/icepyx/lib/python3.9/site-packages/pandas/core/indexing.py in _getitem_axis(self, key, axis)
   1499 
   1500             # validate the location
-> 1501             self._validate_integer(key, axis)
   1502 
   1503             return self.obj._ixs(key, axis=axis)

/opt/anaconda3/envs/icepyx/lib/python3.9/site-packages/pandas/core/indexing.py in _validate_integer(self, key, axis)
   1442         len_axis = len(self.obj._get_axis(axis))
   1443         if key >= len_axis or key < -len_axis:
-> 1444             raise IndexError("single positional indexer is out-of-bounds")
   1445 
   1446     # -------------------------------------------------------------------

IndexError: single positional indexer is out-of-bounds
```

It looks like it's coming from the specification of error type @icetianli  added to line 347 of `visualize.py`. Do we just need to add this error type, or should we be updating the test?

icetianli:
Ah yes! It's another error caused by empty dataframe by `df.query` @JessicaS11 , could just add `IndexError` in the error list `except (NameError, KeyError, IndexError) as error:`, sorry didn't catch it last time!

icetianli:
> Good points, @icetianli and @weiji14. After some more thought, I think that it makes the most sense to encourage users to access the visualizations as methods on the Query object (`reg_a.visualize_spatial_extent()` and `reg_a.visualize_elevation()`), since that's when they're most likely to be using the functions. As with all the other submodules (`granules.py`, `variables.py`, etc.), these can be accessed independently, if the user so chooses, by importing the module directly `from icepyx.core.visualization import Visualize`). I thus took the Visualize object importation out of the `__init__.py` but did leave a note that you could access the module directly at the end of the visualization example notebook. This also will decrease confusion if we ultimately add any other types of visualizations for working with the data.

Thanks for adding these additions @JessicaS11! They look great, agree it's better to call different visualization methods directly from `Query` object as default and in the meantime keep the options open for users to import `visualization`, no further questions from me.

icepyx/core/visualization.py
```diff
@@ -0,0 +1,503 @@
...
+                    format_string = (
+                        "{product:5d}_{datetime:%Y%m%d%H%M%S}_{rgt:04d}{cycle:02d}{"
+                        "region:02d}_{release:03d}_{version:02d}.h5"
+                    )
+                temp = intake.source.utils.reverse_format(
+                    format_string=format_string, resolved_string=fname
+                )
+
+                rgt = temp["rgt"]  # Reference Ground Track
+                cycle = temp["cycle"]  # Cycle number
```
JessicaS11, line 270: As I was looking through `visualization.py` to try and add some more tests, it occurred to me that this is similar to some code Tyler added for querying granules by cycle/track. I'll dig into it more tomorrow, but on a quick look it seems like we could use (`granules.gran_IDs()`)[https://github.com/icesat2py/icepyx/blob/4d6e64b95e61609364385e080195c39c4486b182/icepyx/core/granules.py#L36] to get this info here.

icetianli: Yes looks like so, will take a closer look and test the function!

icetianli: OK, added a new function `gran_paras` based on `granules.gran_IDs()` to get these information: https://github.com/icesat2py/icepyx/pull/144/commits/60c6d8a6157feb1db1a3d057d1341562bb652650

JessicaS11: Awesome! I took it a step further and wrapped your function into `granules.gran_IDs()` directly so that we should only have to maintain the string parsing in one location. I think that long-term we could probably streamline this type of string parsing/handling across all modules, but I don't think we should hold this PR up any more for that sort of thing.

icetianli: Yes, put the date string in one place looks so much better!

JessicaS11:
Thanks @icetianli for taking this on, and @weiji14 for all your review help. I'm going to go ahead and approve this PR, but I won't merge it in case either of you have any last edits/changes/fixes since I was the last one to push a commit. Feel free to merge (and if you're feeling really ambitious open a dev-->main PR, though we'll need to add a changelog before we merge that).

weiji14 (approved):
Nope, this PR looks good to me so feel free to squash and merge this in! Well done on seeing this through to the end, I hope you learned some new things from it, and sorry that it took as long as a paper review!

PR #148, opened 2020-11-03
use readable_granule_name to reduce by RGT, cycle

tsutterley:
readable_granule_name can be the filenames to be used in subsetting

setting the CMR options for "pattern" wildcards and "OR" capability

change params in url requests from dictionary to string to prevent encoding
first creates a list of all params (readable_granule_names will be a list) then combines to a single url string

thanks @wallinb for the help


icepyx/tests/test_granules.py
```diff
@@ -592,16 +592,14 @@ def test_no_granules_in_search_results():
 
 def test_correct_granule_list_returned():
     reg_a = ipx.Query(
-        "ATL06", [-55, 68, -48, 71], ["2019-02-20", "2019-02-28"], version="2"
+        "ATL06", [-55, 68, -48, 71], ["2019-02-20", "2019-02-28"]
```
JessicaS11, line 595: If we don't specify a version here, this test will fail as soon as a new version of the datasets is available. Updating to v3 is a good idea though.

icepyx/core/APIformatting.py
```diff
@@ -142,25 +142,25 @@ def combine_params(*param_dicts):
 
     Returns
     -------
-    single dictionary of all input dictionaries combined
+    url string of all input dictionaries combined
```
JessicaS11, line 145: Discussion question: are there pros/cons to turning this into a url here versus leaving it as a dictionary? Is this a CMR/API change that it now requires as input as a url and can't accept the dictionary input?

JessicaS11 (commented):
@andypbarrett Is this what you were getting at for inputting specific granule filenames to download data? Does this work for your application given that it still requires temporal and spatial inputs? Or should we be trying to update the object initialization to have multiple options for minimal info to include ((dataset and temporal and spatial) or (dataset and cycle and track) or (granule name))?

icepyx/tests/test_granules.py
tsutterley: good point.  I added `version="3"`

icepyx/core/APIformatting.py
tsutterley: keeping it as a dictionary results in the parameters being [percent-encoded](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) which breaks the CMR pattern capability (pattern wildcards get converted).  As a middle ground (if this might break some functionality), could encode some parameters but not anything for the CMR pattern matching.

icepyx/tests/test_granules.py
andypbarrett: I'm thinking of a query

`reg_a = ipx.query(granule=list_of_granules)`

The product name, dates, version etc could be parsed from the granule file names.

I have a workflow to do this with the CMR.  All thoughts on this welcome.

icepyx/core/APIformatting.py
JessicaS11: That makes sense. I haven't gone all the way through to see if the change will break anything (I want to play with some subsetting), but if it doesn't break, then we can use this formatting for all parameters. Should we use urllib for this, or keep it hard-coded for our use case?

JessicaS11:
> I'm thinking of a query
> 
> `reg_a = ipx.query(granule=list_of_granules)`
> 
> The product name, dates, version etc could be parsed from the granule file names.
> 
> I have a workflow to do this with the CMR. All thoughts on this welcome.

@andypbarrett I think this PR will help provide some of the infrastructure to make it easier to implement this within icepyx. If CMR can ultimately handle just a list of granules, then there's no reason to bother with getting the product name/dates/version/etc. unless the user wants it (and then we'll start getting into how to handle multiple products/date ranges/etc). The changes you'd need to make are to `query.py` and `validate_inputs.py` and possibly the formatting and/or granules modules. I can provide some more direction if you'd like. Do you want to contribute that as part of this PR, or should we plan to add it separately in the future?

andypbarrett:
Hi Jessica

My inclination is to add it as a separate PR.

Thanks for the pointers.  I'll dig into this and ask for more direction
once I have a better idea.

Andy

JessicaS11:
@andypbarrett sounds good!

icepyx/core/APIformatting.py
JessicaS11: @tsutterley Have you actually been able to place an order (through icepyx) with these updates? I started going through the example notebook and got errors when I called `region_a.order_granules()` because of internal calls to `req_params` (it's trying to get a value from a dict, but `req_params` is now a string). I didn't dig deeper to see how many instances would need to be modified to handle the change.

Currently the order functions are only weakly tested once a week on master by a cron or in a PR to master (because of the need for a login), but you can trigger them in another build by adding "[nsidc_tests]" to the commit message. There aren't currently enough unit tests to tell you much without viewing the travis output, but it should at least tell you here that it was unable to order and download data for some reason.

tsutterley: @JessicaS11 sorry about that!  yes combining everything to a string broke the verbose output of pages and broke when there would be multiple pages of granules.  I made the dictionary-to-string into a separate function.  I ran it all the way through and it should be working.  I also tried adding url encoding using either `urllib.parse` or `requests` but it ended up breaking the cookies needed to "stay logged in".  So for now the strings are still not encoded but I haven't found an issue.

JessicaS11: I like the updates to have the dict-to-string as a separate function. That's super good to know re `urllib.parse` and `requests` - a problem for another day then. Where is the granule_name parameter being submitted? When I ran `region_a.order_granules(verbose=True)`, which prints out the URL of the order request, I didn't see it as part of the order submission, and it also doesn't show up in CMR_params either. From looking at `query.py`, it seems like they're being essentially initialized and then set again but never actually added to the `CMRparams` dictionary. Let me know if you want line numbers for what I think is happening.

JessicaS11:
@tsutterley this feels like a ridiculous question, but how do we use this? If I'm understanding correctly, we'd have to instantiate a `query` object with dataset and spatial and temporal extents, then append `readable_granule_name` to the `CMRparams` dictionary, right?

I don't think that's a terrible approach for this PR, but I think we should at least let users know it's an option while @andypbarrett is working to get it usable up-front at init. Let me know if you'd like me to try and work on updating the docs and examples for these new pieces.

tsutterley:
> @tsutterley this feels like a ridiculous question, but how do we use this? If I'm understanding correctly, we'd have to instantiate a `query` object with dataset and spatial and temporal extents, then append `readable_granule_name` to the `CMRparams` dictionary, right?
> 
Fair point.  That is how it is currently written.  I wrote it that way to provide an early check that the RGTs and cycles that the user wanted are actually available with the spatial and temporal options provided (checks with `validate_inputs`).  Could get rid of those portions to limit the number of CMR queries (as many users likely don't care about RGTs and those that do might know the numbers already).

JessicaS11:
> I wrote it that way to provide an early check that the RGTs and cycles that the user wanted are actually available with the spatial and temporal options provided (checks with `validate_inputs`).

I think this check is still important if they submit multiple possible inputs. Maybe the path forward is a few different options, with dataset and temporal always required and supplied by the user in all cases except with granule names. Then they can input one or more of:
1. spatial input
2. RGT input
3. Granule name input
We could use the `init` to check for one (or more) of these inputs. For each one given, make sure it's valid (using `validate_inputs`). If only one is provided, then move on. If more than one is given, do the checks to make sure the information is compatible (again, can be outsourced to `validate_inputs` and only called if needed).

> Could get rid of those portions to limit the number of CMR queries (as many users likely don't care about RGTs and those that do might know the numbers already).
With the above plan, we limit the CMR queries to when they're "needed" for verification, so we reduce the number and still save the user from their own humanness (hopefully) even though they already know the RGTs. ; )

Thoughts? Does this get at your bottom line for this PR/your use case? If so, is this something you can work towards implementing with @andypbarrett (I know we said a separate PR for that work, but this PR is slightly turning into what he planned to do).

tsutterley:
> Thoughts? Does this get at your bottom line for this PR/your use case? If so, is this something you can work towards implementing with @andypbarrett (I know we said a separate PR for that work, but this PR is slightly turning into what he planned to do).

Sounds good :+1:

JessicaS11:
@andypbarrett @tsutterley Any status updates on this? I'm just digging into #166 and it reminded me of this slightly related PR.

tsutterley:
I haven't touched anything since the last time we talked.  To refresh my memory, the goal is to refactor the inputs to be (at least?) one of the following:

1. spatial query
2. RGT query
3. granule query

Is that right?

JessicaS11:
Yeah, exactly. Is there any way I can help facilitate next steps on this?

tsutterley:
@JessicaS11 things should be updated now to enable searching by either spatial+temporal or orbital parameters (or some combination).  If only searching by cycle+track: need to set subsetting to False when ordering since there's nothing to subset.

tsutterley:
> @tsutterley this feels like a ridiculous question, but how do we use this? If I'm understanding correctly, we'd have to instantiate a `query` object with dataset and spatial and temporal extents, then append `readable_granule_name` to the `CMRparams` dictionary, right?
This is changed to no longer query for cycles and tracks.  It only checks that the track number (if entered) is one of the 1387 RGTs and that the cycle number is valid using (today-start)/91 days

icepyx/core/APIformatting.py
```diff
@@ -34,6 +34,9 @@ def _fmt_temporal(start, end, key):
     dictionary with properly formatted temporal parameter for CMR search or subsetting
     """
 
+    if (start is None) or (end is None):
+        return dict()
+
```
JessicaS11, line 39: I puzzled for a few moments over this addition. It's to deal with the fact that for a cycles+tracks search you don't have a temporal parameter. Should we just move "temporal" from "default" to "optional" in our possible keys (`APIformatting._get_possible_keys()`)? We'd still be dictating that the user supply enough info to do a search/order through initialization of the query object, so it just takes it out as a mandatory key then. This all will be moot someday when we're hopefully farming this out to the CMR and/or Harmony APIs, but until then...

icepyx/core/query.py
```diff
@@ -125,6 +125,8 @@ def __init__(
 
         if (
             dataset is None or spatial_extent is None or date_range is None
+        ) and (
+            cycles is None or tracks is None
```
JessicaS11, line 129: To submit a query, the user has to submit a dataset. Then, they can submit cycles and tracks OR date_range and spatial. Do we want to require them to submit spatial along with cycles and tracks (since these are otherwise pretty big)? I think that's the way to go, but am not set on it. Then this would be revised along the lines of `if (dataset is None or spatial_extent is None) and ((cycles is None or tracks is None) or (date_range is None)):`. I think the `files` part isn't yet implemented and that's what @andypbarrett was hoping to add, right?

JessicaS11 (changes requested):

icepyx/core/APIformatting.py
tsutterley: Great point!  That would work.  I just didn't want to break something :sweat_smile:

icepyx/core/query.py
tsutterley: It does work without a spatial parameter, BUT any users would have to make sure `subset` is set to False when they create their order (since there's nothing to subset).  Might be better to just require a spatial parameter :+1:

tsutterley:
> Thanks @tsutterley - this is looking great! Before we merge this PR I'd like to add at least one or two basic tests e.g. for out of range track/cycle requests. Is that something you've got bandwidth for? Otherwise I can try to add some next week.

I can put some things together but it might take a few days.  Happy to coordinate :+1:

icepyx/core/query.py
JessicaS11: They could still subset with variables though... Sounds like we agree on requiring the spatial parameter, since otherwise you can end up with a lot more data than you bargained for (e.g. if you take the spatial parameter out of your example). It's easy enough to specify the whole globe if you really want the entire track.

tsutterley: right I forget about the variable subsetting features.  Yep I think that adding back the spatial requirement sounds good :+1:

setup.py
```diff
@@ -29,7 +29,7 @@
     use_scm_version={
         "fallback_version": "unknown",
         "local_scheme": "node-and-date",
-        "write_to": "_version.py",
+        "write_to": "icepyx/core/_version.py",
```
tsutterley, line 32: @weiji14 I needed this to be the write path for pip installations.  does this work on your end?

tsutterley:
@JessicaS11 I added back the spatial requirements, and added tests for out of bounds cycles or tracks.
I also need someone to check that I didn't break the _version.py functionality from a different PR (needed to change for pip installations)

icepyx/__init__.py
```diff
@@ -1,3 +1,3 @@
 from icepyx.core.query import Query
 
-from _version import version as __version__
+from icepyx.core._version import version as __version__
```
weiji14, line 3:
```suggestion
from _version import version as __version__
```
@tsutterley, maybe you need to revert the change here?

weiji14: Ah yes, I see what you mean. But I think this problem should be fixed once we merge #175 (though I'll double check to be sure). See also https://github.com/icesat2py/icepyx/pull/172#discussion_r577778898 (that PR was merged into #175).

tsutterley: Sounds good.  I'll revert these changes and let them be handled in #175

weiji14: #175 has been merged into 'development'. I've pulled in the changes locally into this `rgt` branch and `pip install` seems to work, i.e.:

```
git switch rgt
git merge development
conda activate icepyx
pip install -e .
python -c "import icepyx as ipx; print(ipx.__version__)"
# 0.3.2b1.dev17+g336779c
```
Could you try and see if it works on your side?

tsutterley: yep! the changes in #175 work (also doesn't have to be installed in "editable" mode so the changes should work for people installing with pypi once they are released)

icepyx/core/validate_inputs.py
```diff
@@ -29,70 +29,94 @@ def dset_version(latest_vers, version):
...
+    # get the total number of seconds since the start of ATLAS and now
+    delta_time_epochs = (epoch2 - epoch1).total_seconds()
+    atlas_UNIX_start_time = (atlas_gps_start_time - delta_time_epochs)
+    present_time = dt.datetime.now().timestamp()
+    # divide total time by cycle length to get the maximum number of orbital cycles
+    ncycles = np.ceil((present_time - atlas_UNIX_start_time)/(86400*91)).astype('i')
+    all_cycles = [str(c+1).zfill(cycle_length) for c in range(ncycles)]
+
+    if cycle is None:
+        return all_cycles
```
JessicaS11, line 52: What is the incentive for returning `all_cycles` here versus an empty list? If the user doesn't specify cycles (or tracks), then this function wouldn't be called, and `region_a.cycles` doesn't get set. Currently, if no time is set during initialization then the property (`region_a.dates`) is unavailable (I just made that change to prevent errors). I think we'll decrease confusion if we keep the behavior consistent across unused parameters. Compelling arguments for one route or the other encouraged. : )

icepyx/core/APIformatting.py
```diff
@@ -161,6 +212,35 @@ def combine_params(*param_dicts):
...
+    """
+    param_list = []
+    for k,v in params.items():
+        if isinstance(v,list):
+            for l in v:
+                param_list.append(k+'='+l)
+        else:
+            param_list.append(k+'='+str(v))
+    # return the parameter string
+    return "&".join(param_list)
```
JessicaS11, line 243: This function is currently untested. I can try to add a test for it later this week.

icepyx/core/validate_inputs.py
tsutterley: Good point!  It simplified making the `readable_granule_name` list but you're right.  I'll put in a change to just add the appropriate wildcard if either cycles or tracks is excluded.

icepyx/core/APIformatting.py
tsutterley: I'll make a test from the example I put in the docstring :)

doc/source/community/resources.rst
```diff
@@ -91,7 +91,7 @@ Here we describe a selection of publicly available Python code posted on GitHub
 
 - `Nsidc-subsetter <https://github.com/tsutterley/nsidc-subsetter>`_
 
-  - by Tyler Sutterly
+  - by Tyler Sutterley
```
JessicaS11, line 94: Sorry!

tsutterley: No worries! :+1:

JessicaS11:
This looks about ready to go to me! Thanks for all your hard work and updates, @tsutterley! @weiji14 or @andypbarrett, can you give a re-review to make sure we didn't miss anything?

JessicaS11 (approved):

examples/ICESat-2_DAAC_DataAccess_Example.ipynb
```diff
@@ -184,26 +178,27 @@
...
+    "There are also several optional inputs to allow the user finer control over their search. Start and end time are only valid inputs on a temporally limited search.\n",
+    "- `start_time` = start time to search for data on the start date. If no input is given, this defaults to 00:00:00.\n",
+    "- `end_time` = end time for the end date of the temporal search parameter. If no input is given, this defaults to 23:59:59. Times must be input as 'HH:mm:ss' strings.\n",
+    "- `version` = What version of the dataset to use, input as a numerical string. If no input is given, this value defaults to the most recent version of the dataset specified in `short_name`.\n",
+    "\n",
+    "*NOTE* Version 001 is used as an example in the below cell. However, using it will cause 'no results' errors in granule ordering for some search parameters. These issues have been resolved in later versions of the datasets, so it is best to use the most recent version where possible. Thus, you will need to change the version associated with `region_a` and rerun the next cell for the rest of this notebook to run."
```
weiji14, line 201: Looks like ATL06 version 002 is used below.

```suggestion
    "*NOTE* Version 002 is used as an example in the below cell. However, using it will cause 'no results' errors in granule ordering for some search parameters. These issues have been resolved in later versions of the datasets, so it is best to use the most recent version where possible. Thus, you will need to change the version associated with `region_a` and rerun the next cell for the rest of this notebook to run."
```

weiji14 (approved):
Exciting! I've tested this branch at [Binder](https://binder.pangeo.io/v2/gh/icesat2py/icepyx/rgt) (not very extensively though) and the by Cycle/RGT query seems to work (even tried ATL11 :smiling_imp:). Just one typo to fix, otherwise I'm happy to approve. There's actually a few stylistic things I could nitpick on but I won't hold this up since they're non-user facing.

examples/ICESat-2_DAAC_DataAccess_Example.ipynb
JessicaS11: Haha - nice catch. I just spotted this this morning and added it to my to-fix list!

PR #150, opened 2020-11-05
add icepyx usage tracking (tools and docs)

JessicaS11:
I've added a few different tools for tracking and sharing icepyx usage data:
- new GitHub Action to pull the GitHub traffic data, store it, plot it, and commit it to a "traffic" branch (will ultimately need to be manually moved to the proper directory and the the branch merged for now). This action is done weekly via cron.
- new runnable notebook to download, store, and plot icepyx download statistics from PyPI
- bib file and framework for adding bibtex entries for publications that leverage icepyx
- new docs page on tracking which displays the above information


doc/source/tracking/tracking.rst
```diff
@@ -0,0 +1,48 @@
+Tracking icepyx Usage
+======================
+
+How is icepyx being used by the ICESat-2 data user community?
+
+
+Projects and Organizations
+--------------------------
+Projects and organizations that use icepyx (this list is still being populated as of Nov 2020)
```
lheagy, line 9: This is great! Would you like to also include a statement encouraging folks to add their organization / a link to their project if they are using icepyx?

doc/source/tracking/icepyx_pubs.bib
```diff
@@ -0,0 +1,41 @@
...
+  month        = aug,
+  year         = 2020,
+  publisher    = {Zenodo},
+  version      = {1.0.0},
+  doi          = {10.5281/zenodo.3966463},
+  url          = {https://doi.org/10.5281/zenodo.3966463},
+ comment	= {tutorials}
+}
+
+@misc{js2019agu,
```
lheagy, line 34: here, I am wondering if it would be worth creating a zenodo archive of the poster and using that doi? that way the poster itself is a bit more discoverable

doc/source/tracking/tracking.rst
```diff
@@ -0,0 +1,48 @@
...
+ICESat-2 peer-reviewed research that utilizes icepyx and presentations that feature or explain icepyx
+
+.. bibliography:: icepyx_pubs.bib
+    :style: mystyle
+    :all:
+    
+    
+Downloads
+---------
+Estimating usage of open-source software is a fundamentally difficult task, and "easy" metrics like number of downloads `have the potential to be misleading <https://blog.dask.org/2020/01/14/estimating-users>`_. For icepyx, the number of library downloads further cannot completely capture usage because it is likely downloaded once and then used for multiple data download or analysis sessions.
```
lheagy, line 29: I would hesitate a bit with the second statement here. The pattern of "download + multiple use" is common to most software packages, so in my opinion, it is more straight-forward to leave this with the first sentence that points to why metrics like downloads are imperfect. But just my 2 cents :)

lheagy (commented):
This looks good @JessicaS11! I left a couple comments, but just my 2 cents, so feel free to take what you think works and ignore what doesn't

doc/source/tracking/icepyx_pubs.bib
JessicaS11: Is the ESSOAR archive not open access, or just not as discoverable as Zenodo? I think discoverability is a good point, though I wonder if at this point it just makes the most sense to plan on doing this with this year's AGU presentation.

JessicaS11:
Thanks, @lheagy. Great suggestions! Unless @aaarendt or @fperez have any comments, we can probably go ahead and approve/merge the PR. Getting this into development (default branch) should enable the traffic action to run as scheduled next weekend ([cron jobs aren't triggered on non-default branches](https://stackoverflow.com/questions/63436541/github-action-workflow-schedule-not-working-on-non-default-branch)).

aaarendt:
@JessicaS11 I'm sorry I won't have time to review this so I think you should proceed with merging.

lheagy (approved):

PR #153, opened 2020-11-25
removed normal proj references due to change in capabilities.xml 

asteiker:
After fixing local fork issues, this should now show only the changes I made to is2ref.py. CMR is down due to a major AWS outage so once again I couldn't test this but like i said in my other PR, these changes have already been implemented in one of our NSIDC notebooks so this should resolve the problem due to the capabilities.xml updates in our last API release.

### DIFF
--- a/icepyx/core/is2ref.py
+++ b/icepyx/core/is2ref.py
@@ -81,7 +81,7 @@ def _get_custom_options(session, dataset, version):
     # reprojection only applicable on ICESat-2 L3B products.
 
     # reprojection options
-    projections = [Projection.attrib for Projection in root.iter('Projection')]
+    projections = [Projection.attrib for Projection in root.iter("Projection")]
     proj_vals = []
     for i in range(len(projections)):
         if (projections[i]['value']) != 'NO_CHANGE' :
--- a/icepyx/core/is2ref.py
+++ b/icepyx/core/is2ref.py
@@ -77,29 +77,32 @@ def _get_custom_options(session, dataset, version):
     format_vals = [formats[i]["value"] for i in range(len(formats))]
     format_vals.remove("")
     cust_options.update({"fileformats": format_vals})
-
-    # reprojection only applicable on ICESat-2 L3B products, yet to be available.
-
-    # reformatting options that support reprojection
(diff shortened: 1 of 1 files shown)


JessicaS11 (changes requested):
Hey @asteiker! This looks great and I was able to run an example notebook with no issues. Would you be able to run `pip install pre-commit` on your local system, then re-commit the file to your fix-proj fork (then it should automatically add the updates to this PR)? If git complains about no files changed, you may need to make a minor modification to the file to get it to detect the changes. What I'm trying to have happen is the pre-commit trigger for running black, which is a code formatter. The first time you try to commit, it should say that black failed and then reformat the file. You'll have to then run the commit a second time, and it should all work.
image
This will help keep the code-base consistently formatted so we don't have lines (e.g. 93(orig)/84(updated) - I'll point them out in-line) that show up as changes even though the change was from ' to " around a string, etc. I can't push the changes to your PR since it's on your personal branch.

icepyx/core/is2ref.py
```diff
@@ -77,30 +77,33 @@ def _get_custom_options(session, dataset, version):
...
-    format_proj = normalproj_vals[0].split(",")
-    format_proj.remove("")
-    format_proj.append("No reformatting")
-    cust_options.update({"formatreproj": format_proj})
+    
+    # reprojection only applicable on ICESat-2 L3B products.
 
     # reprojection options
-    projections = [Projection.attrib for Projection in root.iter("Projection")]
+    projections = [Projection.attrib for Projection in root.iter('Projection')]
```
JessicaS11, line 84: formatting inconsistencies that are fixed by black during the pre-commit run to make it clearer where code has been changed in the diff view (i.e. what we're looking at)

asteiker:
@JessicaS11 Sorry for this issue! When I ran `pip install pre-commit` locally it says that it is already installed, which is strange. I ended up just changes those quotes back to manually fix the problem. I hope this somewhat solves the issue. Not sure if you have any other thoughts for why that trigger didnt happen even though I seem to have `pre-commit` already installed?

JessicaS11 (approved):

PR #168, opened 2021-02-02
Add automatic versioning and required files

lsetiawan:
## Overview

This PR add automatic versioning. This also adds requirements.txt as well as LICENSE file to source package since it's missing and causes https://github.com/conda-forge/staged-recipes/pull/13866 to fail.


lsetiawan:
@JessicaS11 Please review and merge :smile:

setup.py
```diff
@@ -8,7 +9,8 @@
 
 setuptools.setup(
     name="icepyx",
-    version="0.3.2",
+    version=versioneer.get_version(),
+    cmdclass=versioneer.get_cmdclass(),
```
weiji14, line 13: Hi @lsetiawan, just a quick comment if you don't mind. Is it possible to use [setuptools_scm](https://github.com/pypa/setuptools_scm) instead of [versioneer](https://github.com/python-versioneer/python-versioneer) to set the version? This will keep the `icepyx` top directory cleaner. See https://github.com/psf/black/pull/1008 for an example PR.

```suggestion
    use_scm_version={"local_scheme": "node-and-date"},
```

lsetiawan: I haven't used setuptools_scm before, but I'll try it out. Seems a lot slimmer! Thanks for providing this suggestion and example :smile:

weiji14: Looks good! Less files to worry about down the track :smile:

.gitignore
```diff
@@ -123,3 +123,6 @@ venv.bak/
 !views.csv
 !downloads_data.csv
 !sys_downloads_data.csv
+
+# setuptools_scm
+_version.py
```
weiji14, line 128: How about moving this line under the `# Distribution / packaging` section?

requirements.txt
```diff
@@ -1,9 +1,7 @@
-datetime
 fiona
 geopandas
 h5py
 matplotlib
 numpy
-pre-commit
```
weiji14, line 7: Just double checking that these two removals are ok? I think the `datetime` removal is fine, but unsure about `pre-commit` which was added in #96.

weiji14 (commented):
Hopefully @JessicaS11 can have a look before merging. This PR looks pretty much ok to me aside from two minor comments. We should try and tag a v0.3.3 release soon after and get that `icepyx` conda package sorted at https://github.com/conda-forge/staged-recipes/pull/13866!

requirements.txt
lsetiawan: Seems to me the `pre-commit` is more for linting the code, not really for actually functionality.

weiji14: Agree. Might need to list it somewhere else as a development-only dependency though (can be done in a separate PR).

JessicaS11: Looks like pre-commit is already in requirements-dev.txt, albeit with a typo so it reads pre_commit instead...

JessicaS11 (approved):

PR #171, opened 2021-02-10
Add a missing descartes dependency

whyjz:
It looks that (at least on my computer) the descartes package is required (for visualizing query area on map) but not specified in the requirement.txt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ matplotlib
 numpy
 requests
 shapely
+descartes


weiji14:
Hi @whyjz, how are you? Thanks for opening the PR! You're right that `descartes` is required for some things, and is an optional dependency of geopandas v0.8.2 (released on Jan 2021, see https://geopandas.readthedocs.io/en/v0.8.2/install.html#dependencies), but I think geopandas v0.9.0 in the future will not require `descartes` anymore (see https://github.com/geopandas/geopandas/issues/1776#issuecomment-765298341).

Do you want to include this in `icepyx` still (i.e. to remain backward compatible with older versions of geopandas <= 0.8.2)? If yes then I can approve this PR and merge.

whyjz:
Hi @weiji14, doing well here and I hope all is well with you! 
Thanks for pointing that geopandas comment out, and it is good to see they decided to retire the descartes dependency in the next release. With this in mind, I have no preference for whether we should include this backward compatibility. Let me know what you think and I can either close the PR or leave it open accordingly.

weiji14:
Not too bad here, should catch up with you sometime :wink:

Anyways, let's leave the `descartes` dependency out. Not sure when the next `geopandas` or `icepyx` release is due, but hopefully users will know to do `pip install descartes` to get the missing dependency as a workaround. Feel free to submit other contributions to `icepyx` though, I'm sure you're working on some pretty cool stuff!

Issue #165, opened 2021-01-25
Add necessary headers/metadata to api requests to facilitate metrics collection at NSIDC


wallinb:
I think adding the 'client-id' header with value 'icepyx' to CMR/EGI requests and downloads would enable NSIDC Ops to identify those requests as from icepyx and subsequently provide usage metrics back to icepyx.

JessicaS11:
Thanks for opening this issue, @wallinb, and being willing to take this on. From some emails I have, it looks like the exact syntax is expected to be: `client_string=icepyx`. I'll defer to your "insider" knowledge if that's no longer the case though.

In having a quick look through the code base, it looks like we can add it as a [required parameter](https://github.com/icesat2py/icepyx/blob/944673723dcf8c5ac498cc9c8ac416583d10e8b1/icepyx/core/APIformatting.py#L244) to the APIformatting module pretty easily. I think the outstanding question is whether it should get attached to both search and order requests. It will also need to be added to the relevant unit test.

wallinb:
Happy to help! I was going off the the CMR documentation for 'client_id':

> 
Client-Id - Indicates a name for the client using the CMR API. Specifying this helps Operations monitor query performance per client. It can also make it easier for them to identify your requests if you contact them for assistance.

But it may be different for the ordering API and I think we would want to have both. I'm checking with our ops to be sure.

JessicaS11:
>  I'm checking with our ops to be sure.

Awesome - thanks! I agree that both sounds like the way to go.

wallinb:
Indeed `client_string=icepyx` as a url parameter for EGI requests is correct (and `Client-Id` header for CMR).  

I'm happy to relay metrics from NSIDC on EGI requests, CMR metrics are a bit harder as they are collected somewhere in NASA but I can look into it.

PR #173, opened 2021-02-25
Add header/parameters to enable metrics collection at NSIDC

wallinb:
Adds 'Client-Id' header to CMR requests and 'client_string' to EGI requests to enable metrics collection upstream for those APIs. I am happy to relay those metrics from NSIDC as desired.

Issue #165

### DIFF
--- a/icepyx/core/APIformatting.py
+++ b/icepyx/core/APIformatting.py
@@ -251,6 +251,7 @@ class Parameters:
                     "token",
                     "email",
                     "include_meta",
+                    "client_string",
                 ],
             }
         elif self.partype == "subset":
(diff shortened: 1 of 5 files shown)


icepyx/core/APIformatting.py
```diff
@@ -352,7 +353,7 @@ def build_params(self, **kwargs):
             self._check_valid_keys()
 
         if self.partype == "required":
-            if self.check_req_values == True and kwargs == {}:
+            if self.check_req_values() and kwargs == {}:
```
wallinb, line 356: Since `check_req_values` is a method this would never evaluate to True. I think this was the intention?

JessicaS11: Thanks for catching and fixing!

JessicaS11 (approved):

Issue #170, opened 2021-02-04
Releasing v0.3.3 to fix conda-forge build issue


lsetiawan:
## Overview

After merging https://github.com/icesat2py/icepyx/pull/168. I think it's good to release a minor version 0.3.3 to resolve https://github.com/conda-forge/staged-recipes/pull/13866. Since requirements.txt and LICENSE files are missing from the package.

Build error for conda package can be found [here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=272717&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=590).

lsetiawan:
Seems like the recent use of `setuptools_scm` is causing versioning error on the pypi side, maybe the format of dev versions need to change: https://travis-ci.com/github/icesat2py/icepyx/builds/216001973#L761

weiji14:
Oops. So we need to use the `no-local-version` scheme when uploading to PyPI. See documentation at https://github.com/pypa/setuptools_scm#version-number-construction and https://github.com/GenericMappingTools/pygmt/pull/695/files#diff-75e8bbdd253f4a71a14114808840e63b6a9c23529499687d148025d860336504R45 for an example on how to fix the CI.

JessicaS11:
When you say requirements.txt and LICENSE files are missing, do you mean in the recipe (or PyPI build that the recipe is pulling from), since they exist in the repo itself? For my own future reference, if these files are updated in the icepyx repo, will I need to also make those changes to the conda-forge recipe?

Does `setuptools_scm` automatically do a new PyPI build every time something is pushed or merged? I was surprised to see that being the cause of the failed Travis build because pushes/PRs to the development branch should only trigger a PyPI_test build, and only if triggered in the commit message.

Thanks for the links for CI fixes. It's getting pretty late here for today but I'll work on this tomorrow.

lsetiawan:
> When you say requirements.txt and LICENSE files are missing, do you mean in the recipe (or PyPI build that the recipe is pulling from), since they exist in the repo itself? 

[...]

PR #174, opened 2021-02-28
Setup Github Action for publishing to PyPI and TestPyPI

weiji14:
Publishing the icepyx package to PyPI on every tagged (published) release. Every push to the development branch is also published to TestPyPI for testing purposes. Addresses https://github.com/icesat2py/icepyx/issues/170#issuecomment-786299408.

Based on https://github.com/GenericMappingTools/pygmt/pull/679 (and subsequent battleproofed patches).

References:
- https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
- https://github.com/pypa/gh-action-pypi-publish/tree/v1.4.2#usage
- https://pypi.org/project/icepyx/#history
- https://test.pypi.org/project/icepyx/#history

P.S. @JessicaS11, could you please set some secrets at https://github.com/icesat2py/icepyx/settings/secrets/actions called `PYPI_API_TOKEN` and `TEST_PYPI_API_TOKEN` following https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github? Ping me when you'll done and I'll test that uploading to TestPyPI works. Happy to get added on TestPyPI/PyPI too if it helps.

### DIFF
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -9,10 +9,8 @@ on:
   push:
     branches:
       - development
-  # Runs for pull requests should be disabled other than for testing purposes
-  pull_request:
-    branches:
-      - development
+  # Trigger manually at https://github.com/icesat2py/icepyx/actions/workflows/publish_to_pypi.yml
+  workflow_dispatch:
 
 jobs:
   publish-pypi:
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -1,0 +1,64 @@
+# Publish archives to PyPI and TestPyPI using GitHub Actions
+name: Publish to PyPI
+
+# Only run for tagged releases and pushes to the development branch
+on:
+  release:
+    types:
+      - published
+  push:
+    branches:
+      - development
+  # Runs for pull requests should be disabled other than for testing purposes
+  pull_request:
+    branches:
+      - development
+
+jobs:
+  publish-pypi:
+    name: Publish to PyPI
+    runs-on: ubuntu-latest
+    if: github.repository == 'icesat2py/icepyx'
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2.3.4
+      with:
+        # fetch all history so that setuptools-scm works
+        fetch-depth: 0
+
+    - name: Set up Python
+      uses: actions/setup-python@v2.2.1
+      with:
+        python-version: 3.9
+
+    - name: Install dependencies
+      run: python -m pip install setuptools wheel
+
+    # This step is only necessary for testing purposes and for TestPyPI
+    - name: Fix up version string for TestPyPI
+      if: ${{ !startsWith(github.ref, 'refs/tags') }}
+      run: |
+        # Change setuptools-scm local_scheme to "no-local-version" so the
+        # local part of the version isn't included, making the version string
+        # compatible with PyPI.
+        sed --in-place "s/node-and-date/no-local-version/g" setup.py
+
+    - name: Build source and wheel distributions
+      run: |
+        python setup.py sdist bdist_wheel
+        echo ""
+        echo "Generated files:"
+        ls -lh dist/
+
+    - name: Publish to Test PyPI
+      uses: pypa/gh-action-pypi-publish@v1.4.2
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+
+    - name: Publish to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@v1.4.2
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,24 +21,3 @@ jobs:
     #   script:
     #     - export NSIDC_LOGIN=$NSIDC_LOGIN
     #     - pytest icepyx/tests/behind_NSIDC_API_login.py
-
-deploy:
-  - provider: pypi
-    server: https://test.pypi.org/legacy/ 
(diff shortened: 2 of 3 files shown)


.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,64 @@
...
+  release:
+    types:
+      - published
+  push:
+    branches:
+      - development
+  # Runs for pull requests should be disabled other than for testing purposes
+  pull_request:
+    branches:
+      - development
```
weiji14, line 15: Note to self: Comment these lines out before this PR is merged, as they are for testing purposes only (check that publishing to TestPyPI works.

```suggestion
  # pull_request:
  #   branches:
  #     - development
```

.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,64 @@
...
+      run: python -m pip install setuptools wheel
+
+    # This step is only necessary for testing purposes and for TestPyPI
+    - name: Fix up version string for TestPyPI
+      if: ${{ !startsWith(github.ref, 'refs/tags') }}
+      run: |
+        # Change setuptools-scm local_scheme to "no-local-version" so the
+        # local part of the version isn't included, making the version string
+        # compatible with PyPI.
+        sed --in-place "s/node-and-date/no-local-version/g" setup.py
```
weiji14, line 45: Just to explain this section a little bit, what `sed` does is to replace `"local_scheme": "node-and-date"` in setup.py with `"local_scheme": "no-local-version"` on **non-tagged** pushes. So:

- On regular pushes to 'development', "no-local-version" is used which means a version string like `0.3.2.dev12`
- On proper tagged/published releases, "node-and-date" is used which means a version string like `0.3.3`

References:
- https://github.com/pypa/setuptools_scm/tree/v5.0.1#version-number-construction

.github/workflows/publish_to_pypi.yml
JessicaS11: Is there a way to have this triggered if we put a certain keyword in a commit message to the development branch? I think that would be a clearer way to trigger test builds to PyPI than having to comment out these lines each time.

.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,64 @@
...
+      uses: pypa/gh-action-pypi-publish@v1.4.2
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+
+    - name: Publish to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@v1.4.2
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
JessicaS11, line 64: Am I understanding correctly that the steps are executed in order, so if the publish to Test PyPI failed, it wouldn't try to publish to PyPI?

JessicaS11:
Pinging @weiji14 to let you know I've added the `PYPI_API_TOKEN` and `TEST_PYPI_API_TOKEN` to the repo. I'd like to be more explicit in our triggering of the test pypi builds (versus needing to comment and uncomment lines), but otherwise this looks great and almost ready to merge! Thanks for putting it together!

.github/workflows/publish_to_pypi.yml
weiji14: Yes, there are ways to trigger builds manually [using slash commands](https://github.com/marketplace/actions/slash-commands) (e.g. `/push-to-testpypi`) but it involves a bit more work.

I think it might be easier to have a 'Run workflow' button to press instead (see https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). I'll try and set it up and test if it'll work to push to TestPyPI.

.github/workflows/publish_to_pypi.yml
weiji14: Yes, failing to publish to TestPyPI will mean the subsequent publish to PyPI step won't run.

weiji14:
> Pinging @weiji14 to let you know I've added the `PYPI_API_TOKEN` and `TEST_PYPI_API_TOKEN` to the repo. I'd like to be more explicit in our triggering of the test pypi builds (versus needing to comment and uncomment lines), but otherwise this looks great and almost ready to merge! Thanks for putting it together!

Ok, I think we should be able to trigger builds manually now with the push of a button using workflow_dispatch (see https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch) as of 1222a031f92bb4338428f9bdc47885a6efcceee2. But might need to merge things this PR into the development branch before we can test it out :sweat_smile:

.github/workflows/publish_to_pypi.yml
JessicaS11: > I think it might be easier to have a 'Run workflow' button to press instead

Awesome! That sounds great to me.

Thanks for clarifying - I meant test builds to test-pypi but did not say that clearly.

JessicaS11: Can we specify that it should only do a test pypi build when it's manually triggered? Otherwise, if the test build is successful it will automatically go on to do the PyPI build, even if it's not a tagged release. Or am I misunderstanding something?

weiji14: Good question. The publishing to PyPI won't be done on manual triggers, it will only be done on published releases because of an `if` clause here:

https://github.com/icesat2py/icepyx/blob/1222a031f92bb4338428f9bdc47885a6efcceee2/.github/workflows/publish_to_pypi.yml#L58-L60

In other words, this publish to PyPI step is only run on tagged pushes.

JessicaS11: I somehow managed to not register that. Thanks for pointing it out!

JessicaS11 (approved):

PR #175, opened 2021-03-05
add manifest and update requirements files pre-conda-forge release

JessicaS11:


JessicaS11:
Just realized that we'll need these updates before we can release 0.3.3 (#170)!

setup.py
```diff
@@ -26,10 +26,11 @@
         "Operating System :: OS Independent",
         "Development Status :: 1 - Planning",
     ],
+    py_modules=["_icepyx_version"],
     use_scm_version={
         "fallback_version": "unknown",
-        "local_scheme": "node-and-date",
-        "write_to": "_version.py",
+        "local_scheme": "no-local-version",
```
weiji14, line 32:
```suggestion
        "local_scheme": "node-and-date",
```
Need to keep this as node-and-date because the publish to pypi Github Action workflow expects it.

weiji14 (commented):
Just one minor suggestion. Otherwise double check with @lsetiawan ;)

weiji14 (approved):

PR #176, opened 2021-03-07
use geoviews to visualize spatial extent interactively

icetianli:
Use geoviews to visualize spatial extent, now supports interactive exploration on geospatial extent:

```
short_name = 'ATL06'
date_range = ['2019-10-01','2019-10-05'] 
bbox = [(-55, 68), (-55, 71), (-48, 71), (-48, 68), (-55, 68)]
region_a = ipx.Query(short_name, bbox, date_range)
region_a.visualize_spatial_extent()
```
create map:
image


requirements.txt
```diff
@@ -1,4 +1,5 @@
 fiona
+geoviews
 geopandas
```
weiji14, line 3: This file is sorted alphabetically

```suggestion
geopandas
geoviews
```

icepyx/core/query.py
```diff
@@ -917,10 +917,12 @@ def visualize_spatial_extent(
...
-        geospatial.geodataframe(self.extent_type, self._spat_extent).plot(
-            ax=ax, color="#FF8C00", alpha=0.7
-        )
-        plt.show()
+        from shapely.geometry import Polygon
+        import geoviews as gv
+        gv.extension('bokeh')
+
+        gdf = geospatial.geodataframe(self.extent_type, self._spat_extent)
+        line_geoms = Polygon(gdf['geometry'][0]).boundary
```
weiji14, line 925: My understanding is that `self._spat_extent` is already a `shapely.geometry.Polygon` object so it should be possible to get the outline boundary directly like this:

```suggestion
        line_geoms = self._spat_extent.boundary
```

icepyx/core/query.py
```diff
@@ -917,10 +917,12 @@ def visualize_spatial_extent(
         [visual map output]
         """
 
-        world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
-        f, ax = plt.subplots(1, figsize=(12, 6))
-        world.plot(ax=ax, facecolor="lightgray", edgecolor="gray")
-        geospatial.geodataframe(self.extent_type, self._spat_extent).plot(
-            ax=ax, color="#FF8C00", alpha=0.7
-        )
-        plt.show()
```
weiji14, line 926: Wondering if we should keep this old code as a fallback in case people don't have [`geoviews`](https://geoviews.org) installed. Apparently `pip install geoviews` doesn't work out of the box (see https://geoviews.org/index.html#additional-dependencies) because it needs `PROJ` (which can only be `conda` installed).

Maybe have something like a try...except block or add a parameter (`interactive=True` or `interactive=False`) so that people can choose whether to use an interactive `geoviews` or static `geopandas.plot`? What do you think?

weiji14 (commented):
Hi @icetianli, this interactive extent viewer will be a great addition! Just a few suggestions/comments for now :smile:

JessicaS11:
Thanks for this great way to visualize extents, @icetianli! @weiji14 made some great suggestions which I recommend we use. Paying attention to the geoviews install challenges is especially important (many users are still new to creating and updating environments, so this presents an extra hurdle for them). It's also the cause of the Travis CI failure - it needs PROJ to install cartopy, which is a geoviews dependency. I'll try to look into how other packages handle this challenge unless you already have a proposed solution.

icepyx/core/query.py
icetianli: Thanks for the advice! Yeah it's a good idea to keep `geopandas.plot` as a backup in case some users can't use geoviews. I think a `try...except` block makes more sense than asking users to add an additional parameter, so if they can use geoviews then it will generate an interactive map, otherwise it will just show a static map.

icepyx/core/query.py
icetianli: Good point! Just checked the `Query` class and found out that `self._spat_extent` is a `shapely.geometry.Polygon` object only when the input `spatial extent` are polygon vertices and polygon file, it will become a list for a bounding box input, so I just kept `Polygon(gdf['geometry'][0]).boundary` for the plot.

icepyx/core/query.py
```diff
@@ -916,11 +916,23 @@ def visualize_spatial_extent(
...
+            tile = gv.tile_sources.EsriImagery.opts(width=500, height=500)
+            return tile * bbox_poly
+
+        except:
+            world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
+            f, ax = plt.subplots(1, figsize=(12, 6))
+            world.plot(ax=ax, facecolor="lightgray", edgecolor="gray")
+            gdf.plot(
+                ax=ax, color="#FF8C00", alpha=0.7
+            )
```
weiji14, line 937: This can fit on one line now.
```suggestion
            gdf.plot(ax=ax, color="#FF8C00", alpha=0.7)
```

icepyx/core/query.py
```diff
@@ -916,11 +916,23 @@ def visualize_spatial_extent(
...
+            from shapely.geometry import Polygon
+            import geoviews as gv
+            gv.extension('bokeh')
+
+            line_geoms = Polygon(gdf['geometry'][0]).boundary
+            bbox_poly = gv.Path(line_geoms).opts(color="red", line_color="red")
+            tile = gv.tile_sources.EsriImagery.opts(width=500, height=500)
+            return tile * bbox_poly
+
+        except:
```
weiji14, line 931: Should we be a bit more specific here perhaps and use ImportError (i.e. fail to load `geoviews`)?

```suggestion
        except ImportError:
```

icepyx/core/query.py
```diff
@@ -916,11 +916,23 @@ def visualize_spatial_extent(
...
-        f, ax = plt.subplots(1, figsize=(12, 6))
-        world.plot(ax=ax, facecolor="lightgray", edgecolor="gray")
-        geospatial.geodataframe(self.extent_type, self._spat_extent).plot(
-            ax=ax, color="#FF8C00", alpha=0.7
-        )
-        plt.show()
+        try:
+            from shapely.geometry import Polygon
+            import geoviews as gv
+            gv.extension('bokeh')
```
weiji14, line 924: The `black` linter suggests adding a line between `import` statements and the line below.
```suggestion
            from shapely.geometry import Polygon
            import geoviews as gv

            gv.extension('bokeh')
```

icepyx/core/query.py
icetianli: yes agree!

requirements.txt
```diff
@@ -1,5 +1,6 @@
 fiona
 geopandas
+geoviews
```
weiji14, line 3: > Paying attention to the geoviews install challenges is especially important (many users are still new to creating and updating environments, so this presents an extra hurdle for them). It's also the cause of the Travis CI failure - it needs PROJ to install cartopy, which is a geoviews dependency. I'll try to look into how other packages handle this challenge unless you already have a proposed solution.

Installing `PROJ` into Travis CI will be a bit involved. Either we install it using `conda` (see https://docs.conda.io/projects/conda/en/4.6.1/user-guide/tasks/use-conda-with-travis-ci.html#the-travis-yml-file), or via something like `apt` (https://docs.travis-ci.com/user/languages/python/#dependency-management). This PR looks good, but we might have to wait to handle this after release v0.3.3 (#170), and it might be best to add `geoviews` as a dependency in a separate Pull Request to keep things manageable.

Locally, I've been using `conda create --name icepyx --file requirements.txt` (to create a fresh `icepyx` conda environment) to test, rather than `pip install requirements.txt`. Will need to figure out how to translate that to work on `.travis.yml` (read: may need to setup miniconda first).

JessicaS11: I've added an issue for further discussion on this: #177

weiji14: Ok, I've opened a separate PR at #193 to handle the `geoviews` installation as an optional dependency in Travis CI. Once that is reviewed (feel free to give a go @icetianli) and approved/merged, we should be able to remove this `geoviews` line from requirements.txt and (finally) merge in this PR!

```suggestion
```

JessicaS11: @icetianli Did you want to give PR #193 a review, or should I plan to do that? I'd like to move forward with merging this PR and I think that's the only holdup.

icetianli: Thanks @weiji14 and @JessicaS11 for putting this together, I'll give it a go today!

JessicaS11:
I'd like to add a note to the example about which visualization to expect (interactive or not) based on whether you've installed geoviews or not. I can take care of that sometime this week, and then this PR will be good to go!

JessicaS11 (approved):

PR #181, opened 2021-03-11
Adding extra docs requirements and tweak conf

lsetiawan:
## Overview

This PR is my attempt in fixing the read the docs issue that's happening. I grabbed an example from https://github.com/django-auth-ldap/django-auth-ldap/blob/master/docs/conf.py

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -33,4 +33,4 @@ jobs:
           python -m pip install -r "requirements-docs.txt"
           python -m pip install -e "."
       - name: Build documentation
-        run: sphinx-build -a -b html -W doc/source/ doc/build/
+        run: make --directory=doc html
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -8,7 +8,7 @@ on:
   pull_request:
     paths:
       - "doc/**"
-      - - "**/doc.yaml"
+      - "**/doc.yaml"
 
 jobs:
   build:
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -1,0 +1,36 @@
+name: Documentation Build
+
+on:
+  push:
+    paths:
+      - "doc/**"
+  pull_request:
+    paths:
+      - "doc/**"
+
+jobs:
+  build:
+    # We want to run on external PRs, but not on our own internal PRs as they'll be run
+    # by the push to the branch. Without this if check, checks are duplicated since
+    # internal PRs match both the push and pull_request events.
+    if:
+      github.event_name == 'push' || github.event.pull_request.head.repo.full_name !=
+      github.repository
+
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python 3.7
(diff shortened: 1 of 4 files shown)


weiji14:
What's the code for running the documentation build? Just wanted to test this out locally to see if it works.

weiji14:
Can you try removing the `import icepyx` line. I don't think the `icepyx` package is used in this `conf.py` file.

weiji14:
I noticed that readthedocs.yaml can use `method: setuptools` (see https://docs.readthedocs.io/en/stable/config-file/v2.html#python). Wonder if that might do the trick?

```diff
diff --git a/readthedocs.yml b/readthedocs.yml
index b969696..f77e621 100644
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -22,3 +22,4 @@ python:
   install:
     - requirements: requirements-docs.txt
     - requirements: requirements.txt
+  method: setuptools
```

lsetiawan:
> I noticed that readthedocs.yaml can use `method: setuptools` (see https://docs.readthedocs.io/en/stable/config-file/v2.html#python). Wonder if that might do the trick?
> 
> ```diff
> diff --git a/readthedocs.yml b/readthedocs.yml
> index b969696..f77e621 100644
> --- a/readthedocs.yml
> +++ b/readthedocs.yml
> @@ -22,3 +22,4 @@ python:
>    install:
>      - requirements: requirements-docs.txt
>      - requirements: requirements.txt
> +  method: setuptools
> ```

I was wondering about that. Currently it's actually doing pip install ... it's strange that doesn't work since the pypi test works and that does a pip install ...

JessicaS11:
> > I noticed that readthedocs.yaml can use `method: setuptools` (see https://docs.readthedocs.io/en/stable/config-file/v2.html#python). Wonder if that might do the trick?
> > ```diff
> > diff --git a/readthedocs.yml b/readthedocs.yml
> > index b969696..f77e621 100644
> > --- a/readthedocs.yml
> > +++ b/readthedocs.yml
> > @@ -22,3 +22,4 @@ python:
> >    install:
> >      - requirements: requirements-docs.txt
> >      - requirements: requirements.txt
> > +  method: setuptools
> > ```
> 
> I was wondering about that. Currently it's actually doing pip install ... it's strange that doesn't work since the pypi test works and that does a pip install ...

I agree - odd indeed. I was able to get the docs to build on readthedocs by using this option (see #182). Thoughts or preferences for which approach we use?

lsetiawan:
>I agree - odd indeed. I was able to get the docs to build on readthedocs by using this option (see #182). Thoughts or preferences for which approach we use?

Cool! If using setuptools work, then cool. That's totally fine. We can close this PR in favor of #182. @JessicaS11 Could you bring in the doc workflow to your PR? Thanks! It would be good for us to test the doc build :smile:

JessicaS11:
> Could you bring in the doc workflow to your PR? Thanks! It would be good for us to test the doc build 😄

You and @weiji14 are on the same page (he just said the same thing in #170). I added it to #182 and am waiting for the checks to finish...

JessicaS11:
@lsetiawan Can I delete this branch?

PR #193, opened 2021-03-30
Add geoviews as extra optional dependency in setup.py

### LINKED ISSUE #177: Geoviews as a dependency given install challenges
>> Paying attention to the geoviews install challenges is especially important (many users are still new to creating and updating environments, so this presents an extra hurdle for them). It's also the cause of the Travis CI failure - it needs PROJ to install cartopy, which is a geoviews dependency. I'll try to look into how other packages handle this challenge unless you already have a proposed solution.

>Installing `PROJ` into Travis CI will be a bit involved. Either we install it using `conda` (see https://docs.conda.io/projects/conda/en/4.6.1/user-guide/tasks/use-conda-with-travis-ci.html#the-travis-yml-file), or via something like `apt` (https://docs.travis-ci.com/user/languages/python/#dependency-management). This PR looks good, but we might have to wait to handle this after release v0.3.3 (#170), and it might be best to add `geoviews` as a dependency in a separate Pull Request to keep things manageable.

>Locally, I've been using `conda create --name icepyx --file requirements.txt` (to create a fresh `icepyx` conda environment) to test, rather than `pip install requirements.txt`. Will need to figure out how to translate that to work on `.travis.yml` (read: may need to setup miniconda first).

_Originally posted by @weiji14 in https://github.com/icesat2py/icepyx/pull/176#discussion_r589778204_

Travis CI's role is to run the testing suite prior to merges. Since geoviews (and other potentially fussy-to-install geospatial packages) are not core requirements for software functionality/testing, perhaps the best approach is to add an environment.yml file that includes geoviews, but not include it as a software requirement (since there is another option for visualization).

weiji14:
List [`geoviews`](https://geoviews.org) as an optional dependency in the [setup.py](https://github.com/icesat2py/icepyx/blob/development/setup.py) file. Also set up travis to install [PROJ](https://proj.org) and [GEOS](https://trac.osgeo.org/geos) first so that [`cartopy`](https://scitools.org.uk/cartopy/docs/v0.18/) (a requirement of `geoviews`) can be installed.

```
conda create -n icepyx-env -c conda-forge geos pip proj=7.2.0
conda activate icepyx-env
pip install -e .       # install required dependencies first 
pip install -e .[viz]  # install optional dependencies later
```

Note that PROJ version is pinned to 7.2.0 because PROJ 8.0.0 doesn't seem to be supported by cartopy yet, wait for https://github.com/SciTools/cartopy/pull/1752.

Fixes #177.

References:
- https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies
- https://laszukdawid.com/2017/06/04/installing-cartopy-on-ubuntu-14-04-or-travis-ci/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ with open("README.rst", "r") as f:
 with open("requirements.txt") as f:
     INSTALL_REQUIRES = f.read().strip().split("\n")
 
-EXTRAS_REQUIRE = {"viz": ["geoviews >= 1.9.0"]}
+EXTRAS_REQUIRE = {"viz": ["geoviews >= 1.9.0", "cartopy >= 0.18.0"]}
 EXTRAS_REQUIRE["complete"] = sorted(set(sum(EXTRAS_REQUIRE.values(), [])))
 
 setuptools.setup(
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ with open("README.rst", "r") as f:
 with open("requirements.txt") as f:
     INSTALL_REQUIRES = f.read().strip().split("\n")
 
-EXTRAS_REQUIRE = {"viz": ["geoviews >= 1.9.0", "cartopy >= 0.18.0"]}
+EXTRAS_REQUIRE = {"viz": ["geoviews >= 1.9.0", "cartopy >= 0.18.0", "scipy"]}
 EXTRAS_REQUIRE["complete"] = sorted(set(sum(EXTRAS_REQUIRE.values(), [])))
 
 setuptools.setup(
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,9 @@ with open("README.rst", "r") as f:
 with open("requirements.txt") as f:
     INSTALL_REQUIRES = f.read().strip().split("\n")
 
+EXTRAS_REQUIRE = {"viz": ["geoviews >= 1.9.0"]}
+EXTRAS_REQUIRE["complete"] = sorted(set(sum(EXTRAS_REQUIRE.values(), [])))
+
 setuptools.setup(
     name="icepyx",
     author="The icepyx Developers",
@@ -19,6 +22,7 @@ setuptools.setup(
     license="BSD 3-Clause",
     packages=setuptools.find_packages(exclude=["*tests"]),
     install_requires=INSTALL_REQUIRES,
+    extras_require=EXTRAS_REQUIRE,
     python_requires=">=3",
     classifiers=[
         "Programming Language :: Python :: 3",
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,22 @@
 language: python
 
+before_install:
+  - sudo apt-get update
+  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda
+  # Useful for debugging any issues with conda
+  - conda info -a
+
 install:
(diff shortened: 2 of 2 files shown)


JessicaS11:
Thanks for putting this together, @weiji14! If @icetianli is unable to review, I will do so Friday.

icetianli:
Thank you @weiji14 for this PR! I have been trying to install geoviews as extra dependency using the instructions in your comment: 
```
conda create -n icepyx-env -c conda-forge pip proj=7.2.0
conda activate icepyx-env
pip install -e .[viz]
```
But it keeps showing the following errors about Fiona and Cartopy:
```
ERROR: Command errored out with exit status 1:
    Complete output (2 lines):
    Failed to get options via gdal-config: [Errno 2] No such file or directory: 'gdal-config'
    A GDAL API version must be specified. Provide a path to gdal-config using a GDAL_CONFIG environment variable or use a GDAL_VERSION environment variable.
```

```
fiona/_transform.cpp:606:10: fatal error: 'cpl_conv.h' file not found
  #include "cpl_conv.h"
           ^~~~~~~~~~~~
  1 error generated.
  error: command '/usr/bin/clang' failed with exit code 1
  ----------------------------------------
  ERROR: Failed building wheel for fiona
  Running setup.py clean for fiona
Failed to build cartopy fiona
ERROR: Could not build wheels for cartopy which use PEP 517 and cannot be installed directly
```
But It works perfectly fine if I didn't add `pip` after `conda-forge`:
```
conda create -n icepyx-env -c conda-forge proj=7.2.0
conda activate icepyx-env
pip install -e .[viz]
```
Any idea why? Or is it just a typo...? Thanks!

weiji14:
Hmm, you're right. This is why it is recommended to install geoviews/cartopy/shapely via `conda` because doing it via `pip` is a nightmare (see https://github.com/SciTools/cartopy/issues/738 and https://github.com/SciTools/cartopy/issues/805). But to explain things:

- The second one works because you're probably installing using the `base` environment's python/pip and it is picking up some dependencies installed elsewhere (not recommended though).
- The first one failed (I think) because [`geos`](https://trac.osgeo.org/geos/) needs to be installed too , can you try:

```
git checkout geoviews-dependency
conda create -y -n icepyx-env -c conda-forge geos pip proj=7.2.0
conda activate icepyx-env
pip install -e .[viz]
```

I've added a commit at 6fab43452f6ba23a3ea689aa0ba9a722ec662628 and edited the top post to mention `geos` too. Just to be clear, the `geoviews` install problem shouldn't matter to users who do `conda install icepyx` (as long as we list `geoviews` as a dependency in https://github.com/conda-forge/icepyx-feedstock/blob/master/recipe/meta.yaml). But users doing `pip install icepyx[viz]` might face issues of needing to install both PROJ and GEOS first.

setup.py
```diff
@@ -6,6 +6,9 @@
 with open("requirements.txt") as f:
     INSTALL_REQUIRES = f.read().strip().split("\n")
 
+EXTRAS_REQUIRE = {"viz": ["geoviews >= 1.9.0", "cartopy >= 0.18.0"]}
```
icetianli, line 9: Thanks for helping me understand @weiji14, I figured out the reason why it kept failing was due to incorrect GDAL configuration in my OS (gdal is not in PATH), but it was solved now. The only issue is that`geoviews` requires `scipy` as a dependency,  which will be solved by adding `scipy` in `EXTRAS_REQUIRE`: 
`EXTRAS_REQUIRE = {"viz": ["geoviews >= 1.9.0", "cartopy >= 0.18.0", "scipy"]}`. After this I think we are good to  merge and move onto https://github.com/icesat2py/icepyx/pull/176.

icetianli (changes requested):

setup.py
weiji14: Ok, done in 5be44370d99b99e638b2a428b0b1c68bb40c7e6e

icetianli (approved):

PR #195, opened 2021-04-08
surface NSIDC order and http download errors more effectively to user

### LINKED ISSUE #79: Surface login/service availability errors immediately to the user if they happen
During the 2020 hack week it seems there was a transient error with NSIDC access, but unfortunately icepyx didn't surface that info right at login time, when it probably received the error message. 

Instead, the error info was stored and upon later running a regular line of code, such as
```python
region_a.order_vars.avail(options=True)
```
the following came back:

image (1)

After inspecting that stack frame interactively (with `%debug`), I saw this text was stored internally in the icepyx Data object:

```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>503 Service Unavailable</title>
</head><body>
<h1>Service Unavailable</h1>
<p>The server is temporarily unable to service your
request due to maintenance downtime or capacity
problems. Please try again later.</p>
</body></html>
```

My suspicion is that 503 code may have come in at some point but wasn't surfaced.  Not sure?

I'm happy to try and help further track/debug this one...

JessicaS11:
A few changes in response messages from NSIDC meant that users weren't be alerted when their subset orders did not contain any data in the region of interest (this used to return an error message).

May close #79.

### DIFF
--- a/icepyx/core/granules.py
+++ b/icepyx/core/granules.py
@@ -165,8 +165,7 @@ class Granules:
         granule_search_url = "https://cmr.earthdata.nasa.gov/search/granules"
 
         headers = {"Accept": "application/json", "Client-Id": "icepyx"}
-        # DevGoal: check the below request/response for errors and show them if they're there; then gather the results
-        # note we should also do this whenever we ping NSIDC-API - make a function to check for errors
+        # note we should also check for errors whenever we ping NSIDC-API - make a function to check for errors
         while True:
             params = apifmt.combine_params(
                 CMRparams, {k: reqparams[k] for k in ("page_size", "page_num")}
@@ -178,18 +177,18 @@ class Granules:
             )
 
             results = json.loads(response.content)
-
-            # print(results)
-
+            
             try:
-                if not results["feed"]["entry"]:
-                    # Out of results, so break out of loop
-                    break
-            except KeyError:
-                if "errors" in results.keys():
-                    raise QueryError.NsidcQueryError(
-                        str(results["errors"])
-                    )
+                response.raise_for_status()
+            except requests.HTTPError as e:
+                if b'errors' in response.content:  # If CMR returns a bad status with extra information, display that
+                    raise icepyx.core.exceptions.NsidcQueryError(str(results["errors"])) # exception chaining will display original exception too 
+                else:  # If no 'errors' key, just reraise original exception
+                    raise
+
+            if not results["feed"]["entry"]:
+                # Out of results, so break out of loop
+                break
 
             # Collect results and increment page_num
             self.avail.extend(results["feed"]["entry"])
--- a/icepyx/core/granules.py
+++ b/icepyx/core/granules.py
@@ -50,8 +50,10 @@ def gran_IDs(grans, ids=True, cycles=False, tracks=False):
     """
     assert len(grans) > 0, "Your data object has no granules associated with it"
     # regular expression for extracting parameters from file names
-    rx = re.compile(r'(ATL\d{2})(-\d{2})?_(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})'
-           r'(\d{2})_(\d{4})(\d{2})(\d{2})_(\d{3})_(\d{2})(.*?).(.*?)$')
+    rx = re.compile(
+        r"(ATL\d{2})(-\d{2})?_(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})"
+        r"(\d{2})_(\d{4})(\d{2})(\d{2})_(\d{3})_(\d{2})(.*?).(.*?)$"
+    )
     gran_ids = []
     gran_cycles = []
     gran_tracks = []
@@ -100,6 +102,7 @@ def gran_IDs(grans, ids=True, cycles=False, tracks=False):
         gran_list.append(gran_tracks)
     # return the list of granule parameters
     return gran_list
+
 
 # DevGoal: this will be a great way/place to manage data from the local file system
 # where the user already has downloaded data!
@@ -171,18 +174,20 @@ class Granules:
                 CMRparams, {k: reqparams[k] for k in ("page_size", "page_num")}
             )
             response = requests.get(
-                granule_search_url,
-                headers=headers,
-                params=apifmt.to_string(params),
+                granule_search_url, headers=headers, params=apifmt.to_string(params),
             )
 
             results = json.loads(response.content)
-            
+
             try:
                 response.raise_for_status()
             except requests.HTTPError as e:
-                if b'errors' in response.content:  # If CMR returns a bad status with extra information, display that
-                    raise icepyx.core.exceptions.NsidcQueryError(str(results["errors"])) # exception chaining will display original exception too 
+                if (
+                    b"errors" in response.content
+                ):  # If CMR returns a bad status with extra information, display that
+                    raise icepyx.core.exceptions.NsidcQueryError(
+                        str(results["errors"])
+                    )  # exception chaining will display original exception too
                 else:  # If no 'errors' key, just reraise original exception
                     raise
 
@@ -318,7 +323,10 @@ class Granules:
             esir_root = ET.fromstring(request.content)
             if verbose is True:
                 print("Order request URL: ", requests.utils.unquote(request.url))
-                print("Order request response XML content: ", request.content.decode('utf-8'))
+                print(
+                    "Order request response XML content: ",
+                    request.content.decode("utf-8"),
+                )
 
             # Look up order ID
             orderlist = []
@@ -489,7 +497,7 @@ class Granules:
             if verbose is True:
                 print("Zip download URL: ", downloadURL)
             print("Beginning download of zipped output...")
-            
+
             try:
                 zip_response = session.get(downloadURL)
                 # Raise bad request: Loop will stop for bad response code.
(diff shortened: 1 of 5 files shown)


icepyx/core/granules.py
```diff
@@ -462,25 +490,34 @@ def download(self, verbose, path, session=None, restart=False):
...
 
             # DevGoal: move this option back out to the is2class level and implement it in an alternate way?
             #         #Note: extract the dataset to save it locally
-            # if extract is True:
-            with zipfile.ZipFile(io.BytesIO(zip_response.content)) as z:
-                for zfile in z.filelist:
-                    # Remove the subfolder name from the filepath
-                    zfile.filename = os.path.basename(zfile.filename)
-                    z.extract(member=zfile, path=path)
+            if success is True:
```
bwallin, line 515: Instead of the `success` variable, you could add an `else` block in the try-catch:

```
try:
...
except:
...
else:
    with zipfile.ZipFile(io.BytesIO(zip_response.content)) as z:
        for zfile in z.filelist:
            # Remove the subfolder name from the filepath
            zfile.filename = os.path.basename(zfile.filename)
            z.extract(member=zfile, path=path)
```

icepyx/core/granules.py
```diff
@@ -165,9 +180,16 @@ def get_avail(self, CMRparams, reqparams):
...
-            if len(results["feed"]["entry"]) == 0:
-                # Out of results, so break out of loop
-                break
+            try:
+                if len(results["feed"]["entry"]) == 0:
+                    # Out of results, so break out of loop
+                    break
+            except KeyError:
+                if "errors" in results.keys():
+                    raise ValueError(
```
bwallin, line 189: Not sure if API errors fit the usual interpretation of `ValueError`. Generally I think it is good to have more specific exceptions like "CMRQueryError" or something like that so that people using the library can handle those if they want to (maybe they are doing many searches for granules and sometimes it fails and they want to continue or retry later etc...), but that could be done later.  Maybe `RuntimeError` instead, for now?

icepyx/core/granules.py
```diff
@@ -165,9 +180,16 @@ def get_avail(self, CMRparams, reqparams):
 
             # print(results)
 
-            if len(results["feed"]["entry"]) == 0:
-                # Out of results, so break out of loop
-                break
+            try:
+                if len(results["feed"]["entry"]) == 0:
```
bwallin, line 184: FWIW I think Pep 8 prefers: `if not results["feed"]["entry"]:`

bwallin (approved):
LGTM, just some nit-picks

icepyx/core/granules.py
JessicaS11: Thanks for pointing that out!

icepyx/core/granules.py
JessicaS11: That's a great point. I'm relatively new to raising errors, so haven't gotten beyond using built in ones yet. I've changed it to a `RuntimeError` for now, but I may see if I have a few minutes before the week is out to create a "CMRQueryError", as you suggest.

icepyx/core/granules.py
JessicaS11: Change made - thanks for the suggestion!

icepyx/core/granules.py
JessicaS11: @wallinb I added a QueryError class and tried to implement a custom error. I don't have any code to force a CMR error though - do you have an example I can use to test it?

wallinb: @JessicaS11 In general, I think sending bad input triggers an error response like this (400 status and 'errors' key in content) e.g.:
[https://cmr.earthdata.nasa.gov/search/granules.json?version=003&temporal=badinput&short_name=ATL08](https://cmr.earthdata.nasa.gov/search/granules.json?version=003&temporal=badinput&short_name=ATL08)

For writing tests, this or other error states should be mocked, e.g. can use a convenient library like [responses](https://github.com/getsentry/responses)

Now that I am looking closer at it, how about catching the error earlier with with `response.raise_for_status()`? Just adding that after the `requests.get` would result in an `HTTPError` exception thrown if there is a bad response status. If you want to do more, like raise a custom exception that prints the error message(s) CMR returns, then can do something like this:

```
try:
    response.raise_for_status()
except HTTPError as e:
    if 'errors' in response.content:  # If CMR returns a bad status with extra information, display that
        raise QueryError(response.content) from e
    else:  # If no 'errors' key, just reraise original exception
        raise
```

icepyx/core/granules.py
```diff
@@ -178,18 +177,18 @@ def get_avail(self, CMRparams, reqparams):
...
-                if "errors" in results.keys():
-                    raise QueryError.NsidcQueryError(
-                        str(results["errors"])
-                    )
+                response.raise_for_status()
+            except requests.HTTPError as e:
+                if b'errors' in response.content:  # If CMR returns a bad status with extra information, display that
+                    raise icepyx.core.exceptions.NsidcQueryError(str(results["errors"])) # exception chaining will display original exception too 
+                else:  # If no 'errors' key, just reraise original exception
+                    raise
```
wallinb, line 187: If you move this bit before `results = ...` instead (and use `response.content` instead of `str(results["errors"])`), I think it is a bit safer/simpler. For example, with some errors like a gateway timeout or server problem, the response content may not be json parse-able, but the status code would still indicate an error. As it is, in that case, the user would get a `JSONParseError` instead of an `HTTPError`.

I hope I am not being too much of a pain...you will probably never request a review from me again :grimacing: 

:)

icepyx/core/granules.py
JessicaS11: I used the responses library, as you suggested, to write a test, but I have no idea if it's doing what I think it is (and it seems more like it's testing the CMR API response than it would be testing the actual lines of code in `granules.py`... or am I missing something? I've read about using mocks for testing, but have never actually written one (until now?).

The primary purpose of this PR is to address the fact that CMR would return an error but the user would never see that particular error, which was making it difficult to diagnose issues. With your sample code, I was able to implement something that seems to be doing what we want. Feel free to make any adjustment/improvements!

icepyx/core/granules.py
JessicaS11: Haha - never fear, you're not off the hook yet. I'm using on the other side of PR reviews, so it's nice to have some feedback and I'm learning a bunch!

What you're saying makes sense. I'll move it.

icepyx/tests/test_granules.py
```diff
@@ -592,14 +594,40 @@ def test_no_granules_in_search_results():
...
+    response = requests.get(badreq)
+
+    assert response.json() == {
+        "errors": "temporal start datetime is invalid: [badinput] is not a valid datetime."
+    }
+    assert responses.calls[0].request.url == badreq
+    assert (
+        responses.calls[0].response.text
+        == '{"errors": "temporal start datetime is invalid: [badinput] is not a valid datetime."}'
+    )
```
wallinb, line 633: You have the right idea, but just need to change the assertions to test the icepyx code.

These assertions are demonstrating what the `responses` library does, which is to make `requests.get` react however you specify just for this test, rather than actually making the request on the web. The url in `responses.add`, e.g. `badreq` is what `responses` is looking for everytime `requests.get` is called, and if it matches it will return the response you configured.

What we want is if you call `Granules.get_avail` in this test, it will use the 'mocked' request. To do that, the url here (`badreq`) needs to match what would be called by `Granules.get_avail` for the mock to recognize and return the bad response. In this case it could probably just be a catchall regular expression like `re.compile(re.escape('http://cmr.earthdata.nasa.gov/search/granules.json') + r'.*')` so that any cmr granule search request in duration of this test returns the error response you specified. Does that make sense? E.g.:
```
responses.add(
    responses.GET,
    re.compile(re.escape('http://cmr.earthdata.nasa.gov/search/granules.json') + r'.*'),
    status=400,
    json={
        "errors": "temporal start datetime is invalid: [badinput] is not a valid datetime."
    },
)
```

Note the `status=400` so the mocked response has that status code. After that, if setup a Granules object and call get_avail(), it should trigger the error and you can, for example, `assert_raises(NsidcQueryError)` to see that your code does the thing you expect.

JessicaS11: I've gotten like 95% of the way there, but I still can't get the test to pass... there's clearly some string matching formatting something I'm missing, but the failed assertion shows nothing obvious to my eye. Any idea?

bwallin: Ah, in `pytest.raises` the `match` argument is interpreted as a regex and as such the square brackets have a special meaning.  To "escape" (signal that they should be treated as literal square brackets) you can do:
```with pytest.raises(AssertionError, match=re.escape(ermsg)):```

wallinb: Ah, in `pytest.raises` the `match` argument is interpreted as a regex and as such the square brackets have a special meaning.  To "escape" (signal that they should be treated as literal square brackets) you can do:
```with pytest.raises(AssertionError, match=re.escape(ermsg)):```

JessicaS11: Thank you - worked perfectly! Nic and I played a similar game with Window's paths in ICESat-2 data yesterday.

JessicaS11:
@wallinb I think this is finally ready to go. Thanks for all your help improving it!

wallinb (approved):
Ignore my twin brother over there...

PR #198, opened 2021-04-19
Added Conda and pip badges 

kelseybisson:
We added conda and pip badges to the home documentation page and readme.


rtilling (approved):
All looks good to me

Issue #185, opened 2021-03-12
add conda info for v0.3.4 release


JessicaS11:
We just officially made icepyx available via conda (see #170 and #179) and are planning to release v0.3.4 very soon to push through some kinks we had to work out after 0.3.3. Though there will be other things added in v0.3.4, I'd like to make sure that we get two conda-related additions made:

- [x] add some [conda badges](https://anaconda.org/conda-forge/icepyx/badges) to our home GitHub page
- [x] add some conda badges to our home docs page
- [x] update the install instructions to at least mention that we're available via conda. More install instructions are needed but can come later
- [x] update the readme to note the option for a conda install

If anyone is able/willing to take on any of these tasks in the next few days, feel free to check or review a PR linked to this issue, otherwise go ahead and submit one (and link this issue).

weiji14:
Ping @surfix who might be interested in helping out :wink:

Surfix:
Yes. Thanks @weiji14 . I am interested. I am just about to start exploring the use of IceSat-2 data for my PhD project and I will be glad to contribute to the package.

kelseybisson:
We added PyPi and Conda badges to the Github home page and are going to update the home docs page next Monday as a subgroup (lead2phyto).

weiji14:
FYI, `icepyx` v0.3.4 is now published on PyPI at https://pypi.org/project/icepyx/0.3.4/, and at conda at https://anaconda.org/conda-forge/icepyx (after merging https://github.com/conda-forge/icepyx-feedstock/pull/1)!

wenjieji86:
Hello everyone, I just started to use Anaconda (thanks @JessicaS11 for the recommendation) for my icesat-python projects. So I installed icepyx to my icesat environment which I believe is the 0.3.4 version. However, I got an error message when trying to import the package, anyone had the same issue and how to resolve it? Thanks!

![Capture](https://user-images.githubusercontent.com/41796259/115267245-32d2b200-a0f6-11eb-8562-9fbe93900621.PNG)
![Capture2](https://user-images.githubusercontent.com/41796259/115267255-349c7580-a0f6-11eb-94f3-1597d5bbcb50.PNG)

JessicaS11:
@wenjieji86 Thanks for sharing your issue and the solution you found!

[...]

PR #199, opened 2021-04-20
Update install.rst

wenjieji86:
Added a solution to the potential icepyx importing error, likely caused by the version of the rtree package.

### DIFF
--- a/doc/source/getting_started/install.rst
+++ b/doc/source/getting_started/install.rst
@@ -37,6 +37,11 @@ To upgrade an installed version of icepyx to the latest stable release, do::
 
     conda update icepyx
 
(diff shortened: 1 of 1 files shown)


JessicaS11:
Thanks for this contribution, @wenjieji86, and for your patience while I got to reviewing it (it's been a busy few weeks with all the webinars!).

As noted in the [linked geopandas issue](https://github.com/geopandas/geopandas/issues/1812#issuecomment-824721987):
> I will close this now with this conclusion and we can reopen if it doesn't help. In any case, this is not an issue in geopandas code but with the installation of `rtree` package.

Similarly, this is not an `icepyx` issue, per se, so rather than add the solution to our documentation directly, I would like to explore alternative ways to share the solution to this installation issue. A [Discourse](https://discourse.pangeo.io/c/science/icesat-2/16) post is one option that comes to mind, and I know it was previously mentioned in [#185](https://github.com/icesat2py/icepyx/issues/185#issuecomment-822671395).

PR #200, opened 2021-04-20
adding environmental variable option for Earthdata credentials

tsutterley:
following some thoughts from a discourse conversions with @kelseybisson , added an option to use a temporary environmental variable `EARTHDATA_PASSWORD` for credentials.  

1. feat: adding environmental variable option for Earthdata credentials
2. test: add test for Earthdata credentials
3. docs: adding readme for Earthdata credentials to help get people started

The **test will currently fail travis** as I would need some environmental variables added.
If you feel like this is a good feature @JessicaS11 , then these variables would need to be added (I think travis allows encrypted secrets similar to github actions :shrug: )
- `EARTHDATA_USERNAME` 
- `EARTHDATA_PASSWORD`
- `EARTHDATA_EMAIL`


kelseybisson:
@tsutterley this would be great!!

doc/source/getting_started/earthdata.rst
```diff
@@ -0,0 +1,40 @@
+=====
```
JessicaS11, line 1: I'm excited by the addition of this documentation to help users get set up and oriented and using their Earthdata login within icepyx (I just realized the other day that all of our password storage options aren't well documented, so I wish I'd been able to tackle this PR sooner). I'm wondering if we should give it a broader name (e.g. data access) with multiple sections: one focused on Earthdata login options (which could (should?) also be added to the API docstrings); other data access examples (though we've also got some links to these types of things on the in-need-of-an-update [ICESat-2 resources](https://icepyx.readthedocs.io/en/latest/community/resources.html#other-ways-to-access-icesat-2-data); and common points of confusion specific to ICESat-2 data access (CMR metadata search versus subsetting returns --> and the NSIDC programmatic access guide has a great visual to help explain this).

icepyx/core/Earthdata.py
```diff
@@ -32,7 +33,7 @@ def __init__(
         uid,
         email,
         capability_url,
-        pswd=None,
+        pswd=os.environ.get('EARTHDATA_PASSWORD'),
```
JessicaS11, line 36: wouldn't this mandate that the user set up `EARTHDADA_PASSWORD` and take away the option to log in interactively? Perhaps we could add another `try` option as with the .netrc login capability?

icepyx/tests/test_Earthdata.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
```
JessicaS11, line 1: Not a "review" comment, really, but I would be eager to spend a few moments learning from you about setting up these tests and fixtures. They're things I've only read about but haven't managed to try using yet.

.travis.yml
```diff
@@ -28,7 +28,7 @@ stages:
 jobs:
   include:
     - stage: basic tests
-      script: pytest icepyx/ --ignore icepyx/tests/behind_NSIDC_API_login.py
+      script: pytest icepyx/ --username=$EARTHDATA_USERNAME --password=$EARTHDATA_PASSWORD --email=$EARTHDATA_EMAIL --ignore icepyx/tests/behind_NSIDC_API_login.py
```
JessicaS11, line 31: This strikes me as another option for providing login info for integrated testing, similar to the one used below in `- stage: behind Earthdata` (where NSIDC_LOGIN is saved within Travis, which currently runs our pytest suite). This first suite of tests explicitly avoids any functions that are behind the Earthdata login so that they're tested less frequently (and thus ping NSIDC less frequently). But I also see the need to test all of our methods of logging in. I'll need to revisit our existing setup to refresh my memory on the details of how it's set up.

icepyx/core/Earthdata.py
tsutterley: Great question!  As written, the `get` function returns `None` if the environmental variable does not exist.  So it is effectively the same as the prior version.

icepyx/tests/test_Earthdata.py
tsutterley: sounds good :+1:

doc/source/getting_started/earthdata.rst
tsutterley: these are great points and great ideas.  Happy to help get these docs going :+1:

.travis.yml
tsutterley: good point.  Could easily not have these checks to ping NSIDC.  The included login test is pretty slim to really just check credentials.  Happy to help think this out on the next team call.

icepyx/core/Earthdata.py
JessicaS11: Awesome - thanks for that explanation!

.travis.yml
JessicaS11: Are you available for the next team call (24 May at 8 UTC)? Can we plan to spend some time on testing per the comments in this PR?

tsutterley: I'll be there :+1: 
Sounds good.  Anytime.

kelseybisson: great - I will be there too!

tsutterley:
added the mock Earthdata class @wallinb @JessicaS11

icepyx/tests/test_Earthdata.py
```diff
@@ -0,0 +1,123 @@
...
+u"""
```
weiji14, line 2: Strings are unicode by default in Python3.

```suggestion
"""
```

doc/source/getting_started/earthdata.rst
```diff
@@ -0,0 +1,40 @@
...
+    .. code-block:: bash
+
+        echo "machine urs.earthdata.nasa.gov login <uid> password <password>" >> ~/.netrc
+        chmod 0600 ~/.netrc
+
+   * Temporarily with an environmental variable before running ``icepyx``
+
+    .. code-block:: bash
+
+        export EARTHDATA_PASSWORD=<password>
```
weiji14, line 33: I think this is how people would set an environment variable on Windows, though haven't tested it properly.

```suggestion
        export EARTHDATA_PASSWORD=<password>  # on macOS/Linux
        set "EARTHDATA_PASSWORD=<password>"  # on Windows
```

icepyx/tests/test_Earthdata.py
```diff
@@ -0,0 +1,123 @@
...
+import shutil
+import getpass
+import warnings
+import posixpath
+# from icepyx.core.Earthdata import Earthdata
+
+# PURPOSE: test different authentication methods
+@pytest.fixture(scope="module", autouse=True)
+def setup_earthdata():
+    netrc_file = os.path.join(os.path.expanduser('~'),'.netrc')
```
weiji14, line 18:
```suggestion
    # Before test - move .netrc file to a temporary place
    netrc_file = os.path.join(os.path.expanduser('~'),'.netrc')
```

icepyx/tests/test_Earthdata.py
```diff
@@ -0,0 +1,123 @@
...
+    temprc_file = f"{netrc_file}.temp"
+    if os.access(netrc_file,os.F_OK):
+        shutil.move(netrc_file,temprc_file)
+    yield
+    # clean up
```
weiji14, line 23:
```suggestion
    # After test - move .netrc file back to original path
```

icepyx/tests/test_Earthdata.py
```diff
@@ -0,0 +1,123 @@
...
+    email : string
+        Complete email address, provided as a string.
+    password : string (encrypted)
+        Password for Earthdata registration associated with the uid.
+    capability_url : string
+        URL required to access Earthdata
+
+    Returns
+    -------
+    True if completed a successful "login"
```
weiji14, line 66:
```suggestion
    success : bool
        True if completed a successful "login", else False.
```

icepyx/core/Earthdata.py
```diff
@@ -110,6 +111,10 @@ def login(self):
         The function checks for this file to retrieve credentials, prior to
         prompting for manual input.
 
+        Parameters
+        ----------
+        attempts : number of retry attempts
```
weiji14, line 116:
```suggestion
        attempts : int
            Number of allowed retry attempts.
```

icepyx/tests/test_Earthdata.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+import os
+import re
+import netrc
+import pytest
+import shutil
+import getpass
+import warnings
+import posixpath
+# from icepyx.core.Earthdata import Earthdata
```
weiji14, line 13: So the tests in test_Earthdata.py are now all testing the mock Earthdata class declared at L49. Is the intention to only test things using the mock on Travis CI, and that we need to uncomment this line manually if we want to test the actual CMR API?

icepyx/tests/test_Earthdata.py
```diff
@@ -0,0 +1,123 @@
...
+    Parameters
+    ----------
+    uid : string
+        Earthdata Login user name (user ID).
+    email : string
+        Complete email address, provided as a string.
+    password : string (encrypted)
+        Password for Earthdata registration associated with the uid.
+    capability_url : string
+        URL required to access Earthdata
```
weiji14, line 62:
```suggestion
    uid : str
        Earthdata Login user name (user ID).
    email : str
        Complete email address, provided as a string.
    password : str
        Password for Earthdata registration associated with the uid.
    capability_url : str
        URL required to access Earthdata
```

icepyx/tests/test_Earthdata.py
tsutterley: that is how it is currently written.  I thought about making it a command line option (like `--use-api`, `--full-integration` or something) and changing the mock definition to `_Earthdata` or something similar.

weiji14: > that is how it is currently written. I thought about making it a command line option (like `--use-api`, `--full-integration` or something) and changing the mock definition to `_Earthdata` or something similar.

This seems rather complicated (and possibly hard to maintain). I don't have a good solution either, but wondering if there's a built-in pytest plugin that can be toggled on/off to do this sort of thing. Or what about scheduling the test_Earthdata.py tests to run only once a day or once per week if pinging the NSIDC servers constantly is an issue?

Side note: Using 'mocks' often seems to be a last resort (in my limited experience of using them in unit tests), so I just want to make they're used here because other options have been exhausted already.

tsutterley: That would be in travis to only run the full integration tests occasionally.  I think that is what @JessicaS11 is doing already for some of the testing suite.  Github actions has cron scheduling as well.

Adding the mock was suggested just to keep the testing suite isolated.  I doubt pinging the CMR system for a token on PRs would be a problem (except for cases when the system is down for scheduled maintenance).  Would come up with a workaround in those cases. 🤷 

Might be worth some further discussion.

JessicaS11: @tsutterley captured my thoughts correctly. I think we should keep `test_Earthdata.py` on a mock to keep the testing isolated. Then we can tear into `behind_NSIDC_API_login.py` (including a rename that includes "test" - there was a major code refactor early on and not adding test to the filename is my cue that these tests/files weren't updated to the new code structure) to do the actual pinging of NSIDC (which is already set up to run as a cron job via Travis, but was commented out to keep it from running until the tests were updated).

I will start into that updating, since I think that's one of the other pieces keeping this PR from being merged.

icepyx/tests/test_Earthdata.py
```diff
@@ -0,0 +1,125 @@
...
+
+def test_netrc(username,password,email):
+    # append to netrc file and set permissions level
+    args = (username,'urs.earthdata.nasa.gov',password)
+    netrc_file = os.path.join(os.path.expanduser('~'),'.netrc')
+    with open(netrc_file,'a+') as f:
+        f.write('machine {1} login {0} password {2}\n'.format(*args))
+        os.chmod(netrc_file, 0o600)
+    url = capability_url("ATL06","003")
+    assert Earthdata(username,email,url).login(attempts=0)
```
JessicaS11, line 48: This is all making my head spin a little, so please bear with me.

My understanding is that we'd want to test:
1. if the password is set as an environment variable, it's used to login
2. if the password is not set as an environment variable, is it within a netrc (which also must contain the user ID, which cannot be entered as an environment variable)
3. otherwise, do we prompt the user for the password

Currently, `test_environment` looks for a netrc file (that doesn't exist) and assuming there isn't one can still login because the environment variable was set. This environment variable is provided by Travis during the test configuration in `conftest.py`.

Then, `test_netrc` creates a netrc file with the requisite info and uses that to log in. However, since the environment variable is already declared for the entire module, wouldn't we just be effectively testing that login again because it tries that first without ever getting to the netrc file? I'd guess this would be fairly easy to change by explicitly setting `password=None` within this test, but I still don't understand all of this well enough to trust that instinct. Or am I just reading this test wrong?

icepyx/tests/test_Earthdata.py
```diff
@@ -0,0 +1,125 @@
...
+        self.capability_url = capability_url
+        self.pswd = pswd
+
+    def _start_session(self):
+        """
+        Checks supplied credentials versus environmental variables
+        """
+        EARTHDATA_USERNAME = os.environ.get('EARTHDATA_USERNAME')
+        EARTHDATA_PASSWORD = os.environ.get('EARTHDATA_PASSWORD')
+        EARTHDATA_EMAIL = os.environ.get('EARTHDATA_EMAIL')
```
JessicaS11, line 95: I'm having trouble wrapping my head around why these need to get set up as environment variables within Travis rather than just being static values we're passing into the tests/mock via `conftest.py`. The only environment value that `Earthdata.py` will attempt to look for is `EARTHDATA_PASSWORD`, so wouldn't it be better to only directly pass username and email as variables into `Earthdata()` like a user would be doing, excepting the case where they're set in the netrc file? Otherwise, I can't see where/how/why we'd expect `EARTHDATA_USERNAME` to be the same as `username` since we could easily change the former on Travis but it's not used to generate `self.uid` during the test mock setup process.

icepyx/tests/test_Earthdata.py
JessicaS11: see #209

doc/source/getting_started/earthdata.rst
JessicaS11: I'm thinking about moving this effort to its own branch - any objections?

tsutterley: Sounds good :+1:

icepyx/core/Earthdata.py
```diff
@@ -122,15 +128,15 @@ def login(self):
             session = self._start_session()
 
         except:
-            self.pswd = getpass.getpass("Earthdata Login password: ")
```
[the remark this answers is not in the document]
tsutterley: You're right this needs to be changed :+1:

doc/source/getting_started/earthdata.rst
JessicaS11: Finally created a new branch for this - login_docs.

icepyx/tests/test_Earthdata.py
JessicaS11: @tsutterley Any updates on this? I've tried a few things but don't recall the intended setup well enough and fear I might be hurting rather than helping...

tsutterley: Hey @JessicaS11 thanks for the reminder.  I think everything in "core" is in working condition (tested it again this morning).  However I'm having a hard time justifying the mock login tests.  They don't really test anything within the Earthdata class since they're isolated.

JessicaS11: Hmmm... Was the idea that we test that the correct login option is being used, which IS a valuable set of tests? Then an actual login test (independent of this PR) that pings NSIDC can verify that `Earthdata._start_session()` is working right.

tsutterley: i think that's right.  i made some edits to conftest to use the same credentials in `test_behind_NSIDC_API_login.py` should we ever decide to go off the mock.  i think it should pass travis now

JessicaS11: Thanks @tsutterley! This looks great (and all the tests pass). Now we can move ahead with finishing up the [login docs](https://github.com/icesat2py/icepyx/tree/login_docs) page.

JessicaS11 (approved):

PR #201, opened 2021-05-04
apply black formatting to code and test files

JessicaS11:
we discovered during #144 that a bunch of the files have fallen behind in their black formatting, so that's applied here.


icetianli (approved):
Looks good to me! @JessicaS11

PR #202, opened 2021-05-12
ATL08 icepyx intro tutorial example

JessicaS11:
I put together a quick intro tutorial for a demo at the ICESat-2 Applied Users Working Group Quarterly call (12 May 2021). I've added it as an example, since I used ATL08 and we often get requests for examples showing non-ice datasets.

I also removed some old dev/draft notebooks that were still hanging around.


JessicaS11:
Closing in favor of the [more recent tutorial](https://nasa-openscapes.github.io/2023-ssc/tutorials/data-access/icepyx.html) documented for adding to the docs in #639

PR #209, opened 2021-06-15
tests that require an active NSIDC Earthdata session

JessicaS11:
As noted in #200, tests behind the NSIDC login were never updated after the code refactor. This PR aims to reactivate those tests.

@wallinb @tsutterley @weiji14 Can you look this over and offer any suggestions (there are a few questions as comments in the code, which in general might be easier read not as a diff from the original file)


weiji14 (approved):
The changes look ok (though I only scanned the git diff). Was there anything in particular you want us to look at closer?

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -2,41 +2,31 @@
...
 @pytest.fixture
-def reg_a(scope="module"):
+def reg(scope="module"):
     return ipx.Query("ATL06", [-55, 68, -48, 71], ["2019-02-22", "2019-02-28"])
 
 
-# @patch('my_module.__get_input', return_value='y')
-
-
+# QUESTION: should we be testing independently to make sure the session starts, or is the fixture sufficient test of that?
```
JessicaS11, line 16: @weiji14

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -2,41 +2,31 @@
...
-# def test_earthdata_session_started(session):
-#     assert isinstance(session, 'requests.sessions.Session')
-
-
+########## is2ref module ##########
 import icepyx.core.is2ref as is2ref
 
-########## _get_custom_options ##########
-# TestQuestion: should this use a mock rather than an actual call to NSIDC?
 # TestQuestion: is there a better way to deal with this than having the super long dictionary copied here?
```
JessicaS11, line 30: @weiji14

JessicaS11:
Thanks, @weiji14! I tagged you for two of the questions I had. I'm still getting the hang of mocks and fixtures for testing, so I was hoping for a higher-level conceptual review similar to what we did during our icepyx meeting for #200 by someone who's got a bit more experience with them (and to make sure I didn't miss any of the key points we're working to address in #200).

icepyx/tests/test_behind_NSIDC_API_login.py
weiji14: I think it's ok to have a fixture that sets up the session, it's similar to the example at https://docs.pytest.org/en/latest/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session where the 'module' scope fixture connects to a server. So no, we don't need an independent test to ensure that the session starts.

icepyx/tests/test_behind_NSIDC_API_login.py
weiji14: If you're referring to the question; "should this use a mock rather than an actual call to NSIDC?", it depends. My preference is to lean towards a simpler solution (don't use a mock) since it's not easy finding people who understand how to set up pytest mocks (maintenance burden). But if the intention (as with https://github.com/icesat2py/icepyx/pull/200#discussion_r642704522) is to **not** ping the CMR system too often, then a mock would be preferable.

If you're referring to the "is there a better way to deal with this than having the super long dictionary copied here?" question. Then I suppose we could store the dictionary in a plain-text JSON or YAML file and load it during the test to compare?

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -0,0 +1,47 @@
...
+    path = "./downloads_subset"
+    reg_a.order_granules(session)
+    reg_a.download_granules(session, path)
+
+
+# def test_download_granules_without_subsetting(reg_a, session):
+#     path = './downloads'
+#     reg_a.order_granules(session, subset=False)
+#     reg_a.download_granules(session, path)
+#     #check that the max extent of the downloaded granules isn't subsetted
```
JessicaS11, line 47: @wallinb Would you recommend that we try to do some sort of testing on our end to make sure we're getting correctly subsetted files, or is it sufficient to test that we're sending proper requests to the subsetter (I think this still needs to be implemented, but that's besides the point) and the accuracy of the input/output of the subsetter is tested on the NSIDC end?

JessicaS11: @wallinb Any thoughts on this? I'd like to get this PR ready for review, so that we can move forward with #200.

JessicaS11: I think this is still a relevant question, but there are some staff changes over at NSIDC that may delay getting an answer to this in the short term, so I'm going to resolve this for now and have added it to my long dev item list.

PR #211, opened 2021-06-21
update variable names for clarity and in line with common usage

### LINKED ISSUE #104: fix duplicate and ambiguous variable names
some attributes (e.g. `source` and `_source`) have different functionalities. These should be made consistent. Relatedly, `files` is an ambiguous variable name.

JessicaS11:
Closes #104, which notes that there are some ambiguous and confusing variable names. For the most part, the user should not be affected by these changes, though a deprecation notice is included for the transition in `Query` object initialization from "dataset" to "product" for users who are using keyword arguments.


icepyx/core/query.py
```diff
@@ -111,22 +111,29 @@ class Query:
...
 
         # warnings.filterwarnings("always")
         # warnings.warn("Please note: as of 2020-05-05, a major reorganization of the core icepyx.query code may result in errors produced by now depricated functions. Please see our documentation pages or example notebooks for updates.")
 
+        if dataset != None:
+            warnings.filterwarnings("always")
+            warnings.warn(
+                "In line with most common usage, the 'dataset' input argument has been replaced by 'product'."
+            )
+
```
icetianli, line 134: Thanks for this PR @JessicaS11, it looks great! Just some small suggestions. Since we have changed all `dataset` into `product`, do we still need to keep `dataset` parameter in `Query` class here? I think we can just update the example notebooks and add a note that we have changed `dataset` into `product`, I reckon most users won't use keyword argument `product = 'ATL03'` to specify the product they want, but would directly pass product name as the first positional argument as the case used in example notebook :`region_a = ipx.Query(short_name, spatial_extent, date_range)`.

icetianli (changes requested):

icepyx/core/query.py
JessicaS11: Thanks for the review, @icetianli, and for catching these problems. I fixed the unintentional non-icepyx library removals of dataset you noted. I suspect you're right that most users aren't using the keyword, so I removed the warning from the `Query.__init__` and added a note, as you suggested. Your comment also made me realize that I didn't update the example notebooks, so I went through and made those changes as well. Since there had been a `dataset` property on the query object that was replaced by a `product` property, I went ahead and made a `dataset` property that has the deprecation warning instead.

icetianli: Thanks for the changes! Regarding the failed pytest on test_visualization.py: `cannot import name 'ABCIndexClass' from 'pandas.core.dtypes.generic'`, I found that it is because the [`holoviews`](https://github.com/holoviz/holoviews/blob/master/holoviews/core/util.py) isn't compatible with the latest `pandas` - version 1.3.0 released four days ago - they have now changed ['ABCIndexClass' into 'ABCIndex' ](https://github.com/pandas-dev/pandas/commit/eb67d87b6b133242e41293054238c17d54a69bd9#diff-d19c748aa6d9ba07f1a0a428729a220e227f8616cddce4a5e202ad793a04d66b) so holoviews cannot import them anymore. To solve this issue, could we specify the default pandas version to be 1.2.5?

JessicaS11:
I know that the Travis tests aren't passing - if you look at the build (click details, then the link on the words "the build"), a bunch of tests are failing related to granule searches and visualization (so not related to this PR). I've reached out to NSIDC (@asteiker @wallinb) about recent CMR updates that could indicate we need to update our tests.

icetianli:
> I know that the Travis tests aren't passing - if you look at the build (click details, then the link on the words "the build"), a bunch of tests are failing related to granule searches and visualization (so not related to this PR). I've reached out to NSIDC (@asteiker @wallinb) about recent CMR updates that could indicate we need to update our tests.

OK sounds good!

JessicaS11:
@icetianli The tests all pass! From @wallinb: there was an issue earlier today with one of the backend components to CMR, which seems the most likely explanation.

icepyx/core/is2ref.py
```diff
@@ -149,14 +149,14 @@ def get_varlist(elem):
```
icetianli, line 149: Under `_get_custom_options` function, need to update the `dataset` in `capability_url =(f"https://n5eil02u.ecs.nsidc.org/egi/capabilities/{dataset}.{version}.xml")` into `product` as well, otherwise `region_a.show_custom_options(dictview=True)` will throw a `NameError: name 'dataset' is not defined`.

icetianli (changes requested):

icetianli:
I notice that the ICESat-2_Data_Visualization_Example notebook still uses the outdated variable name `dataset`: `print(region.dataset)` & `print(region.dataset_version)`, should we also update them accordingly...? I am happy to help with this!

JessicaS11:
> I notice that the ICESat-2_Data_Visualization_Example notebook still uses the outdated variable name `dataset`: `print(region.dataset)` & `print(region.dataset_version)`, should we also update them accordingly...? I am happy to help with this!

Good catches - I updated the visualization example. I'm not going to bother with the DEM comparison example for now since it's fairly unmaintained anyway at this point.

icetianli (approved):
Great - they all look good to me now! @JessicaS11

Issue #207, opened 2021-06-10
Querying granules by names only?


whyjz:
Hi Icepyx devs! Does `icepyx.Query` work if only the granule name is provided? I tried this

```
filename = 'ATL03_20190805232841_05940403_002_01.h5'
query = icepyx.Query(dataset='ATL03', files=filename)
```

and it complained that I didn't enter a valid spatial extent. But I didn't know the spatial extent (unless I check it up on Earthdata Search) because I was following somebody else's work with the only file name specified.

Thanks in advance!

JessicaS11:
Hello @whyjz! If you take a look at the code, the `files` parameter in the initialization of the `query` object is actually a legacy placeholder in work towards reading in local files that's not implemented (hence why it's not included in the docstring or examples) that should probably be removed (would you like to submit a PR to do this?). Since filenames are easy to change, it's not a reliable way to search for data from NSIDC anyway. Since you seem to have an unaltered filename in this case, I'd recommend using it to get the track and cycle and input those parameters along with a whole world spatial extent. We did not remove the spatial extent requirement so that users searching for a larger number of tracks/cycles (versus a specific one) did not inadvertently end up with global data if they were only interested in a small region. I haven't tested this to check that you get exactly this granule, but I imagine your code would look something like:
```
query = ipx.Query(dataset="ATL03", 
                  spatial_extent=[-180, -90, 180, 90], 
                  date_range=['2019-08-05', '2019-08-05'], 
                  cycles=['04'], 
                  tracks=['0594']
                  )
```

If this still returns more granules than you expect, you could further narrow your search if you know what region of the world you're in (so could reduce the spatial extent) or provide start and end times (which can also be extracted from the filename in this case). Please let us know if this works for you!

JessicaS11:
@whyjz Can you confirm that you were able to find specific files as suggested?

whyjz:
Hi @JessicaS11, sorry about the delay from my side (was in quarantine in Taiwan during the past 2 weeks). I actually ended up querying the desired file using a similar way that you posted (see #27 and [here](https://workflow.jupytearth.org/icesat2-crevasses/getdata.html)). And it is understandable why querying using names only is not recommended. Yes, I am happy to do a quick PR to remove the old placeholder on the doc!

[...]

PR #212, opened 2021-06-25
Query docstring

whyjz:
Hi - this is for the discussion in #207. I added a few lines to the Query docstring so users can get a bit more information about the arguments. I do not plan to remove the `files` argument because maybe we will need it someday in the future, as @JessicaS11  mentioned from that issue.

P.S. There are 4 commits in total, but the first 3 of them are actually to revert my previous commit (see #171) that is not going to be merged.


JessicaS11 (approved):

JessicaS11:
Thanks for these updates, @whyjz!

PR #221, opened 2021-09-09
traffic/downloads statistics updates and limit traffic action to parent repo

JessicaS11:
This PR updates the icepyx tracking/usage statistics and modifies the traffic action to only run on the icesat2py organization (so it won't run on forks of the repo).


kelseybisson:
This looks good to me, and I'm guessing it will also prevent the weekly emails to forked accts? My latest says 'The ".github/workflows/traffic_action.yml" workflow in kelseybisson/icepyx will be disabled soon' and previously I received '.github/workflows/traffic_action.yml: All jobs have failed'

@RomiP what do you think?

JessicaS11:
@kelseybisson Thanks for looking this over! When I was looking up how to get the action to not run on forks, I found a few examples of people who'd forked an active library with lots of automated actions and then run out of their personal "server time" (or however it's counted) for running actions, so I think GitHub is working on making it so that cron actions don't run on forks (but I figured it still doesn't hurt to be explicit).

kelseybisson (approved):

PR #244, opened 2021-11-30
update add and commit GitHub Action version

JessicaS11:
Looks like the add-and-commit GitHub Action is out of date, so I'm starting there.


JessicaS11:
@weiji14 We need an approving review to trigger the action. Obviously I won't merge if it doesn't actually successfully update the umls.

weiji14 (approved):
Ok, see if this works.

PR #249, opened 2021-12-06
Super query

### LINKED ISSUE #205: Superclass/subclass type for query object
We’re looking for feedback before making a change to the underlying icepyx architecture.

Following discussions from the Lead2Phytoplankton group, we are working to incorporate Argo and other observations access into icepyx. The big idea is to have a general structure so that new datasets can be implemented relatively simply, with perhaps different spatiotemporal windows of interest, and altogether visualized within a colocated_data object. This sets the framework to allow users to define different spatial/temporal constraints on their queries for different datasets in the future. [Here's an example of how we've structured a superclass during our hackweek project.](https://github.com/ICESAT-2HackWeek/leading_to_phytoplankton/blob/master/scripts/dataset_scripts/dataset.py)

Every dataset that we will query will have something in common (time range, spatial bounding box). Because these attributes are common to all datasets, going forward it could make sense to include them within a superclass, which would include these shared parameters. If we can inherit the time/space functionality from a superclass it would enable easier implementation of other datasets in the future, and reduce maintenance burden by eliminating duplicated code. 

Thanks for any ideas/concerns!

@RomiP @rtilling @alexdibella @JessicaS11

RomiP:
Changed Query framework such that parameters that are common to all datasets (other than IceSat2 e.g. Argo, MODIS, etc.) exist in a superclass structure. This should NOT affect existing Icepyx functionality.
This is to facilitate adding Icepyx functionality to other datasets in the future.

Docstring testing in Query Class has also been improved.

Closes #205


JessicaS11:
Did some work on this branch (including a rebase onto the latest development branch):
@RomiP I took the docstring testing setup out and put it into #256. We'll need to review + merge that PR before this one, as this branch is now on top of it.
@kelseybisson I'm turning things over to you now to remove the sandbox files from this branch so we can review and merge.

JessicaS11 (approved):

PR #259, opened 2021-12-23
Update or skip all existing docstring tests

JessicaS11:
In #252 and #249 we did a bunch of work to improve/use the docstring examples as function tests for the `Query` module. This PR reviews all existing docstring tests and either fixes them to pass or skips them (skips are used in cases where there would be a very long - e.g. 200+ line long output or where the test requires data received from a request that requires an Earthdata login).


JessicaS11:
@RomiP For some reason the polygon files tests in Query and SuperQuery won't pass for me. It's saying it can't find the file, even though the file is there, the path is right, and it works fine if I call it from within the example notebook. Can you try running those docstring tests on your setup so we can see if it's my environment or not (I also can't autocomplete the gpkg filename in terminal, which is why I suspect my setup).

PR #261, opened 2022-01-12
Update readthedocs build dependencies

weiji14:
Fixes the missing bullet points in the Sphinx documentation build.

**Preview** at https://icepyx--261.org.readthedocs.build/en/261/community/resources.html, https://icepyx--261.org.readthedocs.build/en/261/_icepyx/icepyx.Query.html

| Before | After |
|--|--|
| image | image |
| image | image |

References:
- https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python
- https://docs.readthedocs.io/en/stable/builds.html#default-versions-of-dependencies
- https://github.com/readthedocs/sphinx_rtd_theme/issues/752


weiji14:
@JessicaS11, I've managed to track down the problem with rendering lists (i.e. bullet points). Turns out that readthedocs was pinning some Sphinx dependencies to old versions (since icepyx is a project that started before Oct 2020), see https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx

image

So the solution was to explicitly set sphinx and sphinx_rtd_theme to be higher versions. I actually found it strange while my local documentation builds were rendering ok, and it turns out the problem was just on the readthedocs server side :slightly_smiling_face:

JessicaS11:
> I've managed to track down the problem with rendering lists (i.e. bullet points). Turns out that readthedocs was pinning some Sphinx dependencies to old versions (since icepyx is a project that started before Oct 2020), see https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx

Yay! Thank you for figuring this out.

JessicaS11 (approved):

PR #273, opened 2022-02-07
QUEST - Colocated data framework

RomiP:
Introduce framework to implement Icepyx functionality (query & download) to other data  sets 
QUEST
Query_Unify_Explore_SpatioTemporal


JessicaS11:
By way of updating/summarizing my boatload of test commits:
- for right now, we're stuck with __init__ being listed as a method for the class docs. Private functions shouldn't show up by default, but I can't find any `.. autosummary:` docs that talk about it, only ones for `.. automodule:`
- Combining numpy docstrings (which is the style we use) from inherited/super classes is nontrivial. It's easy to just concat/append the two docstrings (sub+superclass), but then sphinx complains there are two of every heading and can't parse the docstring. There are ways around it ([e.g.](https://stackoverflow.com/questions/60000179/sphinx-insert-argument-documentation-from-parent-method/60012943#60012943)), but I didn't want to add a new (otherwise unrelated) dependency and using the pandas `@doc` or local decorator was more involved than a "simple fix" and far beyond the scope of this PR.

zachghiaccio (approved):
This looks good to me.

PR #276, opened 2022-02-11
minor tweaks for gridded product access

JessicaS11:
Gridded ICESat-2 data products are now available from NSIDC. They are available via icepyx, but do not require the same inputs as non-gridded products (e.g. temporal data) and have different granule filename formats. Minor changes were made to ease querying, ordering, and download of gridded products.


doc/source/example_notebooks/IS2_data_access.ipynb
```diff
@@ -83,7 +83,7 @@
     "*NOTE: The input keyword for `short_name` was updated in the code from `dataset` to `product` to match common usage.\n",
     "This should not affect users providing positional inputs as demonstrated in this tutorial.*\n",
     "\n",
-    "Then, you must include AT LEAST one of the following inputs (temporal or orbital constraints):\n",
+    "Then, for all non-gridded products (ATL<=12), you must include AT LEAST one of the following inputs (temporal or orbital constraints):\n",
```
weiji14, line 86:
```suggestion
    "Then, for all non-gridded products (ATL<=13), you must include AT LEAST one of the following inputs (temporal or orbital constraints):\n",
```

doc/source/example_notebooks/IS2_data_read-in.ipynb
```diff
@@ -321,13 +331,15 @@
...
     "icepyx will determine what variables are available for you to read in by creating a list from one of your source files.\n",
     "If you have multiple files that you're reading in, icepyx will automatically generate a list of filenames and take the first one to get the list of available variables.\n",
     "\n",
     "Thus, if you have different variables available across files (even from the same data product), you may run into issues and need to come up with a workaround (we can help you do so!).\n",
     "We anticipate most users will have the minimum set of variables they are seeking to load available across all data files, so we're not currently developing this feature.\n",
-    "Please get in touch if it would be a helpful feature for you or if you encounter this problem!"
+    "Please get in touch if it would be a helpful feature for you or if you encounter this problem!\n",
+    "\n",
+    "You may create a variable list for pre-gridded products. However, all variables in the file will be added to your DataSet. (This is an area we're currently exploring expanding - please let us know if you're working on this and would like to contribute!)"
```
weiji14, line 342: Pre-gridded sounds funny, maybe just say gridded.

```suggestion
    "You may create a variable list for gridded ICESat-2 products. However, all variables in the file will still be added to your DataSet. (This is an area we're currently exploring expanding - please let us know if you're working on this and would like to contribute!)"
```

weiji14 (approved):
I always wondered how icepyx would eventually handle the gridded ICESat-2 products, it just never occurred to me that some if-then statements was all that was needed!

Just some minor suggestions on the jupyter notebook docs. Ideally there would be some proper unit tests for ATL14 and above, but I suppose there isn't much time for that since the hackweek is getting close.

doc/source/example_notebooks/IS2_data_access.ipynb
JessicaS11: I have this fix committed locally and forgot to push! Thanks for catching it again!

JessicaS11:
> I always wondered how icepyx would eventually handle the gridded ICESat-2 products, it just never occurred to me that some if-then statements was all that was needed!
Me too - so I was also pleasantly surprised when I tried it and found that I could resolve the obvious issues with some if statements! I think it may be worth revisiting the read-in portion eventually (it would be nice to enable read-in of only desired variables), but as you said, the hackweek is coming up fast and that's a pretty new/limited functionality module still anyway.

> Ideally there would be some proper unit tests for ATL14 and above, but I suppose there isn't much time for that since the hackweek is getting close.
Agreed. Funny how nobody wants to write tests though...

Issue #272, opened 2022-02-03
PRs from forks fail Travis CI with `Earthdata authentication failed`


JessicaS11:
In #271 we ran into the issue of all Travis-CI tests erroring after the setup in `conftest.py` failed because of secret environment variables. Travis-CI does not allow use of encrypted env variables for external (fork-based) PRs because it's a security risk. We need to modify our CI setup ([e.g.](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions)) so that tests that don't require the env variable can still run for forked PRs.

Example Travis CI failure from a fork PR:

```
FAILED icepyx/tests/test_auth.py::test_get_session - icepyx.core.auth.AuthenticationError: Earthdata authentication failed. Check output for error message

FAILED icepyx/tests/test_auth.py::test_get_s3login_credentials - icepyx.core.auth.AuthenticationError: Earthdata authentication failed. Check output for error message

FAILED icepyx/tests/test_auth.py::test_login_function - icepyx.core.auth.AuthenticationError: Earthdata authentication failed. Check output for error message
```

mfisher87:
I think we can split up our tests into integration tests which require EDL (or other third-party services) and unit tests which don't, and only run unit tests on PRs from forks. Thoughts?

mfisher87:
:scream: I just discovered that branch protections are preventing me from merging PRs from forks (e.g. #557)

Breaking the tests up into two parts would address this, but for now can I be permitted to bypass branch protection?

mfisher87:
I applied a quick fix to split the tests. In the long run, I'd like to organize the tests into two folders (unit, integration), and provide scripts or [tasks](https://learn.scientific-python.org/development/guides/tasks/) to run the tests separately. It will clear which have external dependencies and which don't, but that's not urgent :)

PR #284, opened 2022-03-04
Add __init__.py for quest module

weiji14:
Resolves a ModuleNotFound error:

```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Input In [1], in <cell line: 1>()
----> 1 import icepyx as ipx

File ~/mambaforge/envs/hackweek/lib/python3.9/site-packages/icepyx/__init__.py:3, in <module>
      1 from icepyx.core.query import Query, GenQuery
      2 from icepyx.core.read import Read
----> 3 from icepyx.quest.quest import Quest
      5 from _icepyx_version import version as __version__

ModuleNotFoundError: No module named 'icepyx.quest'
```


weiji14:
Ok hold on, I'm still getting an error, let me double check that this PR works...

JessicaS11:
`pip install -e.` is working fine for me... are you doing a different install?

weiji14:
Nevermind, false alarm. I just needed to restart my jupyter kernel. But all those test failures on Travis CI doesn't look good, do you know what's up with that?

JessicaS11:
> Nevermind, false alarm. I just needed to restart my jupyter kernel. 

I think there's a meme for that. ; )

> But all those test failures on Travis CI doesn't look good, do you know what's up with that? Maybe it's because it's running on a fork rather than the main icesat2py/icepyx repo?

I had the same thought - thanks for pointing out the fork bit. The tests that failed are behind a login, so the encrypted variable can't be accessed when a fork triggers the tests (see also #272).

weiji14:
Err, I still think we need that __init__.py file. Otherwise `import icepyx` won't work.

JessicaS11:
Yeah. My terminal python wasn't pointing where I thought it was, so it was still working for me even after I reinstalled the newer version. I'll try and push this through and re-release later tonight.

JessicaS11 (approved):

Issue #288, opened 2022-03-05
Accessing ATL14 gridded land ice data with icepyx.Query


weiji14:
With #276, there was some logic added that allows icepyx to load ICESat-2 gridded products that have been downloaded already. But in order to **download** these ICESat-2 gridded products like ATL14/ATL15, it looks like more changes are needed. This is the error I'm getting when trying to load an ATL14 sample over Iceland.

```python
import icepyx as ipx

region_iceland = ipx.Query(
    product="ATL14",  # ICESat-2 Gridded Annual Ice Product
    spatial_extent=[-28.0, 62.0, -10.0, 68.0],  # minlon, minlat, maxlon, maxlat
)
region_iceland.earthdata_login(
    uid="penguin123",  # EarthData username
    email="penguin123@southpole.net",  # Email address
)
region_iceland.download_granules(path=".")
```

<details>

```python-traceback
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
File ~/mambaforge/envs/hackweek/lib/python3.9/site-packages/icepyx/core/Earthdata.py:131, in Earthdata.login(self, attempts)
    130 url = "urs.earthdata.nasa.gov"
--> 131 self.uid, _, self.pswd = netrc.netrc(self.netrc).authenticators(url)
    132 session = self._start_session()

File ~/mambaforge/envs/hackweek/lib/python3.9/netrc.py:29, in netrc.__init__(self, file)
     28 self.macros = {}
---> 29 with open(file) as fp:
     30     self._parse(file, fp, default_netrc)

FileNotFoundError: [Errno 2] No such file or directory: '/home/username/.netrc'

During handling of the above exception, another exception occurred:

SSLCertVerificationError                  Traceback (most recent call last)
File ~/mambaforge/envs/hackweek/lib/python3.9/site-packages/urllib3/connectionpool.py:703, in HTTPConnectionPool.urlopen(self, method, url, body, headers, retries, redirect, assert_same_host, timeout, pool_timeout, release_conn, chunked, body_pos, **response_kw)
    702 # Make the request on the httplib connection object.
--> 703 httplib_response = self._make_request(
    704     conn,
    705     method,
    706     url,
    707     timeout=timeout_obj,
    708     body=body,
    709     headers=headers,
    710     chunked=chunked,
    711 )
    713 # If we're going to release the connection in ``finally:``, then
    714 # the response doesn't need to know about the connection. Otherwise
[...]

Issue #286, opened 2022-03-04
fix travis release builds


JessicaS11:
Currently release PRs fail on Travis because it doesn't properly create an Earthdata session (so all the tests behind Earthdata fail).

PR #289, opened 2022-03-06
Testing ATL14 gridded land ice download from earthdata

weiji14:
Adding unit tests for ATL14 data access via Earthdata, using a region over Svalbard (Iceland is smaller, but bounding box overlaps with Greenland which is big).

Towards resolving #288, but also updating tests behind NSIDC login for the ATL06 dataset by updating from v05 to v06.


JessicaS11:
Looks like the issue with adding this test is now #286.

weiji14:
Ok, but I might as well add some unit tests for ATL14 since I've been playing with it recently. Will see what I can do to increase code coverage.

JessicaS11:
> Ok, but I might as well add some unit tests for ATL14 since I've been playing with it recently. Will see what I can do to increase code coverage.

Sounds great (and wasn't suggesting we not add the tests already here).  Just noticed that now that #288 is resolved with updated certificates at NSIDC, the failing Travis errors were "updated" to #286.

JessicaS11:
See #291

icepyx/tests/ATL06v06_options.json
weiji14, line 1: Diff between ATL06v05 and ATL06v06 at https://www.diffchecker.com/uCMMViSG/. Main changes are

```diff
- "fileformats": ["TABULAR_ASCII", "NetCDF4-CF", "Shapefile", "NetCDF-3"]
+ "fileformats": ["TABULAR_ASCII", "NetCDF4-CF", "Shapefile"]
- "formatreproj": ["TABULAR_ASCII", "NetCDF4-CF", "Shapefile", "NetCDF-3"]
+ "formatreproj": ["TABULAR_ASCII", "NetCDF4-CF", "Shapefile"]
...
+ "orbit_info/bounding_polygon_lat1", "orbit_info/bounding_polygon_lon1"
```

weiji14 (commented):
Cool, should be ready for review! Have updated the `test_behind_NSIDC_API_login` to include the new ATL14 (v002) dataset, and also updated the ATL06 schema to v006 for good measure.

weiji14:
Hi @rwegener2, could you give this a review if you have time, since you recently worked on the #435 PR?

rwegener2:
> @rwegener2, could you give this a review if you have time

Absolutely, @weiji14. I've been offline for a day or two while moving apartments, but I'll take a look this afternoon.

rwegener2 (approved):
This looks good to me @weiji14. Nothing really changed with respect to authentication as far as I can tell, was there something specific you wanted a second set of eyes on? The ATL14/06 specific bits I am less familiar with, but from my experience everything looks solid!

icepyx/tests/test_behind_NSIDC_API_login.py
JessicaS11, line 1: @weiji14 If I'm understanding this right, we've added ATL14 as part of the fixture, but then don't actually run any tests on it beyond creating a query object with those parameters (in which case, shouldn't we just create a query object with it elsewhere instead of in `test_behind_NSIDC_API_login.py`). That or the assertions in lines 50-51 will fail because there's no appropriate json file for the ATL14 custom options...

PR #281, opened 2022-02-25
add read-in functionality for deeply nested variables

JessicaS11:
When reading in some deeper nested variables (here `canopy/h_canopy`) in ATL08, xarray does not read in the coordinates for the data which are at the `canopy` level. This caused merge issues when trying to build a DataSet of all wanted variables. This PR adds functionality that handles these more deeply nested variables by assigning them the proper coordinate values. It also fixes a few typos and an index error that was occurring during data read-in.

PR #294, opened 2022-03-16
manually remove 'z' from datetime string

JessicaS11:
Per the segfault issue noted in #281, Xarray/pandas/numpy no longer automatically handle datetimes brought in with a timezone ('Z' at the end of the datetime string). This provides a fix by checking and manually removing the last character of the datetime string if it's a Z.

### DIFF
--- a/icepyx/core/read.py
+++ b/icepyx/core/read.py
@@ -363,13 +363,14 @@ class Read:
                 pass
 
             try:
-                # manually remove 'Z' from datetime to allow conversion to np.datetime64 object (support for timezones is deprecated and causes a seg fault)
-                is2ds["data_start_utc"] = np.datetime64(
-                    is2ds.data_start_utc.data[0].astype(str)[:-1]
-                )
-                is2ds["data_end_utc"] = np.datetime64(
-                    is2ds.data_end_utc.data[0].astype(str)[:-1]
-                )
+                if is2ds["data_end_utc"].data[0] == "Z":
+                    # manually remove 'Z' from datetime to allow conversion to np.datetime64 object (support for timezones is deprecated and causes a seg fault)
+                    is2ds["data_start_utc"] = np.datetime64(
+                        is2ds.data_start_utc.data[0].astype(str)[:-1]
+                    )
+                    is2ds["data_end_utc"] = np.datetime64(
+                        is2ds.data_end_utc.data[0].astype(str)[:-1]
+                    )
 
             except AttributeError:
                 pass
--- a/icepyx/core/read.py
+++ b/icepyx/core/read.py
@@ -363,7 +363,10 @@ class Read:
                 pass
 
             try:
-                if is2ds["data_end_utc"].data[0] == "Z":
+                if (
+                    hasattr(is2ds, "data_start_utc")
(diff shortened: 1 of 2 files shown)


icepyx/core/read.py
```diff
@@ -363,13 +363,18 @@ def _add_var_to_ds(is2ds, ds, grp_path, wanted_groups_tiered, wanted_dict):
...
 
             try:
-                # DevNote: these lines may cause a NumPy Warning, as explained here: https://numpy.org/doc/stable/release/1.11.0-notes.html?
-                # The below line will silence this warning...
-                # warnings.filterwarnings(
-                #     "default", category=DeprecationWarning, module=np.astype()
-                # )
-                is2ds["data_start_utc"] = is2ds.data_start_utc.astype(np.datetime64)
-                is2ds["data_end_utc"] = is2ds.data_end_utc.astype(np.datetime64)
+                if is2ds["data_end_utc"].data[0] == "Z":
```
weiji14, line 366: This fixed an AttributeError for me, but loading an ATL06 dataset still doesn't work. I'll do a bit more investigation first.

```suggestion
                if (
                    hasattr(is2ds, "data_end_utc")
                    and is2ds["data_end_utc"].data[0] == "Z"
                ):
```

weiji14 (commented):
Really think we need a unit test for the Read module, do you have time to add one in?

icepyx/core/read.py
JessicaS11: I pushed a modified version of this - it probably wasn't making it into the if statement because I forgot to compare just the last character...

JessicaS11:
> Does it need to merge this PR, with the other one ( https://github.com/icesat2py/icepyx/pull/281 )?

Yes - it turns out this was a standalone issue occurring for other datasets too (due to a deprecation in the most recent versions of numpy, v 1.22), so we gave it its own PR. Once we merge this into development, we can update #281 and fingers crossed that PR will be good to go too.

JessicaS11:
> Really think we need a unit test for the Read module, do you have time to add one in?

Agreed. I won't have time until after the hackweek. : )

icepyx/core/read.py
```diff
@@ -363,13 +363,21 @@ def _add_var_to_ds(is2ds, ds, grp_path, wanted_groups_tiered, wanted_dict):
...
-                # DevNote: these lines may cause a NumPy Warning, as explained here: https://numpy.org/doc/stable/release/1.11.0-notes.html?
-                # The below line will silence this warning...
-                # warnings.filterwarnings(
-                #     "default", category=DeprecationWarning, module=np.astype()
-                # )
-                is2ds["data_start_utc"] = is2ds.data_start_utc.astype(np.datetime64)
-                is2ds["data_end_utc"] = is2ds.data_end_utc.astype(np.datetime64)
+                if (
+                    hasattr(is2ds, "data_start_utc")
+                    and is2ds["data_start_utc"].data[0].astype(str)[:-1] == "Z"
```
weiji14, line 368: Try this instead.
```suggestion
                    and is2ds["data_start_utc"].data[0].astype(str).endswith("Z")
```

PR #307, opened 2022-05-13
Tracking docs

JessicaS11:
This PR aims to clarify some of the citation and tracking info on the docs pages. Specifically:
- clarify that "citation" provides an easy-to-use icepyx citation
- separate downloads (and traffic figures) from the list of groups and publications using/citing icepyx


JessicaS11:
@kelseybisson Please feel free to further modify/clarify these headings per our discussions on what new users might look for to find this info.

kelseybisson (approved):
I like the changes to the documentation & especially the explicit mention that IP addresses aren't tracked and we are deciding not to go that route!

Issue #318, opened 2022-06-09
Turn validate_inputs.py validation functions into classes


sjh1024:
From comment in validate_inputs.py:

`DevGoal: clean up; turn into classes (see validate_inputs_classes.py)`

The functions this seems to be referring to include:

* spatial
* temporal

From what I understand/can see from the code at this time, there wouldn't be much benefit in changing the following functions ABOVE the comment into classes:

* prod_version
* tracks
* cycles

This is because these functions (as far as I can tell) have much smaller ranges of possible values than the potentially infinite possible inputs for spatial/temporal inputs. But please comment if you disagree and I can add that to the issue as well!

sjh1024:
PR has been created for Spatial class, will be making a separate one for temporal class when completed.

sjh1024:
Issue is resolved with PRs #322 and #327

Issue #317, opened 2022-06-09
Make GenQuery spatial_extent close polygons automatically


sjh1024:
From comment in Query.py, in class GenQuery, regarding parameter `spatial_extent` :
*"DevGoal: adapt code so the polygon is automatically closed if need be."*

1. Ensure Query objects "close the polygon" if the first/last points are equal (i.e. change the "spatial_extent" property to have one more point equal to the first point)
2. Ensure that a warning is thrown to tell the user that the polygon given was NOT closed and that it was closed automatically.
3. Ensure that a unit test is written to test this functionality.

sjh1024:
Steps 1 and 2 are complete, will push change to branch `sjh1024` (and create PR) when tested.

sjh1024:
Step 3 is complete. Will be making a PR containing this fix + [#318](https://github.com/icesat2py/icepyx/issues/318) by end of this week.

Issue #321, opened 2022-06-23
Integrate "geospatial.py" module with new "validate_spatial" class


sjh1024:

sjh1024:
From @JessicaS11's comment on #318 -

It is worth thinking about how the spatial portion might be (potentially) combined with the geospatial.py module as well as utilizing existing libraries to the extent possible to validate and refine inputs. For instance, using Shapely:

```
import shapely.geometry as geom
bbox = [-55, 68, -48, 71]
b_bbox = [-55, 71, -48, 68]

fmtbox = geom.box(*bbox)
b_fmtbox = geom.box(*b_bbox)
```

Both fmtbox and b_fmtbox will output (-55.0, 68.0, -48.0, 71.0) as the valid, correctly formatted bounding box. In this case, we could eliminate completely the check + user warning that they entered the coordinates in the wrong order. However, this DOES NOT deal with https://github.com/icesat2py/icepyx/issues/151 (which I think is due to https://github.com/icesat2py/icepyx/issues/117) and would actually make cross-dateline queries more challenging, so we should keep that in mind before implementing this route (perhaps requiring the user to manually specify via keyword they intend to cross the dateline with their query and then handling the request accordingly?).

sjh1024:
Issue has been resolved with PR #322

PR #322, opened 2022-07-05
validate_inputs_spatial, automatic polygon closure functionality

sjh1024:
From Issue [#318](https://github.com/icesat2py/icepyx/issues/318):

* Created Spatial class in `validate_inputs_spatial.py` from `Spatial` function in `validate_inputs.py`
* Deleted this `Spatial` function from the original `validate_inputs.py` file
* Wrote test suite (`test_spatial.py`) for the new Spatial class; ensured all tests pass
* Updated `query.py` constructor for `GenQuery` to use new Spatial class

From Issue [#317]:
* Updated code to close polygons automatically if first and last points are not identical (connects last point to first point)
* Wrote unit tests to ensure this functionality behaves as expected.


icepyx/tests/test_spatial.py
```diff
@@ -0,0 +1,262 @@
+import pytest
+import warnings
+import datetime as dt
+import numpy as np
+from shapely.geometry import Polygon
```
JessicaS11, line 5: We try to follow the abc order convention for imports of common packages - could you please update this?

icepyx/core/query.py
```diff
@@ -104,11 +104,14 @@ def __init__(
...
     ):
         # validate & init spatial extent
-        self.extent_type, self._spat_extent, self._geom_filepath = val.spatial(
-            spatial_extent
-        )
+
+        sp_extent = sp.Spatial(spatial_extent)
+        self.extent_type = sp_extent.extent_type
+        self._spat_extent = sp_extent.spatial_extent
+        self._geom_filepath = sp_extent.extent_file
```
JessicaS11, line 111: since we're turning spatial into a class, it would be cool if we just attached the spatial object to the GenQuery object. So we might have something like `self.spatial = sp.Spatial(spatial_extent)` and then would access the needed info throughout `query.py` as `self.spatial.extent_type`, rather than having to manually set the variables here. Happy to work through variable names and which should be private (starting with `_`) with you.

JessicaS11:
thanks @sjh1024 - this is looking great! I left a few comments, and noticed that a bunch of the new tests are failing. Let me know if you need help debugging those.

A couple high level questions/comments:
1. I'm guessing the test polygon.gpkg needs to be in the test directory to be found - if not, could we just use the copy that is already docs; if so, could we move the file and update the links to it in the examples that use it?
2. Why not add the spatial class (including validation) directly to `geospatial.py`? I'm wondering if this will make it more streamlined when we get to #321 (so we're not introducing and then removing a new module).

JessicaS11:
@all-contributors add @sjh1024 for bug, code, doc, maintenance, test

icepyx/tests/test_spatial.py
sjh1024: updated with most recent push

icepyx/core/query.py
sjh1024: Resolved with recent pushes to the branch

JessicaS11:
> 2\. Why not add the spatial class (including validation) directly to `geospatial.py`? I'm wondering if this will make it more streamlined when we get to

Could you remind me what we decided on this? Thanks!

sjh1024:
I don't believe we made a final decision on the above comment. We can discuss it in today's meeting before moving forward. Thank you!

icepyx/core/query.py
```diff
@@ -19,8 +19,8 @@
 # QUESTION: why doesn't from granules import Granules as Granules work, since granules=icepyx.core.granules?
 # from icepyx.core.granules import Granules
 from icepyx.core.variables import Variables as Variables
-import icepyx.core.geospatial as geospatial
 import icepyx.core.validate_inputs as val
+import icepyx.core.spatial as sp
```
JessicaS11, line 23: Though technically you should not use `import scipy` over `from scipy import specific_function` ([e.g.](https://stackoverflow.com/questions/36014733/official-abbreviation-for-import-scipy-as-sp-sc)), if someone DOES `sp` is the most likely abbreviation they'll use. So I'd like to switch ours to `spat` or something to avoid any potential conflicts...

icepyx/core/query.py
```diff
@@ -303,6 +305,10 @@ def product_version(self):
         """
         return self._version
 
+    @property
+    def spatial(self):
+        return self._sp_extent
```
JessicaS11, line 310: We should add a docstring here

icepyx/core/query.py
```diff
@@ -329,12 +335,12 @@ def spatial_extent(self):
 
         """
 
-        if self.extent_type == "bounding_box":
-            return ("bounding box", self._spat_extent)
-        elif self.extent_type == "polygon":
+        if self._sp_extent.extent_type == "bounding_box":
+            return "bounding box", self._sp_extent.spatial_extent
```
JessicaS11, line 339:
```suggestion
            return ("bounding box", self._sp_extent.spatial_extent)
```

pretty sure this needs to be a tuple

icepyx/core/spatial.py
```diff
@@ -0,0 +1,317 @@
+import geopandas as gpd
+import os
+import warnings
+import numpy as np
+from shapely.geometry import Polygon
```
JessicaS11, line 5: reorder

icepyx/core/query.py
JessicaS11: Ignore this - I've added one in my current branch!

JessicaS11:
In query.py (circa lines 1010 and the next else), there is a line `geom_filepath=self._geom_filepath` that needs to be updated to `geom_filepath=self._spatial._geom_file`. GitHub won't let me comment directly on the non-altered portions of code.

JessicaS11 (approved):

PR #327, opened 2022-07-21
validate_inputs_temporal

sjh1024:
From Issue [#318](https://github.com/icesat2py/icepyx/issues/318):

* Created Temporal class in `validate_inputs_temporal.py` from `Temporal` function in `validate_inputs.py`
* Deleted this `Temporal` function from the original `validate_inputs.py` file  
* Wrote test suite (`test_temporal.py`) for the new `Temporal` class; ensured all tests pass (these tests replace the old `Temporal` function tests)
* Updated `query.py` constructor for `GenQuery` to use new `Temporal` class w/updated `@property` functions
* Updated other classes to use correct syntax when accessing `temporal` property of Query objects


sjh1024:
I suspect Travis CI will report test failures, awaiting those failures so I can attempt to resolve them (currently working on a different machine than my work machine, unsure how many failures are a result of this on my end).

icepyx/core/query.py
```diff
@@ -303,6 +300,25 @@ def product_version(self):
...
+        Examples
+        --------
+        >>> reg_a = Query('ATL06',[-55, 68, -48, 71],['2019-02-20','2019-02-28'])
+        >>> print(reg_a.temporal)
+        Start date and time: 2019-02-20 00:00:00
+        End date and time: 2019-02-28 23:59:59
+
+        """
+        return self._temporal
+
```
JessicaS11, line 321: Should we add a `if not hasattr...` statement here to let the user know if temporal parameters weren't set (e.g. as in `start_time`)?

icepyx/core/query.py
```diff
@@ -497,9 +513,9 @@ def CMRparams(self):
         # dictionary of optional CMR parameters
         kwargs = {}
         # temporal CMR parameters
-        if hasattr(self, "_start") and hasattr(self, "_end"):
-            kwargs["start"] = self._start
-            kwargs["end"] = self._end
+        if hasattr(self, "_temporal"):
+            kwargs["start"] = self._temporal.start
+            kwargs["end"] = self._temporal.end
```
JessicaS11, line 518: per our ongoing conversation about private variables - should these be `self._temporal._start` instead of `self._temporal.start`?

icepyx/core/query.py
sjh1024: This is a good thing to think about.

If it helps, the current state of the code does not have a start_time/end_time variable, nor does it seem to store any information if these were not passed in (a warning is just thrown to the user).

icepyx/core/temporal.py
```diff
@@ -0,0 +1,490 @@
...
+    Parameters
+    ----------
+    date: string
+        A string representation for the date value. Current supported date formats are:
+            * "YYYY-MM-DD"
+            * "YYYY-DOY"
+    Returns
+    -------
+    datetime object, representing the date from the string parameter.
+
```
JessicaS11, line 24: need to check docstrings for this new module:
- [ ] confirm ordering of entries
- [ ] add missing components (e.g. for properties)

JessicaS11 (approved):

PR #329, opened 2022-07-27
release 0.6.3

JessicaS11:


JessicaS11:
@sjh1024 I opened an issue (#330) for addressing the failed Travis PR test. We should be able to go ahead with the release despite that.

icepyx/core/read.py
```diff
@@ -448,6 +493,13 @@ def _combine_nested_vars(is2ds, ds, grp_path, wanted_dict):
         #     #     UserWarning,
         #     # )
 
+        try:
+            is2ds = _make_np_datetime(is2ds, "data_start_utc")
+            is2ds = _make_np_datetime(is2ds, "data_end_utc")
+
+        except AttributeError:
+            pass
```
sjh1024, line 501: Just to confirm, is it OK to except the AttributeError here and pass/continue if somehow bad input is passed into `_make_np_datetime()`? My understanding is that bad input would include something that could not be converted into an `np_datetime` object, or a DataFrame with incorrect dimensions? 

I assume if this try/except is reached and the AttributeError must be excepted, then the "unaltered" `is2ds` parameter is updated/returned. Just ensuring this is the expected behavior!

sjh1024 (commented):
This looks great! I only had one comment in `read.py`, and this was more of a "sanity check" to ensure all errors are handled properly. 

One other comment I have: For future releases, might we consider moving the function `Read._make_np_datetime` into our future `Temporal` class as a helper function? It may not be helpful since it seems to be specific to this Read class, but it's something to consider as we move forward with our new classes/helper functions.

sjh1024 (approved):

JessicaS11:
> One other comment I have: For future releases, might we consider moving the function `Read._make_np_datetime` into our future `Temporal` class as a helper function? It may not be helpful since it seems to be specific to this Read class, but it's something to consider as we move forward with our new classes/helper functions.

Great idea! Go for it.

icepyx/core/read.py
JessicaS11: Good sanity check. This is because is2ds will not always have "data_start_utc" and "data_end_utc" xarray dimension coordinates, depending on which of those nested groups/variables are read in to `is2ds` that round. In these cases, we don't need to worry about setting the type of the coordinates. If you can think of a more elegant/robust way to do this, I'm all ears.

Issue #159, opened 2021-01-09
Error querying complex polygons with many vertices


facusapienza21:
I am trying to use complex polygons to retrieve ATL06 data but I received a `syntax error`. Here is a fully contained example I made to illustrate this. The polygon I am using is one of the [drainage basins](https://icesat4.gsfc.nasa.gov/cryo_data/ant_grn_drainage_systems.php) of Greenland, with around 18.000 vertices. 

<img width="861" alt="Screen Shot 2021-01-08 at 11 51 57 PM" src="https://user-images.githubusercontent.com/39526081/104081300-80dd3d00-520c-11eb-9802-2e9cd674b2a7.png">

This same error happens if I input the polygon as a list of (lon, lat) or from a `.shp` file. If I simplify the number of vertices in my polygon using the method `simplify` from [Shapely](https://shapely.readthedocs.io/en/stable/manual.html) to a relatively small number of vertices, I don't receive any error. 

I know that currently this function is used in `APIformatting.py`, with tolerance in 0.05. I am wondering if the problem is that for some complex polygons this tolerance is not enough. In such cases, a good way to solve this could be to find a value of tolerance such that 1) the polygon is small enough to be able to been processed and 2) be as close as possible to the original polygon. Also, it is important in cases where the polygon is not large enough to keep the original polygon and do not use the `simplify` method. In cases where a modification of the original polygon has been made, a messege could be also useful for the user. 

I would love to help with this feature if it is required.

JessicaS11:
Hello @facusapienza21! Thanks for sharing your issue and digging in to where it's coming from. Could you provide the full error traceback (it's cut off in your screenshot, and I'd like to see exactly where it's coming from). icepyx actually uses both a simplified (for the initial query, to find potential data) and non-simplified version of your polygon (for the actual subsetting). How we move forward on this issue will depend on whether the error is coming from icepyx or from the NSIDC subsetter.

facusapienza21:
Hi @JessicaS11 ! Here is the full error message I get: 

<img width="1043" alt="Screen Shot 2021-01-11 at 10 20 45 PM" src="https://user-images.githubusercontent.com/39526081/104257118-719ffe80-545b-11eb-868f-a8586d5e12c2.png">

Let me know if you want me to try something else or I can help to solve the issue. Thank you!

JessicaS11:
[...]

PR #331, opened 2022-08-16
add cross-dateline flag and handling in creating geodataframe

JessicaS11:
Multiple issues note that icepyx functionality falters for spatial inputs that cross the dateline. One challenge is that - under the hood - we were using `shapely.geometry.Polygon` features to standardize our spatial inputs and manipulate coordinates. For spatial inputs that cross the dateline and are given in [-180,180] instead of [0,360], shapely was reordering the coordinates to avoid crossing the dateline. Here we add a `xdateline` flag that checks for likely cases and gives the user the ability to force a dateline crossing. Internally, this flag is used to convert longitudes between [-180,180] and [0,360]  depending on where the coordinates are being used (so the former for CMR and NSIDC submissions, the latter for GeoPandas and visualization).

ToDo:
- [x] add tests for new functions
- [x] add failing examples from issues as edge case tests
- [x] run all examples to make sure they run
- [x] follow up with NSIDC about submission issues for valid polygons that were errorring (UPDATE: an internal ticket ("bug report") has been filed to look into this)
- [x] fix formatting and properties as needed for QUEST (and to pass tests)


JessicaS11:
@all-contributors add @trevorskaggs for bug and code

JessicaS11:
Note: Travis tests won't pass for now because we were out of free (OSS) plan credits (as I learned when I went to see why the status wasn't reporting). I've put in a request for more.

JessicaS11:
A bunch of tests are still failing (so feel free to wait to review). Apparently VSCode's testing suite failed to actually pick up and run the tests properly (AGAIN), so it was telling me they were passing!

JessicaS11:
@weiji14 @sjh1024 It's finally ready!

weiji14:
Just wanted to say, I might not have time this week to review this big PR, but will look into it next week (ping me if I forget).

weiji14 (commented):
A few minor comments for now. Gonna look at the big changes in `icepyx/core/spatial.py` next!

icepyx/core/spatial.py
```diff
@@ -510,3 +599,86 @@ def extent_file(self):
...
+            if any(
+                geomtype in ["MultiPoint", "MultiLineString", "MultiPolygon"]
+                for geomtype in poly.geom_type
+            ):
+                poly = poly.convex_hull
+
+            poly = poly.unary_union
+
+            # Simplify polygon. The larger the tolerance value, the more simplified the polygon. See Bruce Wallin's function to do this
+            poly = poly.simplify(0.05, preserve_topology=False)
```
weiji14, line 640: Is this [Douglas-Peucker polygon simplification algorithm](https://shapely.readthedocs.io/en/1.8.5.post1/manual.html#object.simplify) for #159? I realize that getting an error is not nice, but am worried about unintended consequences of simplifying the polygon on behalf of the user. The tolerance of [0.05deg is about 5.6km](https://www.wolframalpha.com/input/?i=0.05deg+to+metres) which is not a small distance.

icepyx/core/spatial.py
```diff
@@ -510,3 +599,86 @@ def extent_file(self):
...
         """
         return self._geom_file
+
+    # ----------------------------------------------------------------------
+    # Methods
+
+    # TODO: can use this docstring as a todo list
+    def fmt_for_CMR(self):
+        """
+        Format the spatial extent for the CMR API.
```
weiji14, line 609:
```suggestion
        Format the spatial extent for NASA's Common Metadata Repository (CMR) API.
```

icepyx/core/spatial.py
```diff
@@ -272,33 +324,29 @@ def validate_polygon_list(spatial_extent):
...
-       ----------
-       spatial_extent: string
-                        A string representing a geospatial polygon file (kml, shp, gpkg)
-                        * must provide full file path
-                        * recommended for file to only contain 1 polygon.
-                            * if multiple polygons, only the first polygon is selected at this time.
+    Parameters
+    ----------
+    spatial_extent: string
+                     A string representing a geospatial polygon file (kml, shp, gpkg)
```
weiji14, line 344: At L361, it says `if spatial_extent.split(".")[-1] in ["kml", "shp", "gpkg"]:` which restricts the possible file types to KMLs, shapefiles and geopackages. However at L114, the code uses `gpd.read_file(spatial_extent)`, which should mean that any other OGR supported vector file format should work (e.g. GeoJSON, FlatGeoBuf, etc).

Do you think it's possible to remove the hardcoded file extension check at L361? Maybe a try-except statement to see if the file can be read directly into a `geopandas.GeoDataFrame`?

icepyx/core/spatial.py
```diff
@@ -32,38 +33,48 @@ def geodataframe(extent_type, spatial_extent, file=False):
...
-            spatial_extent[1],
-            spatial_extent[3],
-            spatial_extent[3],
-            spatial_extent[1],
-            spatial_extent[1],
-        ]
-        # DevGoal: check to see that the box is actually correctly constructed;
-        # have not checked actual location of test coordinates
+        if xdateline == True:
+            cartesian_lons = [i if i > 0 else i + 360 for i in spatial_extent[0:-1:2]]
```
weiji14, line 66: Trying to wrap my head around this logic. So negative longitude values get 360 added to them (e.g. -1 becomes 359), while positive longitude values remain intact (e.g. +1 stays as +1)?

Shouldn't it be that if any longitude values are negative (in the range -180 to 0), 180 gets added to the coordinates?

icepyx/core/query.py
```diff
@@ -61,6 +63,14 @@ class GenQuery:
         End time in UTC/Zulu (24 hour clock). If None, use default.
         TODO: check for time in date-range date-time object, if that's used for input.
 
+    Arguments
+    ---------
```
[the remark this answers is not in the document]
JessicaS11: Thanks for the terminology clarification!

icepyx/core/spatial.py
JessicaS11: > Is this [Douglas-Peucker polygon simplification algorithm](https://shapely.readthedocs.io/en/1.8.5.post1/manual.html#object.simplify) for addressing #159?

Not quite. There are multiple polygon submission issues going on, and this PR does not address #159. This function is only for the CMR step, i.e. during potential granule identification to get the user some info about available data. With cloud access, since we're using this step to actually get granule IDs, we could now be introducing a problem, but for on-prem data order/download, these results are never used and the polygon is not simplified when it's submitted in that step (and this is where #159 comes in). The SlideRule team has been working hard on these issues (there are also issues with the fact that metadata and actual content don't always align for each granlue, especially during off-pointing). After we've switched over to using earthaccess (currently still called earthdata), which is where this type of functionality (simplifying, buffering, etc.) should probably ultimately live anyway, I plan to implement some of the solutions SR has developed to handle some of these tiny and complex polygon issues.

For right now, this formatting is just being moved from APIformatting into the spatial module (and generalized/fixed to handle cross dateline issues).

icepyx/core/spatial.py
JessicaS11: I definitely had to do a few laps on this one too.

>Shouldn't it be that if any longitude values are negative (in the range -180 to 0), 180 gets added to all of the longitude coordinates?

No, because it's only the -180 to 0 (or 181 to 360) range of coordinates that need to be adjusted. If you added 180 to everything, you'd effectively be shifting the coordinates halfway around the world. So...

>So negative longitude values get 360 added to them (e.g. -1 becomes 359), while positive longitude values remain intact (e.g. +1 stays as +1)?

Exactly. It's counterintuitive, but has to do with the fact that Shapely (and GeoPandas) use a cartesian plane system (but not spherical geometry computations). Thus, in order for it to correctly handle something crossing the dateline, it needs coordinates in 0-360 space or will automagically "correct" the geometry to not cross the dateline.

icepyx/core/spatial.py
JessicaS11: Great points!

The file type list limitations is a combination of what [early] NSIDC tutorials explicitly stated were submissable and formats that I (as of 3+ years ago) was familiar with or confident testing. Theoretically, any "vector-based spatial data file" (as stated in [this notebook](https://github.com/nsidc/NSIDC-Data-Access-Notebook/blob/master/notebooks/Customize%20and%20Access%20NSIDC%20Data.ipynb) can be submitted to the ICESat-2 subsetter, which should jibe with:

> any other OGR supported vector file format should work (e.g. GeoJSON, FlatGeoBuf, etc)

I'm hesitant to remove the file-type restrictions until we have tried submitting (and ideally added to the test suite) at least a few of the other more common vector file formats. How would you feel about converting this to an issue for future work?

icepyx/core/spatial.py
weiji14: Wait, I'm getting confused. What is 0 in the 0-360 longitude range, is it at Greenwich or at the International date line? Maybe we need to add a proper unit test for this...

Trying to read https://gis.stackexchange.com/questions/226605/shapely-polygons-crossing-the-antimeridian and https://towardsdatascience.com/around-the-world-in-80-lines-crossing-the-antimeridian-with-python-and-shapely-c87c9b6e1513, and I'm wondering if [`shapely.affinity.translate`](https://shapely.readthedocs.io/en/1.8.5.post1/manual.html#shapely.affinity.translate) might be useful to simplify the (-180 to 180) to (0-360) conversion code (i.e. replace the list-comprehension with `shapely.affinity.translate(geom, xoff=180)`. But let's figure out the logic first :slightly_smiling_face:

doc/source/example_notebooks/IS2_data_access.ipynb
```diff
@@ -83,6 +83,9 @@
     "*NOTE: The input keyword for `short_name` was updated in the code from `dataset` to `product` to match common usage.\n",
     "This should not affect users providing positional inputs as demonstrated in this tutorial.*\n",
     "\n",
+    "*NOTE: You can submit at most one bounding box or a list of lonlat polygon coordinates per object instance.\n",
+    "Geospatial polygon files may contain multiple features (polygons).*\n",
```
JessicaS11, line 87:
```suggestion
    "Per NSIDC requirements, geospatial polygon files may only contain one feature (polygon).*\n",
```

icepyx/core/query.py
```diff
@@ -37,6 +37,8 @@ class GenQuery:
     spatial_extent : list of coordinates or string (i.e. file name)
         Spatial extent of interest, provided as a bounding box, list of polygon coordinates, or
         geospatial polygon file.
+        You can submit at most one bounding box or list of polygon coordinates.
+        Geospatial polygon files may contain multiple features (polygons).
```
JessicaS11, line 41:
```suggestion
        Per NSIDC requirements, geospatial polygon files may only contain one feature (polygon).
```

icepyx/core/spatial.py
weiji14: Yes, let's track the OGR vector file support in a separate issue. My impression is that we're not submitting the vector file (`.shp`, `.geojson`, etc) directly, but the WKT geometry parsed from the vector file, so any geopandas readable file should work in theory.

JessicaS11: See #384. During ordering, the file itself (rather than the WKT geometry) is passed.

icepyx/core/spatial.py
JessicaS11: It's crude (remarkably hard to find a map with both sets of coordinates), but visually:
image

JessicaS11: > I'm wondering if [`shapely.affinity.translate`](https://shapely.readthedocs.io/en/1.8.5.post1/manual.html#shapely.affinity.translate) might be useful to simplify the (-180 to 180) to (0-360) conversion code (i.e. replace the list-comprehension with `shapely.affinity.translate(geom, xoff=180)`

The problem with `shapely.affinity.translate` is that it requires a valid geometry input, and that's where the issue first pops up. Consider:

```
import shapely.geometry as geom
# across dateline example
bbox = [-179,60,179,62]
b_bbox = [179,-60,-179,-62] # this is "auto-corrected" to match bbox 

for box in [bbox, b_bbox]:
  fmtbox = geom.box(*box)
  print(fmtbox.bounds)
  
  poly = geom.Polygon.from_bounds(*box)
  print(poly)
```

versus: 
```
# 0 to 360 ex
bbox = [181,60,179,62] # this is auto-corrected to match b_bbox
b_bbox = [179,60,181,62] # this properly crosses the dateline
```

weiji14: Ah ok, so 0 in 0-360 is actually at Greenwich, not the International date line, which makes sense for solving the cross-dateline issue now that I think of it (my bias with NZ-centred world maps is confusing me :laughing:). In that case the if-then logic is sound, and `shapely.affinity.translate` won't work since it's not a simple translational shift.Thanks again for the clarification!

icepyx/core/query.py
```diff
@@ -37,6 +37,8 @@ class GenQuery:
     spatial_extent : list of coordinates or string (i.e. file name)
         Spatial extent of interest, provided as a bounding box, list of polygon coordinates, or
         geospatial polygon file.
+        You can submit at most one bounding box or list of polygon coordinates.
+        Per NSIDC requirements, geospatial polygon files may only contain one feature (polygon).
         Bounding box coordinates should be provided in decimal degrees as
         [lower-left-longitude, lower-left-latitute, upper-right-longitude, upper-right-latitude].
```
weiji14, line 43: Maybe add some extra notes for the cross-dateline case. Specifically, state that `xdateline` should be set to `True` if lower-left longitude is positive (>0) but upper-right longitude is negative (<0)? Also think we should explicitly mention something about the (-180 to 180) or (0-360) range here. Here's a draft sentence:

```suggestion
        Bounding box coordinates should be provided in decimal degrees as
        [lower-left-longitude, lower-left-latitute, upper-right-longitude, upper-right-latitude].
        Note: longitude values are assumed to go from -180 to +180, with 0 being the Prime Meridian
        at Greenwich. If your region of interest crosses the International Date Line (i.e. lower-left
        longitude is positive (>0) but upper-right longitude is negative (<0), please set `xdateline=True`.
```

weiji14 (approved):
Just 2 more notes on documentation. Feel free to merge after addressing those.

icepyx/core/spatial.py
JessicaS11: Thanks for checking the logic! Having to go through and explain it is always a good exercise to make sure we've got it right. : )

PR #372, opened 2022-09-22
Support Python 3.10

weiji14:
Set upper pin of Python to be 3.10. Original pin to <3.10 was set in 93012a4119d514e2aa3b7bb2777908dcaf551da5 (icepyx v0.6.0).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setuptools.setup(
     packages=setuptools.find_packages(exclude=["*tests"]),
     install_requires=INSTALL_REQUIRES,
     extras_require=EXTRAS_REQUIRE,
-    python_requires=">=3, <3.10",
+    python_requires=">=3, <3.11",
     classifiers=[
         "Programming Language :: Python :: 3",
         "License :: OSI Approved :: BSD License",


setup.py
```diff
@@ -26,7 +26,7 @@
     packages=setuptools.find_packages(exclude=["*tests"]),
     install_requires=INSTALL_REQUIRES,
     extras_require=EXTRAS_REQUIRE,
-    python_requires=">=3, <3.10",
+    python_requires=">=3, <3.11",
```
weiji14, line 29: @JessicaS11, could you remind me what was the reason an upper pin was set in 93012a4119d514e2aa3b7bb2777908dcaf551da5? Just wanted to know because Python 3.11 will be out in about a month (see https://peps.python.org/pep-0664/#schedule), so we would need to update the upper pin again soon.

weiji14: Tempted to set the upper pin to <Python 3.12 since Python 3.11 will be out soon (24 Oct 2022) and many geo-packages like shapely and fiona already have wheels for Python 3.11.

Alternatively, we could remove the upper pin altogether, and not worry about these compatibility issues.

JessicaS11: All I can find/remember is that there was a major issue with the conda-forge feedstock that something (if memory serves, a dependency that needed to update for more recent versions of Python) couldn't be imported and was causing the builds to fail unless we pinned to an older version. I'm happy for us to remove the upper pin and see if we're a go again.

weiji14: Ok, I've removed the upper pin in f4128ec45597b1636c5e6ebc632cd79b65ef2c9b. Once a new release is made, I'll update things on conda-forge too (e.g. at https://github.com/conda-forge/icepyx-feedstock/pull/9).

JessicaS11 (approved):

PR #381, opened 2022-10-19
update cloud tutorial to work with publicly accessible cloud data

JessicaS11:
The goal here was to get the data access notebook working in time for the Oct 2022 IS2 Science Team meeting. This PR updates granules to get the s3 urls directly from the CMR metadata rather than constructing them manually, and the CMR metadata are queried specifically for cloud data (which is identical to the on-premises data that you would download from NSIDC). It does not yet address simplifications to the authentication process (planned with the introduction of [earthaccess](https://github.com/nsidc/earthdata/tree/main), currently still packaged as earthdata).

NOTE: this PR will need to be merged after #331, at which point this branch should be selectively rebased to streamline the review).

### DIFF
--- a/icepyx/core/query.py
+++ b/icepyx/core/query.py
@@ -908,7 +908,7 @@ class Query(GenQuery):
         self._email = email
 
     # DevGoal: check to make sure the see also bits of the docstrings work properly in RTD
-    def avail_granules(self, ids=False, cycles=False, tracks=False, s3urls=False):
+    def avail_granules(self, ids=False, cycles=False, tracks=False, cloud=False):
         """
         Obtain information about the available granules for the query
         object's parameters. By default, a complete list of available granules is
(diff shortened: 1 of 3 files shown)


icepyx/core/query.py
```diff
@@ -908,7 +908,7 @@ def is_ec2():
         self._email = email
 
     # DevGoal: check to make sure the see also bits of the docstrings work properly in RTD
-    def avail_granules(self, ids=False, cycles=False, tracks=False, s3urls=False):
+    def avail_granules(self, ids=False, cycles=False, tracks=False, cloud=False):
```
JessicaS11, line 911: @tsutterley @jpswinski Would this be a breaking change for SlideRule? My aim here was to make it more generic (since it's needed for authentication as well as directing data access), but there's no reason the keyword can't stay as `s3urls` if it prevents issues downstream.

PR #387, opened 2022-10-31
update options file for testing

### LINKED ISSUE #330: update tests to use version 5 products
Tests behind the Earthdata login are failing on Travis (as in #329) because they're using version 004, which is no longer available. To fix this we'll need to:
- [] update the tests/ATl06vxx_options.json
- [] update any tests that use this json to use a current product version

JessicaS11:
closes #330


JessicaS11:
Note: the file (`icepyx/tests/ATL06v05_options.json`)WAS updated, but there were no actual changes so it only tracks in git as a change of filename.

PR #390, opened 2022-12-01
update oa viz for polygons after spatial module change

JessicaS11:
OpenAltimetry visualizations could not be requested for polygon spatial extents due to a missed type change required by the introduction of the spatial module and cross-dateline handling in #331.

The bug was originally documented [here](https://github.com/CryoInTheCloud/CryoCloudWebsite/pull/35#issuecomment-1328613260), and an inline fix can be found [here](https://github.com/CryoInTheCloud/CryoCloudWebsite/blob/63f29c171d02f11c1ccc18eba750305463263dd3/book/tutorials/IS2_cloud_data_integration.ipynb).


icepyx/core/visualization.py
```diff
@@ -159,7 +159,9 @@ def __init__(
             self.bbox = query_obj.spatial.extent
 
         else:
-            mrc_bound = query_obj.spatial.extent.minimum_rotated_rectangle
+            mrc_bound = (
+                query_obj.spatial.extent_as_gdf.geometry.unary_union.minimum_rotated_rectangle.bounds
```
weiji14, line 163: Probably want the `shapely.geometry` object rather than the tuple (`minx`, `miny`, `maxx`, `maxy`).
```suggestion
                query_obj.spatial.extent_as_gdf.geometry.unary_union.minimum_rotated_rectangle
```
Alternatively, it might also possible to use `lonmin, lonmax, latmin, latmax = query_obj.spatial.extent_as_gdf.geometry.total_bounds` if you want to refactor the code from L162 to L169.

JessicaS11: Great catches @weiji14, as usual. Thank you!

JessicaS11:
I'm still working on debugging the tests. The failure is coming from ATL13 queries that return no data (at the CMR query stage).

PR #410, opened 2023-03-13
transition icepyx to using earthaccess library for authentication

JessicaS11:
The long awaited earthaccess (formerly earthdata) python library provides an easy interface for (among other things) authenticating with the NASA EarthData system. This PR remove internal handling of authentication through adoption of earthaccess. The user should see no changes, but this move greatly simplifies things behind the scenes and provides enhanced authentication capabilities for cloud data access.

This work is being done in collaboration with the earthaccess development team, and adoption by icepyx is helping find bugs and drive improvements there (e.g. https://github.com/nsidc/earthaccess/pull/214).


JessicaS11:
Noting that all tests pass currently, but I am about to pin the version of earthaccess to the upcoming release (which thus doesn't yet exist on conda-forge), so they will begin failing because the environment won't be create-able.

doc/source/example_notebooks/IS2_data_read-in.ipynb
```diff
@@ -183,7 +233,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "path_root = '/full/path/to/your/data/'"
+    "path_root = '/Users/jessica/computing/icepyx/github_repo/icepyx/doc/source/example_notebooks/atl11/'"
```
betolink, line 236: Is this hard-coded path intended or just an artifact of running it locally?

icepyx/core/query.py
```diff
@@ -854,58 +849,55 @@ def show_custom_options(self, dictview=False):
...
-
-            # loosely check for AWS login capability without web request
-            assert (
-                is_ec2() == True
-            ), "You must be working from a valid AWS instance to use s3 data access"
-            capability_url = "https://data.nsidc.earthdatacloud.nasa.gov/s3credentials"
+        """
 
-        self._session = Earthdata(uid, email, capability_url).login()
+        auth = earthaccess.login(**kwargs)
```
betolink, line 888: These are not strictly used by earthaccess but I assume is to keep the API the same for Icepyx right?

icepyx/core/query.py
```diff
@@ -854,58 +849,55 @@ def show_custom_options(self, dictview=False):
...
+            self._session = auth.get_session()
 
-        # DevNote: might make sense to do this part elsewhere in the future, but wanted to get it implemented for now
         if s3token == True:
-            self._s3login_credentials = json.loads(
-                self._session.get(self._session.get(capability_url).url).content
+            self._s3login_credentials = auth.get_s3_credentials(daac="NSIDC")
+
+        if uid != None or email != None:
+            warnings.warn(
```
betolink, line 897: Awesome!! (the deprecation warning)

betolink (approved):
I think this is a great PR, the code is simpler with earthaccess (I'm biased). One thing I'm probably missing is how Icepyx will work with ATL data in S3 buckets, especially subsetting. Not related to this PR but I think that down the road either Icepyx or earthaccess should interface with [Harmony](https://github.com/nasa/harmony-py) and transition away from on-prem based services like EGI.

icepyx/core/query.py
JessicaS11: Thanks!

JessicaS11:
> One thing I'm probably missing is how Icepyx will work with ATL data in S3 buckets, especially subsetting.

You're missing it because it doesn't exist yet! This is one of the things I will tackle next. I'd like to think some more about how icepyx can further leverage earthaccess and/or Harmony in combination with the Variables module to (as you suggest) think about moving away from EGI. Some of these things will hopefully be a lot easier now that the ICESat-2 data is publicly available on s3, especially once all the cloud metadata is included in CMR (I noticed last week that for ATL15 the s3 urls were not yet available from the initial CMR query). I'm excited for the possibilities!

icepyx/core/query.py
JessicaS11: the **kwargs? I put this in so that if a savvy user wanted to pass `persist=True` or one of the other earthaccess login keywords they could.

doc/source/example_notebooks/IS2_data_read-in.ipynb
JessicaS11: Great catch - I will fix it!

Issue #114, opened 2020-08-11
large files in git history


JessicaS11:
> Hi @JessicaS11, can't help but notice that you added two Tandem X DEM files in https://github.com/icesat2py/icepyx/pull/95/commits/aa6b5fb49ef5635e0428290ed5ba2cc943fabab8 to the repository which are used for the 'ICESat-2_DEM_comparison_Colombia_working.ipynb' tutorial. However, they're rather big files (25.7MB and 8.65MB), and git isn't usually meant for storing data files like this, so I just wanted to make sure (before you make any new commits) that this wasn't accidental/intentional?

_Originally posted by @weiji14 in https://github.com/icesat2py/icepyx/pull/95#issuecomment-659309089_

Thanks for pointing this out, @weiji14. I think these files have been present for awhile, but they were moved in the commit you point out. In looking into it, there are a few hdf5 files that are much larger that were accidentally uploaded early on and obviously now show up in the git history too. From my experience, the only way to get rid of them is to rewrite the git history. While I think that would ultimately be a good move (maybe we could even clean up the early commit history in the process), I'm hesitant to do so because of the downstream effects if every user does not rebase and remove them from their history, risking reintroduction (I had to remove some large files from another repo, and this was a challenge with a much smaller group). Or if I force change the master and development branches would we be able to catch this at the PR level if someone submits a PR with a branch that contains the large files (though presumably this would require an extra reviewer step to check the git history of the submitted PR)? Do you know if there are any alternative options or other ways to proceed given the group nature of the project (and entry-level git usage)?

JessicaS11:
@scottyhq Any ideas, since you helped solve this issue in the hackweek repos?

scottyhq:
@JessicaS11 - yes I'd recommend scrubbing big files from the commit history, which would mean contributors need to re-sync with this upstream repository. It's a bit disruptive, but better than a bloated repository. 

I see the following big files (using second answer here https://stackoverflow.com/questions/10622179/how-to-find-identify-large-commits-in-git-history)
```
84124d0a539d  3.9MiB doc/examples/ICESat-2_ATL08Parse_and_Visualize.ipynb
4ab8cf835487  8.0MiB dev-notebooks/ICESat-2Hackweek_tutorial_locations_draft.ipynb
c17285052462  8.6MiB examples/supporting_files/TDM1_DEM_90m_colombia_DEM_masked_aea_hs.tif
[...]

PR #431, opened 2023-07-28
add a few tests for Variables class

rwegener2:
## What was Changed

This PR adds a few tests for the Variables class. It doesn't cover all the available arguments, but it adds the following tests:

1. the `.avail()` function, reading from a local file
1. the `.avail()` function, using an order result
1. the `.append()` function with the `defaults` flag set to True
1. the `.remove()` function for all
1. the `.append()` function when given a `vars_list` argument

## How it was Changed

A `test_variables.py` file already existed, so I added additional tests there. For expected outputs longer than a few lines the output is saved in a `txt` or `json` file within the `expected_outputs` folder. These are read into a Python object during the test and compared to the `icepyx` produced result.

For the test using a local file I saved the local file in a `test_data` folder. I used an ATL06 product because it was small (less than 8MB). The `Variables` object that was created using this dataset was created as a `pytest.fixture`, to avoid repeating the class initiation code multiple times.

## How to test it

I actually couldn't get all the tests to run, so I was only testing the `test_variables.py` file. To run the tests I moved into the `tests` folder and ran `pytest test_variables.py`.

### Earthdata Login

In [line 53](https://github.com/icesat2py/icepyx/blob/test_variables/icepyx/tests/test_variables.py#L53) of `test_variables.py` I login to NASA Earthdata. This works because I have a `.netrc` file saved in my environment, so anyone else who runs these tests will need to have that setup as well.  It looks like we maybe have some testing CI setup, so I assume that this method of logging in will fail. How do we deal with logins in CI right now? Do we already have CI EDL credentials stored somewhere?

### Assuming a Query will return consistently

In [line 57](https://github.com/icesat2py/icepyx/blob/test_variables/icepyx/tests/test_variables.py#L57) of `test_variables.py` there is a sort of awkward check to make sure that the Query only returned one granule. I can't come up with any real reason why that would ever change, but I wasn't sure. Do we feel quite certain that past granules (from 2019) won't be added (in which case I'd remove the if statement), or is it worth it to keep that check?


icepyx/tests/test_variables.py
```diff
@@ -1,6 +1,27 @@
...
+from icepyx.core.variables import Variables
+from icepyx.core.query import Query
 import icepyx.core.variables as variables
 
 
+@pytest.fixture()
+def variables_obj_atl06():
+    return Variables(
+        "file", 
+        path='./test_data/processed_ATL06_20191130112041_09860505_006_01.h5',
```
weiji14, line 14: Hmm, we'll need to see if it's possible to avoid storing this 8MB HDF5 file in git, because it will slow down `git clone` for future developers. There's actually some work to prevent big files from being uploaded at #143, and we had some nasty work a few years ago to remove HDF5 files at #114.

There are some tests in the `test_behind_NSIDC_API_login.py` file which actually downloads data from NSIDC. Maybe it is possible to reuse some functionality from there?

icepyx/tests/test_variables.py
```diff
@@ -1,6 +1,27 @@
...
+
+from icepyx.core.variables import Variables
+from icepyx.core.query import Query
 import icepyx.core.variables as variables
 
 
+@pytest.fixture()
+def variables_obj_atl06():
+    return Variables(
+        "file",
```
weiji14, line 13: Is it possible to use `vartype="order"` instead of `vartype="file"` to dynamically download the file?

icepyx/tests/test_variables.py
```diff
@@ -22,3 +43,60 @@ def test_list_of_dict_vals():
...
 
     assert obs == exp
+
+def test_avail_from_file(variables_obj_atl06, avail_vars_atl06):
+    assert variables_obj_atl06.avail() == avail_vars_atl06
+
+
+def test_avail_from_order(avail_vars_atl06):
+    region = Query('ATL06',[-45, 74, -44,75],['2019-11-30','2019-11-30'], \
+                       start_time='00:00:00', end_time='23:59:59')
```
weiji14, line 53: Recommend to set the `version` here, because the returned schema can change from version to version.
```suggestion
    region = Query(
        product="ATL06",
        spatial_extent=[-45, 74, -44, 75],
        date_range=["2019-11-30", "2019-11-30"],
        start_time="00:00:00",
        end_time="23:59:59",
        version="006",
    )
```

weiji14 (commented):
Hi @rwegener2! First off, thanks for all your great work on icepyx recently! Really nice to see more unit tests added to icepyx because it really improves the long term sustainability and quality of the software :smile:

Taking a look at your code, the implementation of the tests are pretty good, and it looks like you've covered most of the public methods (those not starting with an underscore (e.g. `_check_valid_lists`). Ideally though, we'll need to figure a way to avoid storing the 8MB HDF5 file in git, hopefully via a dynamic download step, or some alternative way. See specific comments below.

Issue #433, opened 2023-07-31
Modularizing authentication


rwegener2:
## The Problem
The authentication method that exists right now is inherently tied to the `Query` class because the `.earthdata_login()` method is a method on that class. That has led to passing `._session` values from `Query` objects to `Variables` objects as needed.

As the cloud read effort develops, additional modules, such as `Read`, will also need to authenticate. Because of this I would like to propose that we make authentication in the repository more modular, so it can be more easily extended to different classes.

## Possible Solutions

Both of the ideas below create an `EarthdataAuth` class. This class would have the existing `earthdata_login()` functionality within it.

### 1 - Inheritance

We could create an class called `EarthdataAuth` that holds the `._auth`, `._session`, and `._s3login_credentials` attributes that currently get created in `Query`. `Query`, `Variables`, and other classes could then be subclasses of the `EarthdataAuth` superclass.

**Pros/Cons:** After some reading it seems that isn't exactly the type of relationship we want to convey. The set of complexity added by a new level of inheritance feels disproportionate to the fairly simple need to keep track of 2-3 attributes and one function. The pro of this solution, however, is that I believe we could it up such that nothing about the user experience changes.

### 2 - Creating `EarthdataAuth` as a class that other objects have as an attribute

`Query` and other related objects could have an `.auth` attribute that is an instance of `EarthdataAuth`. `EarthdataAuth` can get created either by calling the `earthdata_login()` function, or automatically when a user needs authentication. The user (or the internal code) would access attributes like `ipx.Query.auth._session`.

**Pros/Cons:** I think this solution could simplify our internal code and at least for me it is more intuitive than inheritance (that may just be me, though). The con of this solution is that it would change how the user authenticates, either by removing that step all together or by requiring they switch to `ipx.Query.auth.earthdata_login()`.

## Questions

- Any opinions about which of the two options to pursue? I am leaning toward the second one, but could be otherwise convinced. @JessicaS11 
- Is there a reason that we prefer to ask users to run the `earthdata_login()` function themself? Are we interested in setting up the workflow such that the user is prompted for login credentials only when the execute a function that requires authentication? I think that could happen fairly easily on our end, but it would change the user experience.

## Code

[...]

PR #434, opened 2023-08-01
WIP expand Variables class to handle s3 urls from NSIDC

rwegener2:
## What was done

At the moment the `Variables` class `.avail()` method would fail if given an s3 url. This PR expands the class to handle this case.

## How it was done

In addition to `file` and `order` a new `vartype`, `nsidc-s3`, was added to the class. `nsidc` was specified (instead of just `s3`) because extracting the product and version from the filepath relies on the nsidc naming structure. So, you couldn't just give `Variables` the s3 path to any IS2 file anywhere on AWS; it needs to be from the `nsidc-cumulus-prod-protected` bucket. If others have feedback on this decision I'm happy to hear. We could alternately try to extract the product and version from the metadata, which would remove this limitation.

Within the `.avail()` method the `nsidc-s3` filetype uses the same variable extraction strategy as `order`. An interesting note about this is that if we use the `order` method of grabbing variables it happens very quickly.  It's the time to ping an API endpoint and parse the request. The `.avail()` function gets very slow with cloud data when trying to walk the full file (as happens when using the `file` argument), but we can avoid that by using the `order` method of extracting variables instead.

## Todo / Next steps

This PR is pretty close, but blocked by determining how to pass authentication (#433). After auth is addressed this PR can be merged into the `getvars` branch, where work has already started changing the way the `Variables` class is called in `Read` and `Query`.

- [ ] add test case
- [x] update after authentication PR
- [x] add `parse` as a library requirement


icepyx/core/variables.py
```diff
@@ -72,6 +74,14 @@ def __init__(
         elif self._vartype == "file":
             # DevGoal: check that the list or string are valid dir/files
             self.path = path
+        elif self._vartype == "nsidc-s3":
+            # Grab metadata from s3 path
+            template = ('s3://nsidc-cumulus-prod-protected/ATLAS/{product}/{version}/'
```
weiji14, line 79: This template looks to be hardcoded to the non-gridded ATLAS datasets like ATL06. Other gridded products (e.g. ATL14, ATL16, ATL20) would have a different template. E.g. looking at https://search.earthdata.nasa.gov/search?ff=Available%20in%20Earthdata%20Cloud&fi=ATLAS&gdf=HDF&fst0=Cryosphere&lat=65.08299573518599&long=-25.69921875&zoom=5:

| Dataset | Sample path |
|--|--|
| ATL06 | s3://nsidc-cumulus-prod-protected/ATLAS/ATL06/006/2023/04/16/ATL06_20230416235213_04061911_006_02.h5 |
| ATL07 | s3://nsidc-cumulus-prod-protected/ATLAS/ATL07/005/2022/10/12/ATL07-02_20221012220720_03391701_005_01.h5 |
| ATL10 | s3://nsidc-cumulus-prod-protected/ATLAS/ATL10/005/2022/10/12/ATL10-01_20221012220720_03391701_005_01.h5 |
| ATL11 | s3://nsidc-cumulus-prod-protected/ATLAS/ATL11/005/2022/03/27/ATL11_006305_0315_005_03.h5 |
| ATL14 | s3://nsidc-cumulus-prod-protected/ATLAS/ATL14/002/2019/ATL14_IS_0314_100m_002_01.nc |
| ATL16 | s3://nsidc-cumulus-prod-protected/ATLAS/ATL16/004/2022/ATL16_20220722003637_04601601_004_01.h5 |
| ATL20 | s3://nsidc-cumulus-prod-protected/ATLAS/ATL20/003/2022/ATL20-01_20220901002201_10861601_003_01.h5 |

Would it be possible to generalize this code to both non-gridded and gridded products?

rwegener2: Ah, thanks for catching this @weiji14! I'll work on a fix and let you know when I'm ready for review!

rwegener2: A question for @JessicaS11 and @weiji14 -- What do we think of getting the version and product name from inside the file instead of parsing it from the filename? I've only checked a handful of products, but those fields seem to be available in top-level metadata in a consistent way. I've been trying to parse those things out of the filename, which is how I believe it is also done elsewhere in the module, but this limits the files icepyx can process to those named in a very specific way. If we grab product/version from inside the file we are able to process more files (ex. cloud icesat-2 files not in nsidc bucket, or local files that have had their name changed). Thoughts?

JessicaS11: The discussion summarized in https://github.com/icesat2py/icepyx/pull/438#issuecomment-1699753154 indicates our intention to move away from requiring the user provide the product as input (unless they are also feeding in a directory containing files from multiple products). This should address the template issues noted here.

rwegener2:
This PR prompted two prior PRs: #444 and #451. I think those updates will are exciting for enabling the goal of this PR, allowing the Variables class to list s3 data variables. The changes in those two prior have made the approach to this goal is quite different than when this PR was opened (it was more complicated back then!). As a result I'm going to close this PR and open a new one to pursue adding s3 url reads to the Variables class.

PR #435, opened 2023-08-03
Refactor authentication

rwegener2:
## What is done

This PR refactors Earthdata authentication to make it more extensible to additional classes that may want to authenticate. Related discussion #433 

## How it was done

Here I'm continuing the discussion from #433. @JessicaS11 made the good point that instead of creating our own EarthdataAuth class we should just try to reference the `auth` object that comes back from `earthaccess.login()`. (Ex. by setting a `self.auth = earthaccess.login()` variable on all the classes that require authentication.  Generally like this:

```python
def Query():
    def __init__():
        self.auth = earthaccess.login()
```

I experimented with this but eventually realized the library would try to regenerate the objects very time `self.auth.get_s3_credentials(daac="NSIDC")` or `self.auth.get_session()` was accessed. For `session` that wouldn't matter so much because that is fast to access, but the s3 creds takes ~5 seconds to retrieve. This seemed like it might be a bit cumbersome over time. We could do separate attributes (`self.session` or `self.s3login_credentials`) and a `@property` getter method to get around this, but that was starting to feel like a lot of code to copy and paste into every class that needed auth. The other major problem is that I don't think we want to make users login when the Query object is created. I think we'd like to let them run Queries and search for data, and only require login once the user tries to `.order_granules()`.

To avoid these challenges I created a mixin class. This differs from an inherited class in that it is built to extend the functionality of other classes with a small, specific set of attributes or functions. The `EarthdataAuthMixin` contains properties for `auth`, `session`, and `s3login_credentials`. These set themselves (login) the first time they are accessed, and reference the previously created object in subsequent accesses. A bonus of this mixin method is that I was able to keep the `.earthdata_login()` method, which should make it backwards compatible. The mixin could be added to any class that needs authentication and the user doesn't need to keep track of if s3 credentials or a regular session are needed.

I'm curious what others think of this technique.  It's a half step more complicated than the `self.auth=earthaccess.login()` technique. It felt to me like we gained in modularity and functionality enough to warrant the complication, but I'd love additional opinions.

## How it can be tested

Both the following code snippets should work:

```python
import icepyx as ipx

region_a = ipx.Query('ATL06',[-45, 74, -44,75],['2019-11-30','2019-11-30'], \
                           start_time='00:00:00', end_time='23:59:59')

region_a.order_vars.avail()   # works without logging in!

region_a.order_granules()  # also works without logging in!

# try running these multiple times to notice that they store the previously created objects. Also notice that
# once `.session` creates the auth object, `.s3login_credentials` doesn't try to repeat logging in
region_a.session
region_a.s3login_credentials

region_a.earthdata_login()  # still works, even though it's not needed
```
```python

var_obj = Variables("file", path='./data/ATL06/processed_ATL06_20191201105502_10010505_006_01.h5',
          product='ATL06')

var_obj.auth
var_obj.session
```

## Todo

- [x] how can we allow the user to regenerate new s3 creds after an hour has passed? Right now there isn't a way to do that.
- [x] tests
- [x] documentation - examples
- [x] documentation - api reference


JessicaS11:
@betolink I'm curious if you have any thoughts on this approach for centralizing auth in icepyx so that it can easily be picked up across Query, Variables, and Read objects. Pulling out some relevant points from @rwegener2's great writeup above:

> we should just try to reference the auth object that comes back from earthaccess.login(). (Ex. by setting a self.auth = earthaccess.login() variable on all the classes that require authentication

but...

> the library would try to regenerate the objects very time self.auth.get_s3_credentials(daac="NSIDC") or self.auth.get_session() was accessed. For session that wouldn't matter so much because that is fast to access, but the s3 creds takes ~5 seconds to retrieve. This seemed like it might be a bit cumbersome over time.

She developed this Mixin approach, but we're still left with (assuming this isn't still outstanding in `earthaccess` too)

> how can we allow the user to regenerate new s3 creds after an hour has passed? Right now there isn't a way to do that.

Do you have any thoughts?

betolink:
Hi @JessicaS11 @rwegener2 this is still an outstanding issue with `earthaccess` the library tries to optimize and cache the `s3fs` [session](https://github.com/nsidc/earthaccess/blob/main/earthaccess/store.py#L170C18-L170C18) but we haven't implemented anything like this at the `get_s3_credentials()` level yet.

For icepyx, this mixing class looks good! maybe when we instantiate the class for the first time we could store the timestamp and return the same cached set of credentials until the time delta goes above say 50 minutes, a little like the `get_s3fs_session()` method.

rwegener2:
Thanks for the feedback @betolink!

I've cleaned up the PR and added:
- some logic that refreshes the s3 token after an hour
- some logic that checks if `earthaccess.login()` succeeded, and raises an error if it did not

The one question I had for @betolink:
- In the event of failure does `earthaccess.login()` return the error message anywhere other than printing it to output? If it does I can pass that along to the user, but if not I think printed output is fine.

We can talk about if it makes sense to merge this / how the best way to that.

icepyx/core/variables.py
```diff
@@ -3,6 +3,7 @@
 import pprint
 
 import icepyx.core.is2ref as is2ref
+from icepyx.core.auth import EarthdataAuthMixin
```
JessicaS11, line 6:
```suggestion
from icepyx.core.auth import EarthdataAuthMixin
import icepyx.core.is2ref as is2ref
```
alphabetical order

icepyx/core/query.py
```diff
@@ -22,6 +21,7 @@
 import icepyx.core.spatial as spat
 import icepyx.core.temporal as tp
 from icepyx.core.visualization import Visualize
+from icepyx.core.auth import EarthdataAuthMixin
```
JessicaS11, line 24:
```suggestion
from icepyx.core.auth import EarthdataAuthMixin
import icepyx.core.spatial as spat
import icepyx.core.temporal as tp
from icepyx.core.visualization import Visualize
```
use alphabetical order convention

icepyx/core/auth.py
```diff
@@ -0,0 +1,118 @@
...
+    def __str__(self):
+        if self.session:
+            repr_string = "EarthdataAuth obj with session initialized"
+        else:
+            repr_string = "EarthdataAuth obj without session initialized"
+        return repr_string
+
+    @property
+    def auth(self):
+        # Only login the first time .auth is accessed
```
JessicaS11, line 32: It would be great if we could add some basic docstrings for these functions (and then also add this new module to the docs).

icepyx/core/auth.py
```diff
@@ -0,0 +1,147 @@
...
+            self._session = auth.get_session()
+
+        if s3token == True:
+            self._s3login_credentials = auth.get_s3_credentials(daac="NSIDC")
+
+        if uid != None or email != None:
+            warnings.warn(
+                "The user id (uid) and/or email keyword arguments are no longer required.",
+                DeprecationWarning,
+            )
```
JessicaS11, line 147: Just noticed in a hackweek tutorial that this warning doesn't show up even if someone supplies email or user id - wondering if it's because of the earthaccess prints resulting from logging in and if putting this check first would make it show up when run in a notebook.

rwegener2: I looked into this and it turns out we need to add a `stacklevel=2` argument. [The docs](https://docs.python.org/3/library/warnings.html#warnings.warn) says:
> The stacklevel argument can be used by wrapper functions written in Python ... [which] makes the warning refer to deprecation()’s caller, rather than to the source of deprecation() itself

To be honest I don't understand _exactly_ what that means, but generally it sounds like we need the argument to push the warning all the way up the stack to the user, as right now it's getting stuck somewhere in between the source code and the user.

rwegener2:
#### Current status

1. All code comments have been addressed
2. I’ve written a few tests for the class

In writing those tests I did also end up changing a few lines of `test_behind_NSIDC_API_login.py`. There was a fixture in that class that writes a .netrc file. I moved that block to the pytest configuration so that the .netrc gets written before all the tests and can also be used by the new authentication class.

#### Todo

1. update docs

I’m going to pause on this PR for the rest of the day, but the last thing I’ve got on the todo list is updating the documentation to reflect that the user is no longer required to explicitly login.

rwegener2:
Ok, I think this PR is basically ready for a full review! I have just one doubt:

- I wasn't able to get the docs to auto build the API reference for the `EarthdataAuthMixin`. I see the `query.rst`, `read.rst`, etc. files, but how could I create one of those for the new class? Are they programmatically generated, or written out by hand?

I've still gone ahead and re-requested review; I'll get the API reference pushed as soon as I hear back. Thank you!

doc/source/example_notebooks/IS2_data_access.ipynb
JessicaS11, line 1: Update the numbering in the "quick-start guide"

JessicaS11: In the Earthdata login section, there's a typo ("featuers"). The admonition has nearly identical text the line after it we'll need to remove.

doc/source/user_guide/documentation/icepyx.rst
```diff
@@ -23,4 +23,5 @@ Diagrams are updated automatically after a pull request (PR) is approved and bef
    query
    read
    quest
+   auth
```
JessicaS11, line 26: Wondering at the "level" auth should be given within the docs. It seems like it might be more appropriate to put it into the components list (via `components.rst`). Thoughts?

icepyx/core/auth.py
```diff
@@ -0,0 +1,152 @@
...
+    
+    The class can be created without any initialization parameters, and the properties will
+    be populated when they are called. It can alternately be initialized with an 
+    earthaccess.auth.Auth object, which will then be used to create a session of 
+    s3login_credentials as they are called.
+    
+    Parameters
+    ----------
+    auth : earthaccess.auth.Auth, default None
+        Optional parameter to pass initialize an object with existing credentials
```
JessicaS11, line 30:
```suggestion
        Optional parameter to initialize an object with existing credentials.
```

icepyx/core/query.py
```diff
@@ -19,6 +18,7 @@
 # from icepyx.core.granules import Granules
 from icepyx.core.variables import Variables as Variables
 import icepyx.core.validate_inputs as val
+from icepyx.core.auth import EarthdataAuthMixin
```
JessicaS11, line 21: I know the list is currently not up to snuff on this, but can we put this in what would-be ABC order (so after APIformatting)?

icepyx/core/query.py
```diff
@@ -886,7 +889,7 @@ def show_custom_options(self, dictview=False):
             all(key in self._cust_options.keys() for key in keys)
         except AttributeError or KeyError:
             self._cust_options = is2ref._get_custom_options(
-                self._session, self.product, self._version
+                self.session, self.product, self._version
```
JessicaS11, line 892: would this now need to be `self.auth.session` (or `self.auth._session`)?

icepyx/core/auth.py
```diff
@@ -0,0 +1,152 @@
...
+    >>> a.session # doctest: +SKIP
+    >>> a.s3login_credentials # doctest: +SKIP
+    """
+    def __init__(self, auth=None):
+        self._auth = copy.deepcopy(auth)
+        # initializatin of session and s3 creds is not allowed because those are generated
+        # from the auth object
+        self._session = None
+        self._s3login_credentials = None
+        self._s3_initial_ts = None  # timer for 1h expiration on s3 credentials
```
JessicaS11, line 44: Trying to think through the arguments for and against having these as private variables. Are they public in earthaccess? If so, it might be good to follow that practice here.

icepyx/core/query.py
```diff
@@ -1102,7 +1054,7 @@ def order_granules(self, verbose=False, subset=True, email=False, **kwargs):
                     self.subsetparams(**kwargs),
                     verbose,
                     subset,
-                    session=self._session,
+                    session=self.session,
```
JessicaS11, line 1057:
```suggestion
                    session=self.auth.session,
```

icepyx/core/query.py
```diff
@@ -1113,7 +1065,7 @@ def order_granules(self, verbose=False, subset=True, email=False, **kwargs):
                 self.subsetparams(**kwargs),
                 verbose,
                 subset,
-                session=self._session,
+                session=self.session,
```
JessicaS11, line 1068:
```suggestion
                session=self.auth.session,
```

icepyx/core/query.py
```diff
@@ -1179,7 +1130,7 @@ def download_granules(
             ):
                 self.order_granules(verbose=verbose, subset=subset, **kwargs)
 
-        self._granules.download(verbose, path, session=self._session, restart=restart)
+        self._granules.download(verbose, path, session=self.session, restart=restart)
```
JessicaS11, line 1133:
```suggestion
        self._granules.download(verbose, path, session=self.auth.session, restart=restart)
```

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -20,16 +20,7 @@ def reg():
...
-    # append to netrc file and set permissions level
-    args = ("icepyx_devteam", "urs.earthdata.nasa.gov", os.getenv("NSIDC_LOGIN"))
-    netrc_file = os.path.join(os.path.expanduser("~"), ".netrc")
-    with open(netrc_file, "a+") as f:
-        f.write("machine {1} login {0} password {2}\n".format(*args))
-        os.chmod(netrc_file, 0o600)
-
-    reg.earthdata_login()
-    ed_obj = reg._session
+    ed_obj = reg.session
```
JessicaS11, line 23:
```suggestion
    ed_obj = reg.auth.session
```

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -51,7 +42,7 @@ def test_get_custom_options_output(session):
 # NOTE: best this test can do at the moment is a successful download with no errors...
 def test_download_granules_with_subsetting(reg, session):
     path = "./downloads_subset"
-    reg._session = session
+    reg.session = session
```
JessicaS11, line 45: do we still need to set this manually if login happens automatically now? Flagging this to double check.

JessicaS11:
> * how could I create one of those for the new class? Are they programmatically generated, or written out by hand?

Somewhere in between, unfortunately. I think the most appropriate spot for this is probably as one of the `components`, so we can just copy-paste a new block into that .rst file. If we now need to access auth info via `query_object.auth`, we should probably include a pointer function to get users from `query_object.earthdata_login` to `query_object.auth.earthdata_login`. In that function our docstring should note the deprecation of the function so users see that in the API docs.

icepyx/core/auth.py
```diff
@@ -0,0 +1,152 @@
...
+        # has passed since the last login    
+        if self._s3login_credentials is None:
+            set_s3_creds()
+        elif (datetime.datetime.now() - self._s3_initial_ts) >= datetime.timedelta(hours=1):
+            set_s3_creds()
+        return self._s3login_credentials
+
+    def earthdata_login(self, uid=None, email=None, s3token=False, **kwargs) -> None:
+        """
+        Authenticate with NASA Earthdata to enable data ordering and download.
```
JessicaS11, line 101: we should probably add a deprecation warning to the docstring too!

JessicaS11 (changes requested):
@rwegener2 This is beautiful! Thanks for all your hard work putting it together. I've left a bunch of comments (all minor things) and plan to check out this branch for local testing tomorrow. We're nearly there!

icepyx/tests/test_auth.py
```diff
@@ -0,0 +1,37 @@
+import pytest
+import requests
+
```
JessicaS11, line 3: should we combine this suite with the `test_Earthdata.py`? The earthdata tests are essentially checking the login methods, which are conceptually part of auth as well, though also not technically directly testing pieces of auth. Open to thoughts and suggestions on what makes the most sense.

doc/source/example_notebooks/IS2_data_access.ipynb
rwegener2: @JessicaS11 I was curious how you wanted to handle these two warnings. One is for the latest change and one is for the depreciated need for the "email" or "user" parameters. Do you think the parameters warning has been around long enough that we can remove it all together? Or should it be combined with the new warning?

icepyx/core/query.py
rwegener2: sure, happy to follow alphabetical as a standard!

icepyx/core/query.py
rwegener2: All instances of `self.auth._session` should now be `self.auth._session`, and similarly for `self.auth.s3login_credentials`. The private variable can be accessed, but doing so would bypass the automatic setting of the session or s3 credentials.

icepyx/core/auth.py
rwegener2: In the `__init__` method these variables need to be private, as this is the machinery for setting them as a read-only property for the user. We internally set the private versions (ex. `_session`) and then make them publicly available by using the `@property` decorator. The public accessors (ex. `.session`) are behind the scenes accessing the private one -- notice that each of the return statements for the properties returns the value of the private variable. As developers we change the value of the private versions, but the user only accesses the public versions.

icepyx/core/query.py
rwegener2: This isn't needed. Same reasoning as [previous comment](https://github.com/icesat2py/icepyx/pull/435#discussion_r1301642848).

rwegener2:
Okay @JessicaS11, I think this is ready for another check. I left comments next to yours, but to summarize:

* ❓ open question [here](https://github.com/icesat2py/icepyx/pull/435#discussion_r1301621620) about if you'd like me to remove or persist the earthdata_login warning about not needing an email or user
* ✅ Thanks for the explanation of how to add docstrings. I support your idea of adding the auth to Components, and it is integrated.
* 🗣️ Even inside the internal modules session objects should be accessed with the public `.session` syntax. This is how the user would in theory access them, if needed (although we hope they never actually need to). Using the public properties instead of the private ones is what enables the auto-checking to see if a session has already been created, and what creates a session if there isn't one already. Going forward I don't expect anything other than the EarthdataAuthMixin to access the private properties.
* 🗣️ On a similar note, the Mixin is added as an inherited class, not as an additional attribute, so the `self.session` syntax is sufficient. `self.auth.session` isn't needed. Let me know if either of the last two bullets don't make sense.
* ❓ Based on the last two bullet points I've been wondering about a good place to make a few dev notes about auth. Following the xarray model I added an "internals" page to the Contributing section, which right now just has a paragraph about authentication. I tried to make it long enough to be explanatory, but short enough not to be a pain to maintain. Let me know what you think of this structure.
* ✅ I believe all your comments about docstrings have been addressed. To this one specificially:

>  we should probably include a pointer function to get users from `query_object.earthdata_login` to `query_object.auth.earthdata_login`. In that function our docstring should note the deprecation of the function so users see that in the API docs.

I left the `Query.earthdata_login` method in the documentation, which has the note about being depreciated in it.

* ❓ I don't have a strong feeling about combining the tests vs. keeping them separate. I left them separate for now since that's my leaning, but will certainly combine if you prefer that.

icepyx/core/query.py
JessicaS11: Thanks for explaining that!

icepyx/tests/test_auth.py
JessicaS11: > I have a very slight lean towards keeping them separate,

A slight lean is enough argument for me!

icepyx/tests/test_behind_NSIDC_API_login.py
JessicaS11:
```suggestion
```

doc/source/contributing/icepyx_internals.rst
```diff
@@ -0,0 +1,70 @@
...
+    region_a.s3login_credentials
+
+
+**Adding authentication to a new class**
+
+To add authentication to an additional icepyx class, one needs to add the Mixin
+to the class. To do this:
+
+1. Add the EarthdataAuthMixin class to the ``class`` constructor (and import the mixin)
+2. Add the EarthdataAuthMixin init method with in the init method of the new class ``EarthdataAuthMixin.__init__(self)``
```
JessicaS11, line 39:
```suggestion
2. Add the EarthdataAuthMixin init method within the init method of the new class ``EarthdataAuthMixin.__init__(self)``
```

icepyx/core/auth.py
```diff
@@ -0,0 +1,155 @@
...
+
+        Generates the needed authentication sessions and tokens, including for cloud access.
+        Authentication is completed using the [earthaccess library](https://nsidc.github.io/earthaccess/).
+        Methods for authenticating are:
+            1. Storing credentials as environment variables ($EARTHDATA_LOGIN and $EARTHDATA_PASSWORD)
+            2. Entering credentials interactively
+            3. Storing credentials in a .netrc file (not recommended for security reasons)
+        More details on using these methods is available in the [earthaccess documentation](https://nsidc.github.io/earthaccess/tutorials/restricted-datasets/#auth).
```
JessicaS11, line 111: I'd like to move this (or something like it) into the mixin docstring - users will still need to know how they can authenticate. This way that guidance won't be lost when we (ideally) someday deprecate this function.

doc/source/example_notebooks/IS2_data_access.ipynb
JessicaS11: For now I'd like to leave both in there, since it's only in the last (0.7.0, March 2023) release that we deprecated the input parameters. And based on the Hackweek, a LOT of folks are still putting those in, so I'd like to encourage folks to update their code accordingly (or at least be less surprised in the future when the login function goes away entirely).

JessicaS11:
@rwegener2 I'm running the basic data access notebook locally and... it fails. I've got valid authentication (it did show an error before bringing up the manual login box, but it did let me authenticate that way and I got valid credentials), but cannot actually do anything that's behind a login. The two failure cases I found are: `region_a.order_vars.avail()` and `region_a.order_granules()`, both of which are getting hung up in the requests stage.

rwegener2:
> I'm running the basic data access notebook locally and... it fails.

Ok, I'm trying it out now.

I also get a failure, but mine is happening on 
```
#search for available granules and provide basic summary info about them
region_a.avail_granules()
```
due to `AssertionError: Your search returned no results; try different search parameters`.  Are you changing something about the default search parameters to locate more data? I just opened `IS2_data_access.ipynb` and did Restart and Run All Cells

JessicaS11:
> I wouldn't think that would be an error with this code, but if it's not expected I can dig into it.

There are a few examples in there that are set up to fail to demo things for the user. With Run All Cells you're likely hitting up against one of those.

icepyx/core/auth.py
```diff
@@ -0,0 +1,152 @@
...
+        
+        auth = earthaccess.login(**kwargs)
+        if auth.authenticated:
+            self._auth = auth
+            self._session = auth.get_session()
+
+        if s3token == True:
+            self._s3login_credentials = auth.get_s3_credentials(daac="NSIDC")
+
+        if uid != None or email != None:
```
JessicaS11, line 148:
```suggestion
        if uid != None or email != None or s3token != None:
```

icepyx/core/auth.py
```diff
@@ -0,0 +1,152 @@
...
+    The input parameters listed here are provided for backwards compatibility;
+    before earthaccess existed, icepyx handled authentication and required these inputs.
+
+    This class can be inherited by any other class that requires authentication. For 
+    example, the `Query` class inherits this one, and so a Query object has the 
+    `.session` property. The method `earthdata_login()` is included for backwards compatibility.
+    
+    The class can be created without any initialization parameters, and the properties will
+    be populated when they are called. It can alternately be initialized with an 
+    earthaccess.auth.Auth object, which will then be used to create a session of
```
JessicaS11, line 32:
```suggestion
    earthaccess.auth.Auth object, which will then be used to create a session or 
```

doc/source/example_notebooks/IS2_data_access.ipynb
JessicaS11, line 1: looks like the features typo got reintroduced... what do you think about removing the "if you choose" language, since we'd ultimately like to discourage deprecated uses of the function.

rwegener2: shoot, sorry about that features typo. I'll push that change now.

I see what you're saying about the "if you choose" language. What do you think of:
> If you are unable to remove  `earthdata_login()` calls from your workflow, you will also note that certain inputs, such as `earthdata_uid` and `email`, are no longer required. e.g. `region_a.earthdata_login(earthdata_uid, email)` becomes `region_a.earthdata_login()`
If you have different phrasing I'm happy with whatever you prefer.

JessicaS11: That looks great to me - it gently pushes folks towards removing it. If we want to mince words we could take out "you will also"... (science writing has gotten to me I guess). :)

rwegener2: I'm glad you've got such a tuned eye for writing! Definitely better to be concise. Both those changes of phrase are pushed

JessicaS11 (approved):
Thanks for all your hard work on this, @rwegener2!

Please note that the uml will auto-rebuild after I approve and add a last commit (which will then require all the tests to re-pass before the merge button is available). And I think it's forced, but a squash merge would be greatly appreciated!

rwegener2:
Yay, soooo exiting! Thanks for all of your attentive review @JessicaS11! Merging now! 🎊

PR #439, opened 2023-08-29
Add QUEST available datasets to readthedocs

zachghiaccio:


doc/source/index.rst
```diff
@@ -145,7 +145,7 @@ ICESat-2 datasets to enable scientific discovery.
    contributing/contributors_link
    contributing/contribution_guidelines
    contributing/how_to_contribute
-   contributing/icepyx_internals
+   contributing/quest-available-datasets
```
JessicaS11, line 148: we'll need to keep `contributing/icepyx_internals` as well.

JessicaS11:
Docs build is failing due to code changes that are not showing up in the diff (so are hard to track down). The base branch for merging this PR was changed, but the diff (three dot by default) isn't updating and showing the appropriate diff for this case (two dot). It's a [longstanding (7 year!) issue with changing PR bases in GitHub](https://github.com/isaacs/github/issues/750), and the "change the base and change it back again" hack didn't work here. Recommend we scrap this PR and start fresh.

PR #441, opened 2023-08-30
update QUEST and GenQuery classes for argo integration

JessicaS11:
With lots of ongoing work to integrate argo into quest (see #427), some updates were needed to the Quest and GenQuery classes. This PR should be merged before #427:
- moves spatial and temporal properties from Query into GenQuery so that they are accessible for all submodules (and not just ICESat-2 Query instances)
- clarifies that quest/dataset_scripts/dataset.py is a template class for adding new datasets
- sets up the QUEST class to add each dataset class currently/soon to be available and search and download data for them (the search+download functions have if statements to handle icesat2 versus other datasets because the initial icepyx.core.query class used different method names; the dataset template requires all future datasets added use the same methods)
- adds docs around what datasets are currently available using quest

TO DO:
- [x] the quest class needs docstrings
- [ ] may need to add some quest functions to the API docs


JessicaS11:
@zachghiaccio @kelseybisson please add your additional docs and docstrings work to this shared_search branch/PR, not the argo one!

RomiP (approved):

PR #451, opened 2023-10-03
Variables as an independent class

rwegener2:
## Goal
The ability to list the available variables and their h5 paths is a unique and powerful feature of icepyx. This PR aims to make that first-order functionality, instead of functionality that is only used internally. I think this will be a useful user feature, and I think it makes sense to make this change before adding the s3 url capability.

## How it was done

We anticipate that the most common use case is reading variables from a file. Alternately, all you need to produce an `.avail()` list of variables is a product and version. This information is used to ping an API, which returns the variables. 

When using `Variables()` the user is required to supply EITHER:
1. A local filepath (variables are read directly from the file)
4. the product

The version argument is optional. If it is not supplied the latest version is assumed. If both `path` and `product` is given an error is raised.

## How it can be tested (updated)
```python

# initializing with a filepath
v = Variables('data/ATL06/processed_ATL06_20190226005526_09100205_006_02.h5')
v.avail()

# initializing with a product only
v = Variables(product='ATL03')
v.avail()
print(v.version)  # returns '006'

# initializing with a product and version
v = Variables(product='ATL03', version='004')
v.avail()
print(v.version)  # returns '004'
```

And testing that Read and Query still work:
```python
from icepyx.core.read import Read
ds = Read('data/ATL06/processed_ATL06_20190226005526_09100205_006_02.h5')
ds.vars.append(beam_list=['gt1l'], var_list=['h_li', "latitude", "longitude"])
ds.load()
```

```python
from icepyx.core.query import Query
from icepyx.core.read import Read

# Create Query object
short_name = 'ATL06'
spatial_extent = [-55, 68, -48, 71]
date_range = ['2019-02-20','2019-02-28']
region = Query(short_name, spatial_extent, date_range)

# order and download granules
region.order_vars.append(beam_list=['gt1l'], var_list=['h_li', "latitude", "longitude"])
region.order_granules(Coverage=region.order_vars.wanted)
path = './download'
region.download_granules(path)

# open the files to see they have the required variables included (ex. sc_orient, atlas_sdp_gps_epoch)
reader = Read(path + '/processed_ATL06_20190222010344_08490205_006_02.h5')
reader.vars.append(beam_list=['gt1l'], var_list=['h_li', "latitude", "longitude"])
reader.load()
```

## Notes
* One function the Variables class was playing was also to enforce that the user requests certain variables. This is done both to help the user know what to want and also to help icepyx with future merges. I'm noting here that as this PR also represents a design change in which Variables no longer enforces that, Read and Query do. We implemented this by:
1. **Read**: The required `var_list` is appended to the wanted variables list (`vars.wanted`) during `load`. If `vars.wanted` does not exist an error is raised instructing the user to add variables before loading the dataset.
2. **Query**: We want the required variables to be appended only if the user has requested a subsetted list of variables. The user does this by passing the `Coverage` kwarg into `order_granules` (Commonly: `region_a.order_granules(Coverage=region_a.order_vars.wanted)`) Required variables will get appended to `Coverage` in the `subsetparams` method of `Query`.

## Open questions (Now resolved)
1. _Do we think it is important to maintain backwards compatibility given that this is a class that has, up until now, been for internal use only? --> We will raise DeprecationErrors for using old arguments, but won't fully support old functionality_
6. _What are the most user-friendly input arguments? (See final decision described above)_
7. _One aspect of reading that has now changed is that a user is allowed to read a data without appending any variables. This previously caused an error (`UnboundLocalError: local variable 'groups_list' referenced before assignment`). Now the code loads a dataset object that is basically empty. Do we want this? --> We will raise a new descriptive error if the user has not added any variables before reading_


rwegener2:
Icepyxstructuresdiagram
A visual guide I was using to think about the structures we have available to users

rwegener2:
Outcomes of a discussion with @JessicaS11:
> Do we think it is important to maintain backwards compatibility given that this is a class that has, up until now, been for internal use only?

* We won't maintain the existing features as fully backwards compatible, but anyone trying to use previous functionality will be given an `DeprecationError`

> What are the most user-friendly input arguments?

We will have:
1. a file (local or, in a future PR, s3). Check to see if there is an existing standard for what to call this, otherwise use `filepath`
2. product (and optionally version) — Default to the latest version if none is given (Check into the functionality for finding the latest version in Query class).
3. We actually don't want Query object as an input. `self.variables` is an attribute of a Query object, so implementing Query as an input arg would be circular.

We won't aggregate these into a single `data_source` argument. Instead make all arguments optional, but enforce that either a `filepath` or a `product` argument (but not both) is given.

> Re: The docs

Plan is to use the existing page for nested variables, but restructure the page so it starts talking generally about the Variables class, then transitions to showing how you can also use the class directly from a Query object.

rwegener2:
This PR is almost totally ready for review. I ran into one bug while running all the example notebooks that I'm looking for some help with.

In [line 479](https://github.com/icesat2py/icepyx/blob/a187328d3ff2640ca59071f8fdacc002dd8ac6e4/icepyx/core/variables.py#L479) of `variables.py` there is an if/else statement:
```python
    478 if self._vartype == "order":
    479     nec_varlist = [
    480         "sc_orient",
    481         "sc_orient_time",
   (...)
    488         "end_delta_time",
    489     ]
    490 elif self._vartype == "file":
```
In this new iteration of Variables there is no more `vartype`. If the vartype used to be order now `Variables` just takes in the product and version and generates a variables list from that.

So my question is: what was the purpose of using different sets of `nec_varlist`s based on the `vartype`?  I'm trying to figure out if this can be modified to run without knowing a `vartype`. cc: @JessicaS11 

### Context
If it is helpful, I found this while running  the [Building the default variables list](https://icepyx.readthedocs.io/en/latest/example_notebooks/IS2_data_access2-subsetting.html#building-the-default-wanted-variable-list) section of the Subsetting notebook. The line:
```python
region_a.order_vars.append(defaults=True)
pprint(region_a.order_vars.wanted)
```
Raised the error:
```
AttributeError                            Traceback (most recent call last)
Cell In[7], line 1
----> 1 region_a.order_vars.append(defaults=True)
      2 pprint(region_a.order_vars.wanted)

File ~/icepyx/icepyx/core/variables.py:478, in Variables.append(self, defaults, var_list, beam_list, keyword_list)
    475 self._check_valid_lists(vgrp, allpaths, var_list, beam_list, keyword_list)
    477 # add the mandatory variables to the data object


AttributeError: 'Variables' object has no attribute '_vartype'
```

doc/source/user_guide/documentation/icepyx.rst
```diff
@@ -23,4 +23,5 @@ Diagrams are updated automatically after a pull request (PR) is approved and bef
    query
    read
    quest
+   variables
```
JessicaS11, line 26: Note to remove variables content from `components.rst`

icepyx/core/variables.py
```diff
@@ -638,3 +657,19 @@ def remove(self, all=False, var_list=None, beam_list=None, keyword_list=None):
                             del self.wanted[vkey]
                     except KeyError:
                         pass
+
+    # DevNote: This is a modified function from the Query class. 
+    def _get_latest_version(self):
```
JessicaS11, line 662: Should we move a version of this function to is2ref instead of duplicating it here and in query? We could allow passing in `product` or `about_info` to accommodate the Variables and Query cases, respectively. Or is this a case where the function itself isn't long enough to warrant not duplicating it?

JessicaS11:
> So my question is: what was the purpose of using different sets of `nec_varlist`s based on the `vartype`? 

Fundamentally, the `Variables` instance from within `query` is of vartype "order", while the `Variables` instance from `read` is of vartype "file". For users using variable subsetting to order data, we wanted to make sure that they got back some minimal set of variables they'd need (or appreciate having) to work with the data (but wouldn't necessarily order on their own initially) without needing to place another data order.  This was likely more important when the data was just coming out and there was no a priori knowledge for most lab groups on what variables/groups/paths they might want. From the read side, there are a few pieces of information we need to be able to index/manipulate/merge across granules, so this allowed us to mandate they're included.

(As an aside: The non-implemented interactions of a `query.variables` instantiation with a type "file" confuses this, and if memory serves were the start of making it possible for the user to create a query object by supplying an ICESat-2 granule (i.e. file), I think with the goal that they could then still use the variables module to explore the variables in the file. It's probably worth my trying to find some more background on this idea and, I'm guessing, removing this non-implemented code from query to clean things up.)

> I'm trying to figure out if this can be modified to run without knowing a `vartype`

I think the short answer is "yes". From the `read` end, we can pretty easily add our own list with an `.append(vars_we_want)` call before we start loading. For the `query` (so on-prem subsetting) usage, the user directly constructs their `vars.wanted` and supplies it via the `coverage` kwarg to `query.order(coverage=[])` or `query.download(coverage=[])`, so we'd have to somehow make sure the defaults were included when a user does variable subsetting.

### Brainstorming Options
I suppose one alternative would be to take the minimal list required for the `read.load` application and use that for all cases (thus removing a few of the `vartype="order"` defaults). I think I would prefer the per-variables-object model described above though because I suspect that will get us closer to generating (for instance) Andy's hackweek list of variables and ultimately leaves things more flexible.

doc/source/user_guide/documentation/icepyx.rst
rwegener2: good catch!

icepyx/core/variables.py
rwegener2: I think it's a good idea to move it into is2ref. I was unsure at first, because the function currently sets the `_about_product` instance attribute when it gets run. If we move `latest_version` to is2ref we can't set that attribute anymore. Looking through query a second time, though, I think it's fine to move `latest_version` is2ref. `latest_version()` only gets called in init, so we don't have to worry about someone calling it multiple times. In theory we loose a little efficiency the next time someone calls `product_summary_info` or `product_all_info`, but my two cents is that speed is worth it to not have a function duplicated.

I just pushed a commit with the function moved, so feel free to comment further in review.

JessicaS11: > In theory we loose a little efficiency the next time someone calls product_summary_info or product_all_info, but my two cents is that speed is worth it to not have a function duplicated.

Agreed. If only we knew how frequently these functions are used...

rwegener2:
Alright, I've marked this are "Ready for Review", although it isn't 100% complete. I'm getting a bit turned around about how to remove the `nec_varlist` from Variables. I think I did it fine for the Read class, but I'm getting lost tracing the `Coverage` parameter for the Query class. Instead of surging ahead on Query I thought it would be worth it to request review for the current state, and then have a conversation to outline how to update the Query class.

icepyx/core/read.py
```diff
@@ -289,6 +289,9 @@ class Read:
     glob_kwargs : dict, default {}
         Additional arguments to be passed into the [glob.glob()](https://docs.python.org/3/library/glob.html#glob.glob)function
 
+    glob_kwargs : dict, default {}
+        Additional arguments to be passed into the [glob.glob()](https://docs.python.org/3/library/glob.html#glob.glob)function
+
```
JessicaS11, line 294:
```suggestion
```

merge issue?

JessicaS11: it might be the diff view. We had it happen in another PR that when we changed the base it mucked up the diff (.. vs ...; I'll try changing the base to see if we can get it to reset)

rwegener2:
Okay, this is ready for review! I've been updating the PR description so it is up-to-date with test examples and description.Two comments:

1. I don't know well how the gridded products are different from the non-gridded products. That's a blind spot I have so keep an eye out for assumptions I may have made.
2. In the way I was testing the Read functionality (example in the PR description) I was downloading a product with Query then reading it with Read. It seems that Read required `cycle_number` and `rgt`, but Query didn't require those. That felt odd, because I couldn't read a file with icepyx that I downloaded with icepyx, so I added those two variables to the Query required `var_list`. If there is a reason that isn't a good idea let me know and I can take them out.
3. I'm requesting a little extra focus on the Variables `append` method. I didn't understand 100% of the parts I was modifying, so just requesting an extra look at that part.

Looking forward to the review!

icepyx/core/query.py
```diff
@@ -646,6 +646,27 @@ def subsetparams(self, **kwargs):
...
+                    "data_end_utc",
+                    "granule_start_utc",
+                    "granule_end_utc",
+                    "start_delta_time",
+                    "end_delta_time",
+                ]
+                self.order_vars.append(defaults=False, var_list=var_list)
+                for var, path in self.order_vars.wanted.items():
+                    if var not in kwargs['Coverage'].keys():
+                        kwargs['Coverage'][var] = path
```
JessicaS11, line 668: I think this is redundant with [`self.order_vars.append()`](https://github.com/icesat2py/icepyx/blob/a184d9cd7bece9da303b3998d841bdcb641a5233/icepyx/core/variables.py#L500)

icepyx/core/query.py
```diff
@@ -688,17 +709,16 @@ def order_vars(self):
                 # DevGoal: check for active session here
                 if hasattr(self, "_cust_options"):
                     self._order_vars = Variables(
-                        self._source,
-                        auth = self.auth,
                         product=self.product,
+                        version = self._version,
+                        auth = self.auth,
                         avail=self._cust_options["variables"],
```
JessicaS11, line 715:
```suggestion
                        avail=self._cust_options["variables"],
                        auth = self.auth,
```
Do you know if there's a "best practice" around having the vars in the order they are in the function's init (even if they're kwargs)? Or is that just my OCD? ; )

icepyx/core/read.py
JessicaS11: Note: remove duplicate docstring entry

icepyx/core/read.py
```diff
@@ -456,11 +458,22 @@ def vars(self):
...
+        Return the list of files represented by this Read object.
+        """
+        return self._filelist
+
+    @property
+    def product(self):
+        """
+        Return the product associated with the Read object.
+        """
+        return self._product
```
JessicaS11, line 476: more duplicate code

icepyx/core/read.py
```diff
@@ -742,8 +754,33 @@ def load(self):
...
+            "data_start_utc",
+            "data_end_utc",
+        ]
+        
+        # Adjust the nec_varlist for individual products
+        if self.product == "ATL11":
+            var_list.remove("sc_orient")
+        
+        # Note: This fails if we are reading a file that doesn't have the var_list
+        # variables in it.
```
JessicaS11, line 779: Good catch. In looking through the code, it seems there are built-in fixes for any of these that are missing (and the user is warned they are happening). We could put it in a try-except if we make sure that `load()` really will work without any of them.

icepyx/core/variables.py
```diff
@@ -97,16 +119,14 @@ def avail(self, options=False, internal=False):
...
-        # if hasattr(self, '_avail'):
-        #         return self._avail
-        # else:
+
         if not hasattr(self, "_avail") or self._avail == None:
-            if self._vartype == "order":
+            if not hasattr(self, 'path'):
                 self._avail = is2ref._get_custom_options(
-                    self.session, self.product, self._version
+                    self.session, self.product, self.version
```
JessicaS11, line 126: This line made me notice how internally inconsistent we are with having (e.g.) `_version` vs `version` across modules (something you previously mentioned, @rwegener2). Since they're generally attributes the user shouldn't change, should we make them private and give them property accessors within the variables module?

rwegener2:
I realized I never documented the outcome of the last synchronous call, so I'll do that here.

This PR moves the functionality that requires certain variables out of Variables and into Query or Read. This is being handled by:
1. **Read**: The required `var_list` is appended to the wanted variables list (`vars.wanted`) during `load`. If `vars.wanted` does not exist an error is raised instructing the user to add variables before loading the dataset.
2. **Query**: We want the required variables to be appended only if the user has requested a subsetted list of variables. The user does this by passing the `Coverage` kwarg into `order_granules` (Commonly: `region_a.order_granules(Coverage=region_a.order_vars.wanted)`) Required variables will get appended to `Coverage` in the `subsetparams` method of `Query`.

icepyx/core/query.py
rwegener2: Good catch. I just removed the `append()` call. Just to make sure, though, I _think_ that the `append()` call was adding the required variables to the wanted variables list, meaning that it wouldn't impact the API call to NSIDC. The loop, on the other hand, was adding the required variables to the coverage variables, meaning it is the one that has an impact on the actual data the user will get. This is why I removed the `append()` instead of the loop. Does that sound correct?

icepyx/core/variables.py
```diff
@@ -446,53 +466,20 @@ def append(self, defaults=False, var_list=None, beam_list=None, keyword_list=Non
...
-        # Adjust the nec_varlist for individual products
-        if self.product == "ATL11":
-            nec_varlist.remove("sc_orient")
-
         try:
-            self._check_valid_lists(vgrp, allpaths, var_list=nec_varlist)
+            self._check_valid_lists(vgrp, allpaths)
         except ValueError:
             # Assume gridded product since user input lists were previously validated
             nec_varlist = []
```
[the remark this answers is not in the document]
rwegener2: That makes sense, I took this try/except out. 
A related thing I have been worried about is how this code will translate to a gridded data product. It seems from this try/except that we don't actually want to ever append required variables if the product is gridded. Is that correct? If that is the case then maybe we should be writing an if statement using the product into the Read and Query classes to skip required variable appending for gridded data products.

icepyx/core/read.py
rwegener2: So I would argue that this code isn't duplicate. Instead it is making a private attribute accessible as a read-only public attribute to the user. The user _could_ technically access and modify `_filelist` or `_product`, but the underscore warns them that they really shouldn't be doing that. Instead we have provided publicly available `filelist` and `product` which the user can see, but can't modify, since we don't want them changing that information.

icepyx/core/variables.py
rwegener2: I think that's a great idea!

icepyx/core/query.py
JessicaS11: Glad you spelled this out - I *was* mistakenly thinking that adding it to the wanted variables list would then get it to the API call. However, in this current iteration I'm not sure we'd be properly adding the `var_list` items because we never append them to `self.order_vars.wanted`. The two approaches coming to mind are (1) `self.order_vars.append(var_list)` (which will do a similar loop to update keys/values) and `kwargs['Coverage'] = self.order_vars.wanted` or (2) create a new vars object, append the `var_list` and run through this loop (though I'm not sure if we'd run into issues if the key is in there, but the value has changed). The only potential downside of the former is that it's changing the user's curated list, but I'm guessing that shouldn't be too big of a problem (plus it means they get an accurate list of variables ordered and it's a lot easier to implement).

icepyx/core/query.py
```diff
@@ -813,22 +832,6 @@ def product_all_info(self):
             self._about_product = is2ref.about_product(self._prod)
         pprint.pprint(self._about_product)
 
-    def latest_version(self):
```
JessicaS11, line 816: Running through the notebooks and we now have this error to address:

`AttributeError: 'Query' object has no attribute 'latest_version'`

rwegener2: I don't see anywhere that Query calls latest_version. Do you recall what line caused this error?

icepyx/core/variables.py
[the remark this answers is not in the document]
rwegener2: Outcome of this comment:
Yes, we want to skip adding additional variables for gridded data products. We will do that:
1. **Read**: TBD
2. **Query**: Probably in `subsetparams`, the same place we add the new vars to Coverage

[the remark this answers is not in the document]
JessicaS11: Not yet thought through idea: could we add some sort of `required=__` where __ = 'read' or 'query' (or something) as an argument to `vars.append()`? It would ultimately behave within append the same way `defaults` does, and there could be a function in `is2ref.py` that then gets the right default list for the required+product (analogous to `is2ref._default_varlists(product)`). It farms the work out of read or query directly but capitalizes on related infrastructure and potentially allows for more future flexibility. Worth exploring?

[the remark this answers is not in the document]
rwegener2: An update to this comment is documented in Issue #467. @JessicaS11 let me know what you think of the proposed path forward in that issue.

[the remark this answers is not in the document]
rwegener2: @JessicaS11 I like the idea that maybe these lists of variables might be stored elsewhere. If we were to pursue it I would be tempted to make required a boolean, where the user gets to pick if they want to include the required variables or not.
However it is setup, I think an argument makes the most sense if it turns out that there is a lot of variability in which products need what variables. From #467 it looks like the groups are 1) ATL14, ATL15, ATL23 which don't have any of the required variables 2) ATL11 which has all but two 3) all the rest, which have all those variables. My leaning is that this is a small enough number of groups that we just add an additional if statement to `Read.load`. What do you think?

icepyx/core/read.py
rwegener2: This should now be resolved with the required variables check by product.

icepyx/core/query.py
rwegener2: Thanks for thinking this through with me. I think I have something working, which I tested by requesting a subset file and checked to make sure the un-requested but required variables also got added. Look good?

icepyx/core/query.py
```diff
@@ -797,6 +817,14 @@ def product_summary_info(self):
         ]
         for key in summ_keys:
             print(key, ": ", self._about_product["feed"]["entry"][-1][key])
+            
+    def latest_version(self):
+        """
+        A reference function to is2ref.lates_version.
```
JessicaS11, line 823:
```suggestion
        A reference function to is2ref.latest_version.
```

icepyx/core/query.py
```diff
@@ -797,6 +817,14 @@ def product_summary_info(self):
         ]
         for key in summ_keys:
             print(key, ": ", self._about_product["feed"]["entry"][-1][key])
+            
+    def latest_version(self):
+        """
+        A reference function to is2ref.latest_version.
+        
+        Determine the most recent version available for the given product.
+        """
```
JessicaS11, line 826:
```suggestion
        Determine the most recent version available for the given product.
        
        Examples
        --------
        >>> reg_a = ipx.Query('ATL06',[-55, 68, -48, 71],['2019-02-20','2019-02-28'])
        >>> reg_a.latest_version()
        '006'
        """
```
Figured since we already had an example there we might as well use it!

JessicaS11 (approved):
Thanks for all the tremendous work on this, @rwegener2!

PR #452, opened 2023-10-06
enable QUEST kwarg handling

JessicaS11:
Improve QUEST module functionality by adding kwarg handling.

Follows #441, and includes a few related updates for that PR:
- additional docstring details
- add new docs page to TOC
- removes dev notes
- adds (for download_all) quest continuation after a dataset fails to download


JessicaS11:
@zachghiaccio want to add the docs TOC piece here (so it renders on RTD). Then we can add the links and example to the argo branch? Happy to hear other suggestions too.

zachghiaccio (approved):

PR #458, opened 2023-10-17
fix spot number calculation

JessicaS11:
Recent addressing of a dev note brought to light the fact that internal spot number calculations were incorrect. Further investigation showed that this was [previously addressed](https://github.com/ICESat2-SlideRule/sliderule/commit/d051948547fc5b2de348bf63204b9ef59314ae89) in SlideRule (the function upon which the original icepyx implementation was based) and is now clearly displayed in a table in the [ATL03 ATBD](https://nsidc.org/sites/default/files/documents/technical-reference/icesat2_atl03_atbd_r005.pdf) (p128), but the changes did not make it upstream into icepyx at the time the error was discovered. This PR updates the spot number calculation and adds a warning for users.

### DIFF
--- a/icepyx/core/is2ref.py
+++ b/icepyx/core/is2ref.py
@@ -265,8 +265,11 @@ def _default_varlists(product):
         return common_list
 
 
-# dev goal: check and test this function
 def gt2spot(gt, sc_orient):
+    warnings.warn(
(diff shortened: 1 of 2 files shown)


jpswinski (approved):
Changes look good.  I doubled checked and went through and matched them to the updated SlideRule code.

Issue #454, opened 2023-10-09
calls to OpenAltimetry API time out


JessicaS11:
Last week, OpenAltimetry transitioned their backend from its original domain to NSIDC and from grabbing data from a local server to the cloud-hosted copy of the data (ICESat-2 only so far). This has caused icepyx tests that utilize the OA programmatic API to start timing out and thus fail (see https://app.travis-ci.com/github/icesat2py/icepyx/builds/266433627 from #452).

JessicaS11:
After communicating with NSIDC on this, there are a few open tickets to deal with issues surrounding the migration of OpenAltimetry to NSIDC. Fixing the programmatic API and related functionality is one of them. Currently open and near-merging PRs in icepyx (e.g. #451, #452) are not time critical, so we will wait a few days to see if the issue is resolved and we can put through a patch to fix the failing tests (i.e. update the base_url for programmatic requests) before merging those.

mfisher87:
I believe this symptom has changed since the issue was opened from a timeout to a 404 (see #540).

mlampros:
@JessicaS11, @mfisher87 the API endpoints that are used in the Python code base (http://openaltimetry.earthdatacloud.nasa.gov/data/api/icesat2) are not the same I use for the IceSat2R package (https://openaltimetry.earthdatacloud.nasa.gov/data/openapi/swagger-ui/index.html). The latter url currently works except for level3a

mfisher87:
Nice, thank you! I was looking all over trying to find if there was a working API URL someplace, and just couldn't find anything! :bow:

JessicaS11:
Sorry @mfisher87 that I referred you here and this issue isn't fully updated. I was working with Mikala via email on this. The [url was the easy part once it stabilized](https://github.com/icesat2py/icepyx/compare/development...oa_api)(though at the time there wasn't any documentation, so I'm excited to see that - thanks @mlampros!). As noted in #495 the data format itself had changed, which would require rewriting things to parse (and plot) it. It shouldn't be too hard, but the bigger question was whether or not it was worth dev time and effort if nobody was using it...

mfisher87:
Right on, thanks for clarifying that, Jessica! That was exactly what I needed to know; updating #540 !

PR #459, opened 2023-10-18
temporarily disable OpenAltimetry API tests

JessicaS11:
As noted in #454, the OpenAltimetry programmatic API is currently unavailable during the transition of OA hosting from UCSD (Univ CA San Diego) to NSIDC (NASA DAAC). A ticket marked critical was issued, but the fix is not anticipated for another week or two. In order to allow other development to proceed, this PR disables tests depending on the OA API. The plan is to reinstate them as soon as the API becomes available again.


rwegener2 (approved):
Looks good! Great idea to put a warning in.

PR #468, opened 2023-11-07
Expand icepyx to read s3 data

rwegener2:
## Goal

NSIDC has put IS-2 data in the cloud, so we would like icepyx to be able to access that data for users. ☁️💙

## How it was done

The `EarthdataAuth` class was added as a mixin to the `Read` class. The product and version are read from the file the same way they are read in using a local file. The python standard glob module does not accept s3 paths, so s3 paths are not glob-able (i.e. the user must always give the full exact filepath). There is an s3 version of glob, so this could be implemented in the future. Mixed lists of s3 paths and local paths are not allowed. If a user tries to load more than two files or three data variables at the same time they are given a warning that discourages continuing and requires user confirmation to proceed.

Very loose timing suggests that it takes about 5 minutes per data variable per file to load s3 data (testing on the `dem_h` variable of an `ATL09` file). Running two files took 11 minutes.

## How to test it

```python
from icepyx.core.read import Read

s3_path = 's3://nsidc-cumulus-prod-protected/ATLAS/ATL09/006/2020/06/18/ATL09_20200618112352_12820701_006_01.h5'
reader = Read(s3_path)
reader.vars.append(var_list=['dem_h'])
reader.load()
```


rwegener2:
Results of the first timing run. I would like to do this testing a bit more systematically, but to give the first look from opening 1 variable `dem_h` of an ATL09 file (because I couldn't resist trying it out) we have:
| Description     |  creating the reader (No. 1) | loading some data (No. 2) |
| ----------- | ----------- | ---- |
| Local File     | 429 ms ± 6.8 ms    | 11.1 s ± 43.2 ms |
| s3 File   | 3.18 s ± 58 ms  | 5min 58s |

#### Reference Code

```python
from icepyx.core.read import Read

local_path = '/home/jovyan/data/ATL09/processed_ATL09_20200618112352_12820701_006_01.h5'
s3_path = 's3://nsidc-cumulus-prod-protected/ATLAS/ATL09/006/2020/06/18/ATL09_20200618112352_12820701_006_01.h5'

# `path` changes to either `local_path` or `s3_path`
reader = Read(local_path)  # This step was timed (No. 1)
reader.vars.append(var_list=['dem_h'])
reader.load()  # This step was timed (No. 2)
```

rwegener2:
Following up on the todo
> consider how we want to handle [it] if a list of s3 files or mixed s3/local paths are given as data_source

Allowing users to mix s3 and local paths sounds technically complex for what I imagine is a marginal use case, so I suggest we explicitly error if that happens. The options for what we could allow I see as:
1. Allow users to give lists of files so long as they are all s3 paths
2. Only allow users to give a single file if that file is on s3

I'm leaning towards the second one, just because it's barely practical for a user to open a variable from one file, let alone multiple. I could be swayed either way though. I'd love to hear additional opinions here.

JessicaS11:
> Following up on the todo
> 
> > consider how we want to handle [it] if a list of s3 files or mixed s3/local paths are given as data_source
> 
> Allowing users to mix s3 and local paths sounds technically complex for what I imagine is a marginal use case, so I suggest we explicitly error if that happens. 

Agreed. The error could suggest creating multiple read objects and merging the datasets after they're loaded in that use case.

> The options for what we could allow I see as:
> 
>     1. Allow users to give lists of files so long as they are all s3 paths
> 
>     2. Only allow users to give a single file if that file is on s3
> 
> 
> I'm leaning towards the second one, just because it's barely practical for a user to open a variable from one file, let alone multiple. I could be swayed either way though. I'd love to hear additional opinions here.

@scottyhq gave [this tutorial](https://github.com/CryoInTheCloud/CryoCloudWebsite/blob/main/book/tutorials/SlideRule_applications/SlideRule_applications.ipynb) during the AGU CryoCloud workshop yesterday. At the very end there's an example of directly reading in some ATL06 data using h5coro. Not sure if there's currently a way to leverage this implementation without the xarray extension piece, but it's worth thinking about how this space might evolve (there's another effort working on creating kerchunk files for IS2 data) and what the implications are for how data is read-in behind the scenes. I'm thinking that users should be allowed to submit a list of s3 urls (option 1), but if there's more than say 2 files and/or more than 3 variables requested they get a "it will take forever" warning (at a minimum) and we could consider requiring them to acknowledge it (e.g. require user input to proceed). For some reason I'm not feeling great about enforcing a one s3url limit, even if that's clearly the "right" thing to do to read in multiple files in a reasonable amount of time.

rwegener2:
> I'm thinking that users should be allowed to submit a list of s3 urls (option 1), but if there's more than say 2 files and/or more than 3 variables requested they get a "it will take forever" warning (at a minimum) and we could consider requiring them to acknowledge it (e.g. require user input to proceed)

This makes sense to me as a path forward, @JessicaS11. 

I looked at the tutorial of @scottyhq's you linked and it looks like it was accessing h5coro through sliderule. So to use this method we would have to use sliderule in icepyx and that sounds a bit like a spaghetti of data opening softwares. 

It's good to know about all the other efforts going on. I am doubtful that icepyx will open data quickly until there a change of file format/structure/kerchunk reference/similar effort. As for h5coro @jpswinski has made a lot of helpful updates in the past few months and we are meeting tomorrow to talk about moving the xarray backend forward. 🤞🏻 that can be an option soon.

icepyx/core/read.py
```diff
@@ -367,20 +386,53 @@ def __init__(
...
         product_dict = {}
-        for file_ in self._filelist:
-            product_dict[file_] = is2ref.extract_product(file_)
-
+        self.is_s3 = [False]*len(self._filelist)
+        for i, file_ in enumerate(self._filelist):
+            # If the path is an s3 path set the respective element of self.is_s3 to True
+            if file_.startswith('s3'):
+                self.is_s3[i] = True
+            product_dict[file_] = is2ref.extract_product(file_, auth=self.auth)
```
JessicaS11, line 418: Haven't tested it yet, but are we going to run into the same auth issue (see [this fix](https://github.com/icesat2py/icepyx/pull/464/commits/454e30b2ecfe33c4ea0a3d240481db59fd0852ce)) we did with the variables module? An easy fix would be to indent this line and add an else with `auth=None` instead of `auth=self.auth`.

icepyx/core/read.py
```diff
@@ -457,6 +509,20 @@ def vars(self):
...
+        Return the list of files represented by this Read object.
+        """
+        return self._filelist
+
+    @property
+    def product(self):
+        """
+        Return the product associated with the Read object.
+        """
+        return self._product
```
JessicaS11, line 525:
```suggestion
```
Duplicate code.

icepyx/core/read.py
rwegener2: Good catch! Just pushed a fix along these lines

rwegener2:
Alright this PR is ready for re-review. Changes since the last review:

- addressing [this catch](https://github.com/icesat2py/icepyx/pull/468#discussion_r1425736122) from @JessicaS11 about asking for authentication even for local reads
- fix a small error about parsing bytes for reading ATL03
- update the cloud tutorial notebook

icepyx/core/read.py
```diff
@@ -752,9 +817,16 @@ def load(self):
         # In these situations, xarray recommends manually controlling the merge/concat process yourself.
         # While unlikely to be a broad issue, I've heard of multiple matching timestamps causing issues for combining multiple IS2 datasets.
         for file in self.filelist:
+            if file.startswith("s3"):
+                # If path is an s3 path create an s3fs filesystem to reference the file
+                # TODO would it be better to be able to generate an s3fs session from the Mixin?
+                s3 = earthaccess.get_s3fs_session(daac="NSIDC", provider=self.auth)
```
JessicaS11, line 823: Good question. I searched and currently this is the second module (other than is2ref.py) that we use `get_s3fs_session()`. With three uses I'm fine to leave it here for now, but I think if we add another in the future the answer to your question would be a resounding "yes".

doc/source/example_notebooks/IS2_cloud_data_access.ipynb
```diff
@@ -12,8 +12,16 @@
     "## Notes\n",
```
JessicaS11, line 12: <pre class="ql-syntax" spellcheck="false">Line #1.    ipx.Variables(path=gran_ids[1][0]).avail()
</pre><p>This cell errors if you've run the above two cells (not sure if it does on it's own).</p><pre class="ql-syntax" spellcheck="false">---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [8], line 1
----&gt; 1 ipx.Variables(path=gran_ids[1][0]).avail()

File ~/icepyx/icepyx/core/variables.py:95, in Variables.__init__(self, vartype, path, product, version, avail, wanted, auth)
     93         auth = None
     94     # Read the product and version from the file
---&gt; 95     self._product = is2ref.extract_product(self._path, auth=auth)
     96     self._version = is2ref.extract_version(self._path, auth=auth)
     97 elif product:
     98     # Check for valid product string

File ~/icepyx/icepyx/core/is2ref.py:365, in extract_product(filepath, auth)
    361         raise AttributeError(
    362             "Must provide credentials to auth if accessing s3 data"
    363         )
    364     # Read the s3 file
--&gt; 365     s3 = earthaccess.get_s3fs_session(daac="NSIDC", provider=auth)
    366     f = h5py.File(s3.open(filepath, "rb"))
    367 else:
    368     # Otherwise assume a local filepath. Read with h5py.

File /srv/conda/envs/notebook/lib/python3.10/site-packages/earthaccess/api.py:337, in get_s3fs_session(daac, provider, results)
    326 """Returns a fsspec s3fs file session for direct access when we are in us-west-2
    327 
    328 Parameters:
   (...)
    334     class s3fs.S3FileSystem: an authenticated s3fs session valid for 1 hour
    335 """
    336 daac = _normalize_location(daac)
--&gt; 337 provider = _normalize_location(provider)
    338 if results is not None:
    339     endpoint = results[0].get_s3_credentials_endpoint()

File /srv/conda/envs/notebook/lib/python3.10/site-packages/earthaccess/api.py:27, in _normalize_location(location)
     17 """Handle user-provided daac and provider values
     18 
     19 These values must have a capital letter as the first character
   (...)
     24 https://wiki.earthdata.nasa.gov/display/ED/CMR+Data+Partner+User+Guide?src=contextnavpagetreemode
     25 """
     26 if location is not None:
---&gt; 27     location = location.upper()
     28 return location

AttributeError: 'Auth' object has no attribute 'upper'
</pre><p>I'm also getting the same auth error when I try to run read later on in the notebook...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/icesat2py/icepyx/pull/468/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

rwegener2: @JessicaS11 When I "Run All" cells in this notebook everything goes smoothly. Can you perhaps describe with some more detail how you got the error? Are you authenticating with a .netrc, manual entry, or env variables?

JessicaS11: @rwegener2 I'm authenticating manually and was running the notebook in CryoCloud. Unfortunately, in trying to reproduce again just now I've instead spent the last 30 minutes trying to get the notebook to install a dev version of icepyx witihin CryoCloud (again), to no avail. However, I did try it locally for fun and was surprised I got a permissions error, which I wasn't expecting since for this line it should only need my normal earthdata login, right?

JessicaS11: What version of earthdata are you running? I'll look into that as a potential error source if I ever get CryoCloud install/import the right version of icepyx.

rwegener2: Thanks for digging into this @JessicaS11. I have earthaccess 0.5.2 installed in my dev env. Sounds like what needs to happen is a change in syntax, which seems doable.

icepyx/core/read.py
```diff
@@ -752,9 +817,16 @@ def load(self):
...
+            if file.startswith("s3"):
+                # If path is an s3 path create an s3fs filesystem to reference the file
+                # TODO would it be better to be able to generate an s3fs session from the Mixin?
+                s3 = earthaccess.get_s3fs_session(daac="NSIDC")
+                file = s3.open(file, "rb")
+
             all_dss.append(
                 self._build_single_file_dataset(file, groups_list)
             )  # wanted_groups, vgrp.keys()))
+            file.close()
```
JessicaS11, line 829: this line causes an error when `file` is a string pointing to a local file (`AttributeError: 'str' object has no attribute 'close'`).

rwegener2: Good catch. Just pushed a fix.

icepyx/core/read.py
```diff
@@ -367,19 +392,56 @@ def __init__(
...
+        self.is_s3 = self.is_s3[0]  # Change is_s3 into one boolean value for _filelist
+        # Raise warning if more than 2 s3 files are given
+        if self.is_s3 is True and len(self._filelist) > 2:
+            warnings.warn(
+                "Processing more than two s3 files can take a prohibitively long time. "
+                "Approximate access time (using `.load()`) can exceed 6 minutes per data "
+                "variable.",
+                stacklevel=2,
+            )
+            _confirm_proceed()
```
JessicaS11, line 444: @rwegener2 I wanted to trigger the `_confirm_proceed()` so I passed in a list of s3 urls... I got the confirmation, but the warning never showed (and the confirmation is confusing without it). Wondering if it makes sense to kick the warning into the function?

rwegener2: Hm, interesting. The warning appeared when I was running it. That was a few commits back, though, so 🤷‍♀️ 
I think it's fine to add the warning to the `_confirm_proceed()` function. My guess is that either way we'll need to increase the `stacklevel` to 3 or 4 if the warning isn't showing up at level 2.

JessicaS11: I seem to recall I was messing with the warnings settings in my hub for a presentation recently... since this is an important one, I added a cell to turn warnings on (changing the stack level didn't help in my case). Let me know if you think we should take a different approach. Otherwise, this should be good to go.

Issue #466, opened 2023-11-01
ValueError: cannot reindex or align along dimension 'delta_time' because of conflicting dimension sizes: {1112, 1098}


eliskasieglova:
Hi!

I am having trouble reading ATL08 data. I followed the tutorial both for downloading and reading the data. This is my code:

```
short_name = 'ATL08'
spatial_extent = [15, 77, 18, 78]
date_range = ['2020-01-01', '2023-01-01']

# accessing data
region_a = ipx.Query(short_name, spatial_extent, date_range, start_time='00:00:00', end_time='23:59:59')
region_a.avail_granules()
region_a.earthdata_login()
region_a.order_granules(subset=False)
region_a.download_granules('cache/is2_ATL08')

# reading
pattern = "processed_ATL{product:2}_{datetime:%Y%m%d%H%M%S}_{rgt:4}{cycle:2}{orbitsegment:2}_{version:3}_{revision:2}.h5"
reader = ipx.Read(data_source=str(input_path), product=data_product, filename_pattern=pattern)
reader.vars.append(var_list=["h_te_best_fit", "latitude", "longitude"])
ds = reader.load()
```

This is the error message I get:

```
C:\Users\eliss\anaconda3\envs\SvalbardSurges\Lib\site-packages\icepyx\core\read.py:49: UserWarning: Converting non-nanosecond precision datetime values to nanosecond precision. This behavior can eventually be relaxed in xarray, as it is an artifact from pandas which is now beginning to support non-nanosecond precision values. This warning is caused by passing non-nanosecond np.datetime64 or np.timedelta64 values to the DataArray or Variable constructor; it can be silenced by converting the values to nanosecond precision ahead of time.
  df.update({keyword: df[keyword].str[:-1].astype(np.datetime64)})
C:\Users\eliss\anaconda3\envs\SvalbardSurges\Lib\site-packages\icepyx\core\read.py:49: UserWarning: Converting non-nanosecond precision datetime values to nanosecond precision. This behavior can eventually be relaxed in xarray, as it is an artifact from pandas which is now beginning to support non-nanosecond precision values. This warning is caused by passing non-nanosecond np.datetime64 or np.timedelta64 values to the DataArray or Variable constructor; it can be silenced by converting the values to nanosecond precision ahead of time.
  df.update({keyword: df[keyword].str[:-1].astype(np.datetime64)})
[...]

Issue #376, opened 2022-09-29
read-in: the index error on ATL08


liuh886:
I got a index error on ATL08 similar to these cases, but I am using `Icepyx version 0.6.3`:
- https://github.com/icesat2py/icepyx/issues/277
- https://discourse.pangeo.io/t/error-while-trying-to-open-atl08-file-with-icepyx/2313

Here is a minimal reproduction. I conduct it on two PC with the same result:

```
>>> region_a = ipx.Query('ATL08', any_spatial_extent, any_date_range)
>>> region_a.order_vars.wanted
>>> region_a.order_vars.append(keyword_list=['land', 'land_segments','terrain'])
>>> region_a.order_granules(Coverage=region_a.order_vars.wanted)
>>> region_a.download_granules(path)
>>> pattern = "processed_ATL{product:2}_{datetime:%Y%m%d%H%M%S}_{rgt:4}{cycle:2}{orbitsegment:2}_{version:3}_{revision:2}.h5"
>>> reader = ipx.Read(data_source=path, product="ATL08", filename_pattern=pattern)
You have 1 files matching the filename pattern to be read in.

>>> reader.vars.wanted
>>> reader.vars.append(var_list=['h_te_best_fit'])
>>> ds = reader.load()

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Input In [191], in <cell line: 1>()
----> 1 ds = reader.load()

File ~\.conda\envs\snowdepth\lib\site-packages\icepyx\core\read.py:542, in Read.load(self)
    535 # DevNote: I'd originally hoped to rely on intake-xarray in order to not have to iterate through the files myself,
    536 # by providing a generalized url/source in building the catalog.
    537 # However, this led to errors when I tried to combine two identical datasets because the single dimension was equal.
    538 # In these situations, xarray recommends manually controlling the merge/concat process yourself.
    539 # While unlikely to be a broad issue, I've heard of multiple matching timestamps causing issues for combining multiple IS2 datasets.
    540 for file in self._filelist:
    541     all_dss.append(
--> 542         self._build_single_file_dataset(file, groups_list)
    543     )  # wanted_groups, vgrp.keys()))
    545 if len(all_dss) == 1:
    546     return all_dss[0]

File ~\.conda\envs\snowdepth\lib\site-packages\icepyx\core\read.py:682, in Read._build_single_file_dataset(self, file, groups_list)
    680 grp_path = wanted_groups_list[0]
[...]

PR #470, opened 2023-11-13
fix ATL08 delta_time dimension read error

JessicaS11:
Per #376, #466, and #469 ATL08 data could not be read in.

During a read in of one of the deeply nested group variables, delta_time was being turned into a dimension (rather than staying as coordinates), causing merge conflicts when added to portions of the dataset that only had delta_time as coordinates due to the conflicting sizes of the delta_time coordinate. By dropping the delta_time dimension, we are able to merge properly.

This PR is being submitted as a patch bug fix directly to main in order to create a patch 0.8.1 release for an upcoming workshop.

### DIFF
--- a/icepyx/core/read.py
+++ b/icepyx/core/read.py
@@ -616,11 +616,12 @@ class Read:
         except (AttributeError, KeyError):
             pass
 
-        try:
-            is2ds = is2ds.assign(ds[grp_spec_vars])
-        except xr.MergeError:
-            ds = ds[grp_spec_vars].reset_coords()
-            is2ds = is2ds.assign(ds)
+        ds = ds[grp_spec_vars].reset_coords()
+        is2ds = is2ds.assign(ds)
+
+        # manually remove delta time as a dimension
+        if "delta_time" in is2ds.dims:
+            is2ds = is2ds.drop_dims("delta_time")
 
         return is2ds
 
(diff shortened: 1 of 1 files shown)


rwegener2:
Thanks for working on a fix @JessicaS11!

### 🎉
I tried this out and it runs smoothly for a single ATL08 file, multiple ATL08 files, and multiple ATL06 files. I'm getting an error trying to read ATL03 files, but I think that one may be a me problem because I get the same error using version 0.8.0.

### 😕
Looking at the output the ATL08 produces I am not seeing what I expected, but I also may just be misunderstanding. The output `ds` from the code in #469 (used `reader.vars.append(beam_list=['gt1l', 'gt3r'], var_list=['h_canopy', "latitude", "longitude"])`) was:

Screenshot 2023-11-14 at 9 06 06 AM
My confusions:

1. Shouldn't `h_canopy` be a Data Variable?
2. Shouldn't only `gt1l` and `gt3r` be listed in `gt`?

If you want to brainstorm together @JessicaS11 just let me know.

rwegener2:
As for the erroring build it seems like the error was a timeout on the visualization tests. Maybe it's an API issue? It's surprising that the API itself doesn't provide a timeout response. icepyx could write their own timeouts into it's code, so we get a recognizable error when the API called in the visualization code is non-responsive. But that, I think, would be a task for another day.

JessicaS11:
> My confusions:
> 
>     1. Shouldn't `h_canopy` be a Data Variable?
> 
>     2. Shouldn't only `gt1l` and `gt3r` be listed in `gt`?

Confusing indeed... I will see if I can make heads or tail of this.

rwegener2 (approved):

PR #472, opened 2023-11-14
add minor/patch release docs

JessicaS11:
release documents for v0.8.1 patch release once #470 is merged.

### DIFF
--- a/doc/source/user_guide/changelog/index.rst
+++ b/doc/source/user_guide/changelog/index.rst
@@ -6,8 +6,16 @@ icepyx ChangeLog
 This is the list of changes made to icepyx in between each release.
 Full details can be found in the `commit logs <https://github.com/icesat2py/icepyx/commits>`_.
 
-Latest Release (Version 0.8.0)
+Latest Release (Version 0.8.1)
 ------------------------------
+
+.. toctree::
+   :maxdepth: 2
+
+   v0.8.1
+   
+Version 0.8.0
+-------------
 
 .. toctree::
    :maxdepth: 2
--- a/doc/source/user_guide/changelog/v0.8.1.rst
+++ b/doc/source/user_guide/changelog/v0.8.1.rst
@@ -9,7 +9,7 @@ including other versions of icepyx.
 Bug fixes
 ~~~~~~~~~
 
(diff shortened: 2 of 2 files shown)


JessicaS11:
Not sure why the readthedocs builds are hanging. I just triggered one manually for this branch; #470 (just merged) built successfully and never had the status update properly in the PR yesterday/today...

rwegener2:
> Not sure why the readthedocs builds are hanging.

That is odd. I found [an instance](https://github.com/readthedocs/readthedocs.org/issues/2255) in the readthedocs GH issue history back in 2016 where the servers froze and someone had to go physically inspect to find the problem. I'm not sure if that could be happening here, too.

I have been able to keep working with a local install for the day. Maybe we leave it for a while and if it still isn't working in the morning submit an issue to readthedocs? Or we could just submit an issue now. 🤷🏻‍♀️

JessicaS11:
> Maybe we leave it for a while and if it still isn't working in the morning submit an issue to readthedocs? 

Ordinarily I'd be fine with that plan, but if we want the conda build to be picked up I'd prefer we have the release done tonight since [it can take a few hours to get picked up by their bot](https://conda-forge.org/docs/maintainer/updating_pkgs.html). I'd have concerns that even if you're showing a pip install someone will try to conda install and then you'll have version issues during the tutorial.

JessicaS11:
GitHub is getting a 400 from the readthedocs webhook with message `{"detail":"Parameter \"ref\" is required"}`. `ref` seems to be in some of the payloads submitted and not others (depending on the trigger type), but was not previously a problem (across trigger types). So right now `pull_request.push` triggers seem to be accepted by RTD, but other actions (e.g. `pull_request.synchronize`) are rejected.

weiji14:
Seems like an issue with the GitHub webhook integration - https://github.com/readthedocs/readthedocs.org/issues/10902? Looks like a fix is on the way.

rwegener2:
Good find @weiji14! It appears the issue was just resolved. I just re-pushed to trigger a build and the docs seem to have built properly. 🎉

rwegener2 (approved):

PR #473, opened 2023-11-20
Update read module coordinate dimension manipulations to use new xarray index

JessicaS11:
Since the data read-in module was built, xarray has added explicit indexing, which was causing a warning (`read.py:599: UserWarning: rename 'delta_time' to 'photon_idx' does not create an index anymore. Try using swap_dims instead or use set_index after rename to create an indexed coordinate.`) to appear during the `reader.load()` step when coordinate dimensions were manipulated internally. This PR updates that portion of the code to use `.swap_dims`, greatly simplifying it and removing the warning.


weiji14 (approved):
Nice, great to see that the code can be simplified (#222 was a mess)! I haven't tested this locally, but the unit tests are passing so should be fine :crossed_fingers:

PR #475, opened 2023-12-12
update github actions and add black linter for PRs

JessicaS11:
This PR aims to update and improve GH actions:
- update pypi action to use OIDC trusted publisher mgmt
- generalize the flake8 action to a general linting action and add black
- put flake8 config parameters into a separate file (`.flake8`)
- update versions of actions/pre-commit hooks
- specify uml updates only need to run on PRs to development (#449)
- update docs config files to be compliant

PR #479, opened 2023-12-15
remove `file` input arg and `_source` property from query (and improve some formatting)

JessicaS11:
This PR removes the unused `file` input kwarg to Query and takes out the associated `_source` attribute and any cases where it was called (usually in if statements without else cases).

To work towards PEP compliance as indicated by flake8 (see #475), this PR also shortens some lines and cleans up other aspects of query.py.


rwegener2 (approved):
Looks good to me! Getting rid of unneeded code is exciting!

Issue #448, opened 2023-09-14
debug tests behind EDL - failing builds for PRs to main


rwegener2:
PR #447 is an example of a build that ran well with merges to `development` or `release`, but the build fails due to an authentication error when merging to `main`. This was an issue for the release of both `0.7.0` and `0.8.0`, after the integration of `earthaccess`.

The main branch is failing because:
> because the behind_NSIDC_login test suite is only run on the main branch (so in the list of checks, the "branch" test run passes, but when the additional tests are triggered for the "pull request" (what the merged code will look like) it fails).
(From @JessicaS11 in #447)

JessicaS11:
Per the `.travis.yml` setup, including the string `nsidc_tests` in the commit message should trigger the `test_behind_earthdata_login` tests.

PR #480, opened 2023-12-18
Fix behind EDL tests

JessicaS11:
Per #448, Earthdatalogin tests are failing since we upgraded to earthaccess for auth. This PR is to fix those tests so they run.


JessicaS11:
@all-contributors please add @lheagy for mentoring, review

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -23,6 +24,7 @@ def reg():
 
 @pytest.fixture(scope="module")
 def session(reg):
+    os.environ = {"EARTHDATA_USERNAME": "icepyx_devteam"}
```
betolink, line 27: This is probably beyond the PR but have you all considered having a "default" username in case users want to use icepyx but don't have an EDL login?  (don't quote me =) )

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -32,8 +34,8 @@ def session(reg):
 
 
 def test_get_custom_options_output(session):
-    obs = is2ref._get_custom_options(session, "ATL06", "005")
-    with open("./icepyx/tests/ATL06v05_options.json") as exp_json:
+    obs = is2ref._get_custom_options(session, "ATL06", "006")
+    with open("./icepyx/tests/ATL06v06_options.json") as exp_json:
```
betolink, line 38: What would be the behavior if a script points to V005?

betolink (approved):
Looks good to me, the only observation is to what users will see if they have hard-coded queries to V005.

icepyx/tests/test_behind_NSIDC_API_login.py
JessicaS11: As long as it's still available from NSIDC, the version specified will be ordered. If not, whatever error comes from NSIDC in the request will be surfaced. The default is to the most recent version, so we hard code the version into the test so that when a new version is released any changes in the options don't cause an error before we can manually update the json.

icepyx/tests/test_behind_NSIDC_API_login.py
JessicaS11: We have, and we were told in no uncertain terms not to do it (it would also be very hard to do without somehow exposing the password). Since it's an ubiquitous problem, as you are well aware, we opted to just have users get an EDL (since they would have to just about anywhere except OA and SlideRule, I think), and we store the secrets for a dev account for CI testing (which isn't accessible for PRs from forks either).

PR #482, opened 2023-12-20
Convert deprecation warnings to errors

JessicaS11:
icepyx has a series of deprecations that are still warnings. Some are new deprecations during the modifications for cloud reading, and some are stale but were never turned into errors prior to planned removal. This PR aims to turn all current deprecations into errors for the v1 release, with planned removal of the errors in a future minor release.

### DIFF
--- a/icepyx/core/read.py
+++ b/icepyx/core/read.py
@@ -3,7 +3,6 @@ import glob
 import os
 import warnings
 
-import h5py
 import numpy as np
 import xarray as xr
 
@@ -157,9 +156,9 @@ def _validate_source(source):
     # acceptable inputs (for now) are a single file or directory
     # would ultimately like to make a Path (from pathlib import Path; isinstance(source, Path)) an option
     # see https://github.com/OSOceanAcoustics/echopype/blob/ab5128fb8580f135d875580f0469e5fba3193b84/echopype/utils/io.py#L82
-    assert type(source) == str, "You must enter your input as a string."
+    assert type(source) is str, "You must enter your input as a string."
     assert (
-        os.path.isdir(source) == True or os.path.isfile(source) == True
+        os.path.isdir(source) is True or os.path.isfile(source) is True
     ), "Your data source string is not a valid data source."
     return True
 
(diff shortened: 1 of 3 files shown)


JessicaS11:
@rwegener2 This is basically ready for review, but I figured we'd want to merge #468 first since that's got some code changes in the same areas and there could be merge conflicts.

JessicaS11:
@rwegener2 This is all set for review!

icepyx/core/read.py
```diff
@@ -344,54 +343,37 @@ class Read(EarthdataAuthMixin):
...
-        out_obj_type=None,  # xr.Dataset,
     ):
-        # Raise error for deprecated argument
-        if catalog:
-            raise DeprecationError(
-                "The `catalog` argument has been deprecated and intake is no longer supported. "
-                "Please use the `data_source` argument to specify your dataset instead."
-            )
-
         if data_source is None:
```
rwegener2, line 354:
```suggestion
```
Now that `data_source=None` Python will enforce this check

icepyx/core/read.py
```diff
@@ -344,54 +343,37 @@ class Read(EarthdataAuthMixin):
...
-            raise ValueError("data_source is a required arguemnt")
+            raise ValueError("data_source is a required argument")
```
rwegener2, line 355:
```suggestion
```
Now that `data_source=None` Python will enforce this check

rwegener2 (approved):
Looks good to me! I'm loving all that code in the Read init that can be eliminated. I had a few suggestions, but I had to commit them directly since Github doesn't let you suggest on un-modified code. I tried to make small descriptive commits so you can see what is changing but overall it was just small updates to deprecated docstrings and condensing the deprecation errors. 
Nice PR! 🎉

icepyx/core/read.py
JessicaS11: Good catch!

JessicaS11:
> I'm loving all that code in the Read init that can be eliminated.

Me too! Thanks for the edits.

Issue #485, opened 2023-12-22
figure out where the codecov reports went


JessicaS11:
Just noticed today that the codecov reports aren't being added to PRs, and the last PR that has one in codecov is from 9 months ago! Will need to dig in and update the integration to get the reports added back.

weiji14:
Codecov announced in Sep 2021 that their old bash uploader is being deprecated, could that be related? https://about.codecov.io/blog/codecov-uploader-deprecation-plan

JessicaS11:
Thanks for pointing me towards that, @weiji14!

PR #496, opened 2024-01-18
update codecov uploader to start getting coverage reports

JessicaS11:
At some point we stopped getting code testing coverage reports (it was failing silently in Travis CI). This PR updates to use the most recent uploader. Addresses #485.

References:
- https://docs.codecov.com/docs/codecov-uploader#integrity-checking-the-uploader


weiji14 (approved):
Sorry for the late reply, was away for a few days. I see that the code is copied from the official instructions at https://docs.codecov.com/docs/codecov-uploader#integrity-checking-the-uploader, and the Codecov bot commented above, so should be all good!

PR #500, opened 2024-01-26
pkg_resources deprecation: stop calling setup.py directly

JessicaS11:
[pkg_resources](https://setuptools.pypa.io/en/latest/pkg_resources.html) was deprecated as an API several years ago, resulting in numerous warnings. Clear guidance on what, exactly, needs to change is difficult to find if you're not intimately familiar with the build process (even considering the [deprecation warning-cited blog post](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html). This is an effort to update actions and package builds to deprecate directly calling `setup.py`.


JessicaS11:
@lsetiawan If you have a moment, could you review this PR? I finally dealt with the deprecated pkg_resources warning, and you'd helped set all the packaging up originally and I'm still not confident on packaging stuff.

PR #502, opened 2024-01-30
clean up the read module after adding cloud-read capabilities

JessicaS11:
In getting to release of v1.0.0 and getting the functionality of cloud reading in (e.g. #438, #468), we ended up with a few deprecated functions in the read module. This PR removes those and moves some of the input parsing/checking for read into a standalone function that's now tested.


JessicaS11:
@rwegener2 What are the odds you'd have a few moments in the next couple of weeks to review this PR? The diff is slightly confusing based on how GH is formatting it, but the commit messages are a pretty clear list of what's happening for navigation. I may add some more linting fixes before it's merged, but they should only be making long lines shorter so won't change the content.

icepyx/core/read.py
```diff
@@ -370,29 +252,7 @@ def __init__(
...
-                "or a glob string."
-            )
-
-        # Remove any directories from the list (these get generated during recursive
-        # glob search)
-        self._filelist = [f for f in self._filelist if not os.path.isdir(f)]
+        self._filelist = _parse_source(data_source, glob_kwargs)
 
         # Create a dictionary of the products as read from the metadata
         product_dict = {}
```
JessicaS11, line 258: since it's not changed I can't call it out directly, but I was looking at line 294 (`elif len(all_products) == 0:`). From a coding perspective it makes sense to cover this case (possible list lengths we care about are 0, 1, >1). But, given line 275 (`self.is_s3 = self.is_s3[0]`) will fail with an index out of range error if the list length is 0, and the `is2ref.extract_product()` of line 267 will fail if the file isn't valid (i.e. doesn't contain a product), could we actually generate the case where `len(all_products)==0`? Is best coding practice to include the case anyway?

rwegener2: This is a good catch. I think this is a remnant of when we used to take a product as a user argument and need to confirm that the found files match the product. Given that `is2ref.extract_product` will raise an error if no products are found I think it's safe to say that `product_dict` should never have 0 entries and this statement can be removed.

rwegener2: I realized there is a second task that this if statement is performing, which is that it's the only place we actually check to make sure a non-zero number of files was given (or found by the glob statement). I think this is an important check, but leaving it here is a bit of an obfuscated way of checking. I added an if statement in the `_parse_source` function to do that instead.

rwegener2 (approved):
Wooohooo for removing unneeded code! Nice PR, @JessicaS11!

I made two changes but I think they're pretty small - one was a syntax error in the existing code I found in is2ref.py. The other was related to the comment you left about checking if 0 products are found in the product list.

icepyx/core/read.py
```diff
@@ -370,29 +252,7 @@ def __init__(
...
-        else:
-            raise TypeError(
-                "data_source should be a list of files, a directory, the path to a file, "
-                "or a glob string."
-            )
-
-        # Remove any directories from the list (these get generated during recursive
-        # glob search)
-        self._filelist = [f for f in self._filelist if not os.path.isdir(f)]
+        self._filelist = _parse_source(data_source, glob_kwargs)
```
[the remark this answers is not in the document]
JessicaS11: Thanks!

icepyx/core/read.py
JessicaS11: Good catch. I updated the test I had checking for an empty list to check for the error instead.

PR #508, opened 2024-02-09
Add capability to created wanted variables list with an explicit list of paths

### LINKED ISSUE #504: Allow generating variables wanted from a list of h5 paths
I'm trying to subset ATL03 granules for a list of variables that I have as a list of H5 paths

```
['gt1l/heights/delta_time',
 'gt1l/heights/dist_ph_across',
 'gt1l/heights/dist_ph_along',
 'gt1l/heights/h_ph',
 'gt1l/heights/lat_ph',
 'gt1l/heights/lon_ph',
 'gt1l/heights/pce_mframe_cnt',
 'gt1l/heights/ph_id_channel',
 'gt1l/heights/ph_id_pulse',
 'gt1l/heights/quality_ph',
 'gt1l/heights/signal_conf_ph',
 'gt1l/geolocation/delta_time',
 'gt1l/geolocation/ph_index_beg',
 'gt1l/geolocation/podppd_flag',
 'gt1l/geolocation/segment_dist_x',
 'gt1l/geolocation/segment_id',
 'gt1l/geolocation/segment_length',
 'gt1l/geolocation/segment_ph_cnt',
 'gt1l/geolocation/sigma_across',
 'gt1l/geolocation/sigma_along',
 'gt1l/geolocation/sigma_h',
 'gt1l/geolocation/surf_type',
 'gt1l/geolocation/velocity_sc']
```

However, I can not find an option in `Query.order_vars.append` to do this.  I would expect to be able to do something like;

```
region_a.order_vars.append(var_list=['gt1l/heights/delta_time', 'gt1l/heights/dist_ph_across', 'gt1l/heights/dist_ph_along',...])
```
Am I missing something?

andypbarrett:
This pull request adds the capability to the `Variables.append` method to create wanted variables using a list of explicit paths.

- a `path_list` keyword argument was added to `append`.
- `parse_var_list` and `set` are used to generate unique `var_list`, `beam_list`, `keyword_list` variables.  These are then used as they had been in the original version of append. 
- the docstring for `append` was updated to describe the new keyword and give a usage example.
- keyword checking is added to ensure path_list but not `var_list`, `beam_list`, or `keyword_list` are set.  A `ValueError` is raised if this check fails.
- The check that at least one combination of keywords is set was updated to include `path_list` and is modified to use `if` instead of `assert`.  This allows a more informative Exception to be raised than an `AssertionError`


rwegener2 (approved):
Looks good to me! Thanks, @andypbarrett (especially for updating the docstring with an example, too!)

icepyx/core/variables.py
```diff
@@ -457,6 +464,9 @@ def append(self, defaults=False, var_list=None, beam_list=None, keyword_list=Non
             the product that include that keyword in their path. A list of availble keywords can be obtained by
             entering `keyword_list=['']` into the function.
 
+        path_list : list of strings, default None
+            A list of variable path names given as a the full path from the H5 root path.
```
JessicaS11, line 468:
```suggestion
            A list of variable path names given as the full path from the H5 root path.
```

icepyx/core/variables.py
```diff
@@ -482,25 +492,59 @@ def append(self, defaults=False, var_list=None, beam_list=None, keyword_list=Non
...
+            not defaults
+            and not (var_list or beam_list or keyword_list)
+            and not path_list
+        ):
+            raise ValueError(
+                "Either default or path_list, or at least one of var_list, "
+                "beam_list or keyword_list must be set\n"
+                "If you do not want to subset by variable, "
+                "ensure your is2.subsetparams dictionary does "
+                "not contain the key 'Coverage'."
```
JessicaS11, line 519: Thanks for this update to a ValueError!

Given the Variables module is used by Query, Read, and set up to work independently, I'm wondering if it might be confusing to include a "query-specific" component to the warning. Thoughts?

icepyx/core/variables.py
```diff
@@ -482,25 +492,59 @@ def append(self, defaults=False, var_list=None, beam_list=None, keyword_list=Non
...
+            )
+            var_list = list(set(variables))
+            beam_list = list(set(beams))
+            keyword_list = list(set(keywords))
+
         vgrp, allpaths = self.avail(options=True, internal=True)
         self._check_valid_lists(vgrp, allpaths, var_list, beam_list, keyword_list)
 
-        # Instantiate self.wanted to an empty dictionary if it doesn't exist
+        # Instantiate self.wanted to an empty dictionaryto an  if it doesn't exist
```
JessicaS11, line 543:
```suggestion
        # Instantiate self.wanted to an empty dictionary if it doesn't exist
```

JessicaS11:
Thanks for the great PR, @andypbarrett! My comments are mostly typos, and hopefully a fix to get the tests via Travis to pass.

andypbarrett:
Thanks @JessicaS11 and @rwegener2 .  I plan on fixing these on Thursday 15 Feb.

icepyx/core/variables.py
andypbarrett: Good point!  Should I delete these lines (517 to 519)?

Does the `Variables.append` method get called "under-the-hood" or would it always be called through `order_vars` or a `Variables` object.  That said a warning `ValueError("Either default or path_list, or at least one of var_list, beam_list or keyword_list must be set")` seems explicit and clear enough.

JessicaS11: If memory serves, `.append` is called under the hood during query or read to ensure the user has the "required" [metadata] variables (I don't recall the exact list, but for subsetting it makes sure things like delta_time are there, and for read it adds things like spot/gt that icepyx needs to parse data into xarray). But even that is done on an instance of `Variables` via `query.order_vars` or `read.vars`.

>a warning ValueError("Either default or path_list, or at least one of var_list, beam_list or keyword_list must be set") seems explicit and clear enough.

Agreed.
> Should I delete these lines (517 to 519)?

Sounds good!

andypbarrett: I removed the lines indicated above and pushed.

icepyx/core/variables.py
```diff
@@ -482,25 +492,56 @@ def append(self, defaults=False, var_list=None, beam_list=None, keyword_list=Non
...
 
         final_vars = {}
 
+        if path_list:
+            _, (beams, keywords, variables) = self.parse_var_list(
+                path_list, tiered_vars=True
+            )
+            var_list = list(set(variables))
+            beam_list = list(set(beams))
+            keyword_list = list(set(keywords))
```
JessicaS11, line 535: I went to add the new example in the docstring to the variables tutorial notebook it raised 2 issues:
1. The example fails because it's for ATL03, and the examples are on ATL06 (and in the notebook, there's a second one for ATL09)
2. Trying to use a `path_list` input on ATL06 fails because it's got four layers of variables, not the three we collect here.

I haven't reminded myself of how we handle this in general in the module, so that may recommend a solution for (2). (1) is easy if we just sub in another list.

Issue #488, opened 2024-01-04
NSIDC request url generation currently includes duplicate keywords


JessicaS11:
In digging into #457, we discovered that order request urls were being submitted with both CMR and order parameters (e.g. Order request URL:  https://n5eil02u.ecs.nsidc.org/egi/request?short_name=ATL11&version=006&bounding_box=-87.0,74.7,-85.0,75.4&page_size=2000&page_num=1&request_mode=async&include_meta=Y&client_string=icepyx&time=2018-09-01T00:00:00,2022-01-01T23:59:59&bbox=-87.0,74.7,-85.0,75.4), making it impossible to submit a request without a temporal parameter (the short-term solution to #457). Thus, we'll need to dig in to what's causing both (e.g.) "bounding_box" and "bbox" to be submitted and fix it.

JessicaS11:
In trying to address this issue, which I was thinking was a "bug", I discovered that submitting both CMR and EGI keywords (for spatial and temporal) seems to enable the subsetter to get a shorter list of granules to try and process (i.e., use CMR to narrow the granule search before actually trying to subset them for the order). Specifically (with primary differences in the submitted request url and in the returned "Instructions"):

Case both CMR and EGI keywords are submitted
```
https://n5eil02u.ecs.nsidc.org/egi/request?short_name=ATL06&version=006&temporal=2019-02-20T00:00:00Z,2019-02-28T23:59:59Z&bounding_box=-55.0,68.0,-48.0,71.0&page_size=2000&page_num=1&request_mode=async&include_meta=Y&client_string=icepyx&time=2019-02-20T00:00:00,2019-02-28T23:59:59&bbox=-55.0,68.0,-48.0,71.0
Order request response XML content:  <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<eesi:agentResponse xmlns="" xmlns:iesi="http://eosdis.nasa.gov/esi/rsp/i" xmlns:ssw="http://newsroom.gsfc.nasa.gov/esi/rsp/ssw" xmlns:eesi="http://eosdis.nasa.gov/esi/rsp/e" xmlns:esi="http://eosdis.nasa.gov/esi/rsp" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <order>
        <orderId>5000005341041</orderId>
[...]

PR #515, opened 2024-02-27
Clean up APIformatting module and fix ATL11 temporal kwarg submission

JessicaS11:
See #457, which results in an error if temporal keywords are submitted with a request for ATL11. This prompted #488, which it turns out having both CMR and EGI keywords submitted in the request is the desired behavior.

Tasks completed:
- During this debugging and exploration, it was noted that the "default" key in `_get_possible_keys` wasn't being used. This was removed, and "short_name" and "version" were moved from the CMR list to the required list.
- The explicit passing of a session into `Granules` was also removed and replaced with the `EarthdataAuthMixin`.
- Touched files were also cleaned up for linting.


JessicaS11:
@whyjz What are the odds you'd be able to review this PR (or if not review the whole thing, since it has multiple changes, test it against your ATL11 workflow)? It would be greatly appreciated!

whyjz:
@JessicaS11 Thanks a lot for this! I tested the `orderurl` branch with the following code:
```
reg_b = ipx.Query("ATL11", [-87, 74.7, -85, 75.4], ['2018-09-01','2022-01-01'])
reg_b.download_granules('./icepyx_test/')
```
Now it seemed to go through the order request (with the message showing `Your order is: complete`) but stopped immediately after that with this error message:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 reg_b.download_granules('./icepyx_test/')

File ~/Projects/Github/icepyx/icepyx/core/query.py:1113, in Query.download_granules(self, path, verbose, subset, restart, **kwargs)
   1107     if (
   1108         not hasattr(self._granules, "orderIDs")
   1109         or len(self._granules.orderIDs) == 0
   1110     ):
   1111         self.order_granules(verbose=verbose, subset=subset, **kwargs)
-> 1113 self._granules.download(verbose, path, session=self.session, restart=restart)

TypeError: Granules.download() got an unexpected keyword argument 'session'
```
The same error also happens when I run the almost identical code to download ATL06:
```
reg_a = ipx.Query('ATL06',[-55, 68, -48, 71],['2019-02-20','2019-02-28']) 
reg_a.download_granules('./icepyx_test/')
```
This error does not appear when I use the code from the `development` branch. So I guess we solved the original issue, but the new changes also messed up something unexpectedly?

icepyx/core/granules.py
```diff
@@ -473,7 +462,7 @@ def place_order(
 
         return self.orderIDs
 
-    def download(self, verbose, path, session=None, restart=False):
+    def download(self, verbose, path, restart=False):
```
whyjz, line 465: It looks like this is the change that conflicts with `icepyx/core/query.py:1113` (see my most recent comment for details)

JessicaS11: Nice catch! I'll approve and merge #521 - thanks for finding the issue and submitting a fix (the failing CI tests are behind authentication, so they fail because forks cannot access the secrets). I will send you an invite to the icepyx dev team here so you can edit the PR directly, too.

whyjz (approved):
This all looks good to me! I am not sure how to resolve the Codecov insufficient coverage, though.

JessicaS11:
> I am not sure how to resolve the Codecov insufficient coverage, though.

Unfortunately, the answer to that would be improving our test coverage (which I strive to do, but isn't always practical). In this case, what's happening is Codecov checks to see if the lines of code modified in the PR are tested and compares the percent tested of those lines to the percent tested of the software (overall) on the development branch. If the line of code that was modified did not already have a test covering it, it brings down the "patch" coverage percentage and thus more likely to fail this check. There's not a great way to fix this without having the modified lines of code tested to start with, and my code testing skills are novice enough to make this a nontrivial task. The Codecov checks are still really helpful though for finding some of the lower lift ways to improve test coverage, even if they don't always pass.

Thanks so much for the PR review!

PR #529, opened 2024-06-18
fix bib entry

JessicaS11:


JessicaS11:
Note: the test failure is likely caused by breaking changes with Numpy 2.0 ([e.g. as reported here](https://github.com/astropy/pyvo/issues/464)). Because the output now includes the dtype, the strings don't match. We'll need to update our tests once we require Numpy >=2.0.

PR #539, opened 2024-08-06
Replace `setup.py` with equivalent `pyproject.toml`

mfisher87:
:wave:

https://setuptools.pypa.io/en/latest/userguide/quickstart.html#transitioning-from-setup-py-to-declarative-config


pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+    "Programming Language :: Python :: 3.12",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Scientific/Engineering :: GIS",
+    "Topic :: Software Development :: Libraries",
+]
+
+[project.urls]
+Homepage = "https://icepyx.readthedocs.io"
+Documentation = "https://icepyx.readthedocs.io"
+Repository = "https://github.com/icesat2py/icepyx.git"
```
weiji14, line 37: Can probably omit the `.git` part from the URL.

```suggestion
Repository = "https://github.com/icesat2py/icepyx"
```

weiji14 (approved):
Tested locally using `python -m pip install -U git+https://github.com/icesat2py/icepyx.git@pyproject` and install worked:

```
Building wheels for collected packages: icepyx
  Building wheel for icepyx (pyproject.toml) ... done
  Created wheel for icepyx: filename=icepyx-1.1.1.dev6+g0c167ac-py3-none-any.whl size=84019 sha256=639020f3dae60c23936d3658e4b0a92a47855f997587f693fe58c6de632b14b4
  Stored in directory: /tmp/pip-ephem-wheel-cache-263bvbb_/wheels/82/71/dc/27dcd9147f97ee37fa865a9c6c34839d9163fb43e99929bc32
Successfully built icepyx
Installing collected packages: icepyx
Successfully installed icepyx-1.1.1.dev6+g0c167ac
```

Just one other small suggestion below :point_down:

pyproject.toml
mfisher87: Woops, thanks for catching that!

mfisher87:
Thanks @weiji14 for the review! I'm looking at those UML changes and they look really similar, but shifted around a little bit. Should I just ignore them?

weiji14:
> I'm looking at those UML changes and they look really similar, but shifted around a little bit. Should I just ignore them?

Probably ok to ignore them, I think the UML generator isn't very deterministic... Can check with @JessicaS11?

pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+
+[tool.setuptools]
+py-modules = ["_icepyx_version"]
+
+[tool.setuptools.dynamic]
+dependencies = {file = ["requirements.txt"]}
+
+[project.optional-dependencies]
+viz = ["geoviews >= 1.9.0", "cartopy >= 0.18.0", "scipy"]
+complete = ["icepyx[viz]"]
```
weiji14, line 54: Just wanted to check on this line. Will "icepyx[viz]" pull in the `viz` extras based on the icepyx version on PyPI, or the current dev icepyx version?

mfisher87: Tested this by adding `cowsay` to the list and then `pip install --editable .[complete]`, and `cowsay` was installed, so `pyproject.toml` is the source for an editable install  like I hoped :tada: :)

weiji14:
Oops, the publish to TestPyPI CI is failing, @mfisher87, could you update this line as well?

https://github.com/icesat2py/icepyx/blob/38ebec9dfe01f510b45326e962df6a035d83bda6/.github/workflows/publish_to_pypi.yml#L45

mfisher87:
Woops! My mistake! https://github.com/icesat2py/icepyx/pull/541

pyproject.toml
```diff
@@ -0,0 +1,63 @@
+[project]
+name = "icepyx"
+description = "Python tools for obtaining and working with ICESat-2 data"
+license = {file = "LICENSE"}
+readme = "README.rst"
+
+requires-python = "~=3.7"
```
weiji14, line 7: Oh shoot, totally overlooked this line. This should have been `>=3.7` instead...

mfisher87: On it!

PR #540, opened 2024-08-08
Update viz doc to clarify its current status (partial functionality - incompatible with OA API changes)

mfisher87:
Related: #454


icepyx/core/visualization.py
```diff
@@ -312,7 +312,13 @@ def make_request(self, base_url, payload):
         --------
         request_OA_data
         """
-        return requests.get(base_url, params=payload)
+        response = requests.get(base_url, params=payload)
+        if not response.ok:
+            raise RuntimeError(
+                f"Status {response.status_code} requesting url {response.request.url}"
+            )
```
mfisher87, line 319: We're currently getting a 404 from the API, and later on we try to access the "data" key from the response. Instead of a KeyError at that time, we'll now error at the point of failure.

mfisher87:
Thoughts on the coverage change? The function I changed is currently not covered, but I can make it covered. I'd probably mock the response and verify a 400 raises and a 200 doesn't, but my personal feeling is there's not a ton of value there :)

icepyx/core/visualization.py
JessicaS11: > We're currently getting a 404 from the API

Good to know - I haven't been tracking how the response has changed as things have on the API side.

> Instead of a KeyError at that time, we'll now error at the point of failure.

Awesome!

JessicaS11:
> Thoughts on the coverage change? The function I changed is currently not covered, but I can make it covered. I'd probably mock the response and verify a 400 raises and a 200 doesn't, but my personal feeling is there's not a ton of value there :)

Agreed. Since #454, the broader conversation around this module is that given the lack of response to #495 and minimal complaints that this module stopped working, it may make sense to rethink how we offer any quick data visualization within icepyx.

doc/source/example_notebooks/IS2_data_visualization.ipynb
```diff
@@ -7,7 +7,13 @@
    "source": [
```
JessicaS11, line 7: <p>The whole notebook isn't non-functional. It should work up until the "Visualize ICESat-2 elevation using OpenAltimetry API" header.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/icesat2py/icepyx/pull/540/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

mfisher87: I meant to communicate more that it's not going to get to the intended point due to the error, so I thought it made sense to callout up-front "this doc may not take you where you want to go". What do you think about how to adjust the wording?

mfisher87:
From #454, perhaps we can fix the notebook by updating the URL. I have no idea if the API has breaking changes or not yet :grimacing:

doc/source/example_notebooks/IS2_data_visualization.ipynb
JessicaS11: <p>That makes sense - and I like the callout happening at the start. I like how you wrote it if you read it immediately after the notebook description, but I skipped right to the "Danger" part (so it's working as intended and got my attention!). How about something like: "Some of this notebook is currently non-functional"?</p>

mfisher87: Sounds good! :rocket:  Thanks Jessica!

mfisher87: 9c9e722

JessicaS11 (approved):
I like the updates to the warning too - thanks @mfisher87!

mfisher87:
Thanks for the review! :rock: :star:

PR #542, opened 2024-08-08
Autofix flake8 ignores E711, E712, E714, F401, F841 with Ruff

mfisher87:
I haven't added Ruff as a dependency, but I do think we should replace flake8 with it because it's capable of autofixing. I've seen an immense difference in cognitive load and labor over the last year or so I've used Ruff! Thoughts?

### DIFF
--- a/icepyx/core/query.py
+++ b/icepyx/core/query.py
@@ -1,6 +1,5 @@
 import geopandas as gpd
 import matplotlib.pyplot as plt
-from pathlib import Path  # used in docstring tests
 import pprint
 
 import icepyx.core.APIformatting as apifmt
@@ -103,9 +102,10 @@ class GenQuery:
 
     Initializing Query with a geospatial polygon file.
 
-    >>> aoi = str(Path('./doc/source/example_notebooks/supporting_files/simple_test_poly.gpkg').resolve())
+    >>> from pathlib import Path
+    >>> aoi = Path('./doc/source/example_notebooks/supporting_files/simple_test_poly.gpkg').resolve()
     >>> reg_a_dates = ['2019-02-22','2019-02-28']
-    >>> reg_a = GenQuery(aoi, reg_a_dates)
+    >>> reg_a = GenQuery(str(aoi), reg_a_dates)
     >>> print(reg_a)
     Extent type: polygon
     Coordinates: [-55.0, 68.0, -55.0, 71.0, -48.0, 71.0, -48.0, 68.0, -55.0, 68.0]
@@ -378,9 +378,10 @@ class Query(GenQuery, EarthdataAuthMixin):
 
     Initializing Query with a geospatial polygon file.
 
-    >>> aoi = str(Path('./doc/source/example_notebooks/supporting_files/simple_test_poly.gpkg').resolve())
+    >>> from pathlib import Path
+    >>> aoi = Path('./doc/source/example_notebooks/supporting_files/simple_test_poly.gpkg').resolve()
     >>> reg_a_dates = ['2019-02-22','2019-02-28']
-    >>> reg_a = Query('ATL06', aoi, reg_a_dates)
+    >>> reg_a = Query('ATL06', str(aoi), reg_a_dates)
     >>> print(reg_a)
     Product ATL06 v006
     ('polygon', [-55.0, 68.0, -55.0, 71.0, -48.0, 71.0, -48.0, 68.0, -55.0, 68.0])
@@ -1132,7 +1133,7 @@ class Query(GenQuery, EarthdataAuthMixin):
         gdf = self._spatial.extent_as_gdf
 
         try:
-            from shapely.geometry import Polygon
+            from shapely.geometry import Polygon  # noqa: F401
             import geoviews as gv
 
             gv.extension("bokeh")
--- a/icepyx/core/read.py
+++ b/icepyx/core/read.py
@@ -633,6 +633,7 @@ class Read(EarthdataAuthMixin):
             )  # wanted_groups, vgrp.keys()))
 
             # Closing the file prevents further operations on the dataset
+            # from s3fs.core import S3File
             # if isinstance(file, S3File):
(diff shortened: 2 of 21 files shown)


weiji14:
> I haven't added Ruff as a dependency, but I do think we should replace flake8 with it because it's capable of autofixing. I've seen an immense difference in cognitive load and labor over the last year or so I've used Ruff! Thoughts?

Maybe we can do a two-step process. Fix the flake8 errors here, and then another PR to migrate from `flake8` -> `ruff`? To make sure the transition is ok, you can leave some small flake8 errors around so we can check that the autofix works later.

mfisher87:
@weiji14 #543 I was already working on it! :rocket:

weiji14:
Cool, I'll just note that code coverage on icepyx is a bit low (<70%), so you'll want to be a bit careful with refactoring :slightly_smiling_face:

mfisher87:
As I'm working on #544 I hope to increase coverage!

icepyx/core/query.py
```diff
@@ -1132,7 +1133,7 @@ def visualize_spatial_extent(
         gdf = self._spatial.extent_as_gdf
 
         try:
-            from shapely.geometry import Polygon
+            from shapely.geometry import Polygon  # noqa: F401
```
JessicaS11, line 1136: I didn't know you could do this - awesome!

icepyx/core/query.py
```diff
@@ -1,6 +1,5 @@
 import geopandas as gpd
 import matplotlib.pyplot as plt
-from pathlib import Path  # used in docstring tests
```
JessicaS11, line 3: Is there a best practice for cases like this where it means you now have to have the import line twice (i.e. how many times do you need to import it for examples before it's enough to import it generally?)

icepyx/core/read.py
```diff
@@ -5,7 +5,6 @@
 
 import earthaccess
 import numpy as np
-from s3fs.core import S3File
```
JessicaS11, line 8: Can we comment this one out for now instead of removing? The linter is catching it because the line of code is commented (line 637), but ultimately I'd like to have better handling there that makes sure the file gets closed.

JessicaS11 (approved):
I could have sworn I'd handled most of the "imported but unused" errors (F401). Apparently not... this looks great, @mfisher87!

icepyx/core/query.py
mfisher87: None that I know of. My personal opinion is that this is a case where WET is a better practice than DRY. The examples in the tests end up in the docs, so IMO they should be self-contained and complete!

icepyx/core/read.py
mfisher87: I'm cool with that. What do you think of including the commented import with the commented code? Then it's obvious to a reader not to erase the commented import. And when we un-comment the code, our linter will ensure we put the import in the right place!

weiji14: Yep, commenting that line and moving it down below sounds good, can you apply that change please?

weiji14 (approved):
Pre-approving since this looks good (mostly changes of `==` -> `is` and removing unused imports). Should be ok to merge once https://github.com/icesat2py/icepyx/pull/542#discussion_r1712333998 is addressed.

icepyx/core/read.py
mfisher87: Can do!

mfisher87: ad95e2e

mfisher87:
There's a failing test! I found that test is non-deterministic. It relies on the order of a list not being changed when it's converted to a set and then back to a list, then compares that list with another list. The reason this is a problem is that [sets are unordered](https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset), but lists are ordered:

> Being an unordered collection, sets do not record element position or order of insertion.

We can verify. Takes a few tries.  


```bash
$ python -c "print(list({'temperature', 'salinity'}))"
['temperature', 'salinity']

$ python -c "print(list({'temperature', 'salinity'}))"
['temperature', 'salinity']

$ python -c "print(list({'temperature', 'salinity'}))"
['temperature', 'salinity']

$ python -c "print(list({'temperature', 'salinity'}))"
['temperature', 'salinity']

$ python -c "print(list({'temperature', 'salinity'}))"
['temperature', 'salinity']

$ python -c "print(list({'temperature', 'salinity'}))"
['salinity', 'temperature']
```

I believe, under the hood, the set is sorted by object hash:

```bash
$ python -c "print(hash('temperature'))"
-7033926737989545164

$ python -c "print(hash('temperature'))"
7247647540643673123
```

icepyx/tests/test_quest_argo.py
```diff
@@ -59,7 +59,7 @@ def test_param_setter(argo_quest_instance):
 
     reg_a.params = ["temperature", "salinity"]
 
-    exp = list(set(["temperature", "salinity"]))
+    exp = ["temperature", "salinity"]
```
weiji14, line 62: > > The reason this is a problem is that [sets are unordered](https://docs.python.org/3/library/stdtypes.html#set-types-set-frozenset), but lists are ordered

TIL! Thanks for the in-depth explanation!

JessicaS11:
> There's a failing test! I found that test is non-deterministic.

This test has been ailing us for some time (and often if I just rerun the CI it passes). I think the `list(set())` bit got added to try and fix that (it clearly didn't!).

> Not clear if this test is actually useful

If memory serves, there's a setter for that method so it might be trying to get at defaults/valid inputs. If that's the case, I'd be happy to learn if there's a better way!

mfisher87:
For this test, should order matter? We can do a set comparison instead of list comparison if it doesn't.

mfisher87:
https://app.travis-ci.com/github/icesat2py/icepyx/builds/271849698

I think this answers my question: The setter can change the order of the input data. I will change the test to do set comparison!

PR #543, opened 2024-08-08
Switch to ruff

mfisher87:
@JessicaS11 and I had a chat about this on the phone today. I feel Ruff's capacity to automatically fix so much stuff is a game-changer! I have to think about linting much less.


pyproject.toml
```diff
@@ -61,3 +61,38 @@ version_file = "_icepyx_version.py"
...
+  # overlong comments.
+  # See: https://github.com/psf/black/issues/1713#issuecomment-1357045092
+  "E501",
+  # TODO: remove ignores below this line
+  # comparison syntax in tests
+  "E721",
+  # bare except
+  "E722",
+  # unable to detect undefined names
+  "F403",
```
mfisher87, line 85: Some of these are pretty easy to polish off. F403 is one -- here we import `*` in an `__init__.py`. Instead, we should import the objects we want to expose by name. Any thoughts on what objects we want to expose here?

https://github.com/icesat2py/icepyx/blob/development/icepyx/quest/dataset_scripts/__init__.py

.github/workflows/linter_actions.yml
```diff
@@ -10,9 +10,10 @@ jobs:
...
-        uses: psf/black@stable
-      # use the flake8 linter to annotate improperly formatted code
-      # note linter arguments are supplied via the .flake8 config file
-      - name: Annotate PR after running flake8
-        uses: TrueBrain/actions-flake8@v2
+
+      # Use the Ruff linter to annotate code style / best-practice issues
+      # NOTE: More config provided in pyproject.toml
+      - name: Lint and annotate PR
+        uses: chartboost/ruff-action@v1
```
weiji14, line 17: Looks like this GitHub Action isn't compatible with ruff>0.5.0 (https://github.com/ChartBoost/ruff-action/issues/30) and there's also a separate issue that they're looking into transferring this to another organization (which might take a while).

So... could we just use [`pre-commit.ci`](https://pre-commit.ci/) instead? Or are the annotations good to have?

mfisher87: I'm cool with just using pre-commit.ci. I was trying to avoid any regression in functionality. I personally don't find annotation of PRs to be that valuable, but I imagine the feature would be more valuable to someone who doesn't know how to use the linter at the CLI. :thinking: 

It looks to me from reading the issue like this isn't an incompatibility issue _yet_ though? Or has Ruff gone forward with the breaking change?

mfisher87: Fixed linting on this PR. I was using an outdated argument form!

.flake8
```diff
@@ -1,37 +0,0 @@
...
-    E741
-    # unable to detect undefined names
-    F403
-    # line break before binary operator
-    # NOTE: This is a formatting concern
-    W503
-
-    # GOAL:
-    # syntax check doctests in docstrings
-    # doctests = True
```
weiji14, line 37: No way to syntax check docstrings in ruff yet - https://github.com/astral-sh/ruff/issues/3542. But we can at least format the docstrings, see below.

pyproject.toml
```diff
@@ -61,3 +61,38 @@ version_file = "_icepyx_version.py"
 version_file_template = 'version = "{version}"'
 local_scheme = "node-and-date"
 fallback_version = "unknown"
+
```
weiji14, line 64: Using https://docs.astral.sh/ruff/formatter/#docstring-formatting to format the docstrings. This will result in a big diff, so commenting out for now

```suggestion
# [tool.ruff.format]
# docstring-code-format = true
# docstring-code-line-length = "dynamic"
```

pyproject.toml
```diff
@@ -61,3 +61,38 @@ version_file = "_icepyx_version.py"
...
+
+[tool.ruff.lint]
+select = [
+  "E",
+  "F",
```
weiji14, line 69: Annotate what these single-letter rules are.

```suggestion
  "E",    # pycodestyle
  "F",    # pyflakes
```

pyproject.toml
weiji14, line 1: The indent on this TOML file is a bit inconsistent, there are 4-spaces and 2-spaces, should maybe standardize on one.

.flake8
mfisher87: Are we executing doctests? That'll give us the syntax check!

pyproject.toml
mfisher87: Great call :)

pyproject.toml
mfisher87: 447b8f4

.flake8
weiji14: No, not yet, but can do that later!

weiji14 (approved):
Super, thanks again @mfisher87!

.github/workflows/linter_actions.yml
JessicaS11: > I personally don't find annotation of PRs to be that valuable, but I imagine the feature would be more valuable to someone who doesn't know how to use the linter at the CLI.

You've identified exactly why we have the PR annotations set up. Precommit is an additional barrier to new contributors, so I wanted to be able to enforce the linting and formatting without absolutely requiring they complete yet another setup step to contribute.

mfisher87: :100:

requirements-dev.txt
```diff
@@ -1,5 +1,3 @@
-black
-flake8
```
JessicaS11, line 2: Suggest we add ruff here (unless it only ever runs on CI?)

mfisher87: It runs in GitHub Actions and in pre-commit.ci, but seems useful to also have it in the dev requirements!

PR #546, opened 2024-08-08
More ruff & pre-commit rules

mfisher87:


mfisher87:
Since pre-commit on the default branch is configured to use Black without isort, pre-commit.ci is fighting with the new formatter config. Because of that, we'll need to format again after merging. I prefer to turn off this auto-commit behavior so it can be controlled by the user posting an autofix comment.

.pre-commit-config.yaml
```diff
@@ -6,9 +6,17 @@ repos:
...
+      - id: check-merge-conflict
+      - id: check-symlinks
+      - id: debug-statements
+      - id: name-tests-test
+        args: ["--pytest-test-first"]
       - id: end-of-file-fixer
+      - id: mixed-line-ending
       - id: trailing-whitespace
         args: [--markdown-linebreak-ext=md]
+      - id: requirements-txt-fixer
```
weiji14, line 19: Won't be needed if https://github.com/icesat2py/icepyx/issues/552 goes ahead.

.pre-commit-config.yaml
```diff
@@ -6,9 +6,17 @@ repos:
...
       - id: check-added-large-files
         args: ["--maxkb=5000"]
+      - id: check-case-conflict
+      - id: check-merge-conflict
+      - id: check-symlinks
+      - id: debug-statements
+      - id: name-tests-test
+        args: ["--pytest-test-first"]
       - id: end-of-file-fixer
+      - id: mixed-line-ending
```
weiji14, line 16: What about using https://docs.astral.sh/ruff/settings/#format_line-ending instead?

pyproject.toml
```diff
@@ -65,10 +65,32 @@ fallback_version = "unknown"
 # docstring-code-format = true
 # docstring-code-line-length = "dynamic"
 
+
 [tool.ruff.lint]
-select = [
-  "E",    # pycodestyle
-  "F",    # pyflakes
+extend-select = [
```
weiji14, line 70:
```suggestion
select = [
```

weiji14 (commented):
Fixed thee crazy merge conflict. Maybe we should sort the rules under `[tool.ruff.lint]` alphabetically too?

.pre-commit-config.yaml
mfisher87: This will catch non-python files as well!

weiji14: Oh yes, let's keep this then!

pyproject.toml
```diff
@@ -65,10 +65,32 @@ fallback_version = "unknown"
...
+  # "PTH",      # flake8-use-pathlib
+  # "RET",      # flake8-return
+  # "RUF",      # Ruff-specific
+  # "SIM",      # flake8-simplify
+  # "T20",      # flake8-print
+  # "UP",       # pyupgrade
+  "YTT",      # flake8-2020
+  "EXE",      # flake8-executable
+  "NPY",      # NumPy specific rules
+  # "PD",       # pandas-vet
```
weiji14, line 93: Sort alphabetically
```suggestion
  # "ARG",  # flake8-unused-arguments IMPORTANT
  # "B",    # flake8-bugbear IMPORTANT
  # "C4",   # flake8-comprehensions
  "E",    # pycodestyle
  # "EM",   # flake8-errmsg
  "EXE",  # flake8-executable
  "F",    # pyflakes
  "G",    # flake8-logging-format
  "I",    # isort
  "ICN",  # flake8-import-conventions
  "NPY",  # NumPy specific rules
  # "PD",   # pandas-vet
  "PGH",  # pygrep-hooks
  "PIE",  # flake8-pie
  # "PL",   # pylint
  # "PT",   # flake8-pytest-style
  # "PTH",  # flake8-use-pathlib
  # "RET",  # flake8-return
  # "RUF",  # Ruff-specific
  # "SIM",  # flake8-simplify
  # "T20",  # flake8-print
  # "UP",   # pyupgrade
  "YTT",  # flake8-2020
```

mfisher87:
Interesting. I've not worked on a project that requires squash merges for more than a PR or two. It really causes problems for chained PRs. Is there a common practice for dealing with this?

mfisher87:
I'm not sure what's up with Travis. I get a blank response: https://app.travis-ci.com/github/icesat2py/icepyx/builds/271864435

weiji14:
> Interesting. I've not worked on a project that requires squash merges for more than a PR or two. It really causes problems for chained PRs. Is there a common practice for dealing with this?

Best practice is to have small commits that don't change so many files at once :slightly_smiling_face: I've got a good side-by-side merge conflict resolving tool with [Gitkraken](https://www.gitkraken.com/developer-problems/merge-conflicts), so it wasn't too bad to resolve things.

> If we always updated the feature branch via rebase, then it would be a bit less messy.

Personally I dislike rebases because as a reviewer, it undos all the 'Viewed' checkboxes I've clicked on the 'Files Changed' tab, and I have to go back and re-read the lines of code (that may not have changed), instead of focusing on the lines that have actually changed. Also, if there are co-authored commits, some of those commits will become 'Unverified' because you do not have the co-author's GPG signing keys.

> Briefly, I'd like to ask if there's any way you can think of we can utilize your review time more efficiently! I really appreciate all the work you've been doing on this! Would you be OK with "conditional approvals" in cases like this one where there are some "must-change" items, like sort order or config keys, and some "prefer-change" items like removing the pyproject checker rule, and once the "must-change" items are addressed, the PR can merge without additional review? I'm totally OK with waiting for a final check, I'm just worried I'm taking a lot of your time! ❤️

First off, I think you're doing a good job at this. Icepyx's development has been focused more on features for the longest time (i.e. as long as it works, doesn't matter if it's not formatted nicely). Usually in unformatted codebases like this, I'd use an incremental linter like [`darker`](https://pypi.org/project/darker/) to only ensure the lines I change are formatted nicely rather than doing blanket changes on the whole codebase. But I do think that it's time for icepyx to look a bit more professional, and having newer linters like `ruff` with quick autofix definitely helps, so really appreciate you taking the effort here with >2 PRs!

On the topic of "must-" vs "prefer-" changes, one way I usually do this is to pre-approve once all the "must-" changes are satisfied, and the "prefer-" changes is left to the contributor to decide if they want to apply it or not. Generally, I try to suggest changes that will make it easier for the next person looking at the code, whether it's by making it clear where a new Ruff rule will go in (hence alphabetical sorting), or minimizing the diff between commits (easier to do `git blame` later). If you don't mind waiting, it's usually good to leave PRs that are almost ready sitting there for a few days, just so there's a chance for more eyeballs to look at it. The process here is still quicker than traditional peer review :laughing:

pyproject.toml
```diff
@@ -65,10 +65,32 @@ fallback_version = "unknown"
 # docstring-code-format = true
 # docstring-code-line-length = "dynamic"
 
+
 [tool.ruff.lint]
 select = [
+  # "ARG",  # flake8-unused-arguments IMPORTANT
+  # "B",    # flake8-bugbear IMPORTANT
```
weiji14, line 72: Why are these marked important?

weiji14 (approved):
With that said :point_up:, approving with one comment :rocket:

mfisher87:
>Best practice is to have small commits that don't change so many files at once 🙂 

That's my usual practice, that's the tradeoff with autofixes :) Please let me know if see any instances where I can break things up better!

>Personally I dislike rebases because as a reviewer, it undos all the 'Viewed' checkboxes I've clicked on the 'Files Changed' tab, and I have to go back and re-read the lines of code (that may not have changed), instead of focusing on the lines that have actually changed

Agree, that's frustrating. On the flip-side, all of the commits from a dependent PR _are_ in here :(

> ...on reviewing stuff...

That all sounds great! Do you have any thoughts about a process for reviewing `v2`? Do you want to review along the way before each branch is merged to `v2`, review along the way as time allows pre- or post-merge, review `v2` before it's merged in to `development`, something else, or some combination?

pyproject.toml
mfisher87: Bugbear because I feel it's really effective at catching latent bugs and I _think_ it detected some stuff that concerned me (I don't remember the specifics so I won't commit to this :laughing:). unused-args detected several cases of functions with unused arguments, and while that's not the end of the world, I personally find unused arguments act as red herrings when reading code, which can cause us to write bugs in the future!

mfisher87:
Another thought on squash merges: on other projects I've used [`.git-blame-ignore-revs`] to ignore changes which are strictly formatting when using `git blame` (also works in GitHub's blame feature). Because we squash, we can't take advantage of that!

icepyx/core/exceptions.py
```diff
@@ -3,16 +3,12 @@ class DeprecationError(Exception):
...
-    pass
-
 
 class QueryError(Exception):
     """
     Base class for Query object exceptions
     """
 
-    pass
-
```
JessicaS11, line 15: why isn't this needed (isn't the class empty otherwise)?

icepyx/core/granules.py
```diff
@@ -1,18 +1,19 @@
...
 import zipfile
+from xml.etree import ElementTree as ET
+
+import numpy as np
+import requests
 
 import icepyx.core.APIformatting as apifmt
-from icepyx.core.auth import EarthdataAuthMixin
 import icepyx.core.exceptions
+from icepyx.core.auth import EarthdataAuthMixin
```
JessicaS11, line 16: So either the linter and I disagree on the order of the letters of the alphabet, or there's a rule that says anything with `from` comes after anything starting with `import`... is this a setting/preference or PEP rule I just never knew about?

pyproject.toml
JessicaS11: Is this list the full list of linting options ruff has, and the ones that are not commented are the ones it will address (i.e. change)?

I like the idea of eventually including the options you've noted as IMPORTANT as ways to improve code as we go (or they could be part of a targeted effort). Perhaps labeling them as "DEV-GOAL" or "PLANNED-FOR-FUTURE" is a clearer way to indicate this in the meantime?

icepyx/core/exceptions.py
mfisher87: When a class has a docstring, it doesn't need a placeholder like `pass` or `...`.

icepyx/tests/test_is2class_query.py
JessicaS11, line 1: We'll have to revert this file rename if we want CI to pass (the tests in it are not working - I can't recall at this point if they ever did (e.g. were an early attempt on my part to add tests when I was just learning to write them). I'd left `test` out of the filename so it wasn't picked up by pytest.

icepyx/core/exceptions.py
JessicaS11: Fascinating. I had no idea. Thanks for teaching me!

icepyx/core/granules.py
mfisher87: >So either the linter and I disagree on the order of the letters of the alphabet, or

:laughing: Hey, maybe the linter is just living its bliss! :woozy_face: :laughing: 

PEP8 doesn't specify how imports should be ordered within groups, IIRC. This is a formatting behavior from [the "isort" ruleset](https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html#controlling-how-isort-sections-from-imports). In each import group, "from" imports are sub-grouped below "straight" imports. The behavior can be disabled, but I personally find it makes the imports much easier to read because there's less zigging and zagging :)

icepyx/tests/test_is2class_query.py
mfisher87: Good catch on this. I was looking for why the tests failed and I kept glossing over this rename!

pyproject.toml
mfisher87: I believe it was the full list at one time, but I copied this from another project. It's likely there are new rulesets that aren't included. The ones that are not commented are the ones that will be checked and changed, correct.

Good call on clearer wording! Would you prefer to have this reflected in the comment, or as a GitHub issue to enable those rules? I have no strong preference, I think I did it the way I did because I "flowing" at the time and wrote it down in situ to think about more later :)

icepyx/core/granules.py
JessicaS11: > but I personally find it makes the imports much easier to read because there's less zigging and zagging :)

Totally makes sense - and of course I find it makes them harder to read because I have to read through more lists (and those lists aren't separated by newlines)!

This is probably one of those rock-paper-scissors decisions, yeah?

pyproject.toml
JessicaS11: > I believe it was the full list at one time, but I copied this from another project.

Does it then make sense to just remove the ones that are commented that we're not "planning" for future dev?

> Would you prefer to have this reflected in the comment, or as a GitHub issue to enable those rules?

Personally I'm happy with the comment (I've loosely used "DevGoal" like this throughout the code, so I can do a search and find things I hoped/planned to do in the past). For a while I also added things as issues, but because I was the primary contributor it meant I was spending a lot of time on admin of the issues instead of more dev work, so I started being more selective about when I opened an issue. Not the most helpful answer, but if you have a personal preference feel free to exercise that!

icepyx/core/granules.py
mfisher87: I'm happy for you to decide, I'm the newbie around here :)

mfisher87: If we change our mind later, it's a simple config change and reformat away! :rocket:

pyproject.toml
mfisher87: >Does it then make sense to just remove the ones that are commented that we're not "planning" for future dev?

Personally, on most projects I have all of these rules on, and I think it's not unrealistic to target that for v2!

icepyx/tests/test_is2class_query.py
mfisher87: How do you feel about this path forward? 3956ca7

It felt better to switch to an explicit skip instruction than to add an exception to our linter. But I'm open to the other approach, this was just a gut thing for me :)

pyproject.toml
mfisher87: a4cf9df

Happy to make more changes, let me know your thoughts!

icepyx/core/granules.py
JessicaS11: Can we leave it as abc by name (regardless of "from" or "import" preceding) for now? Thanks!

> If we change our mind later, it's a simple config change and reformat away! 🚀

Awesome - that's super good to know.

icepyx/tests/test_is2class_query.py
JessicaS11: > It felt better to switch to an explicit skip instruction than to add an exception to our linter. But I'm open to the other approach, this was just a gut thing for me :)

Personally I'm more likely to notice (and do something about) it again in the future if it's an exception I'm trying to remove from the linter than a skip instruction within the file (why would I open the file, which now has a matching filename pattern, to notice the tests in it aren't being run?). I use the fact that it's "breaking a rule" (doesn't have test in the filename or needs an exception) as a reminder to myself that it's something that could be improved without having to manually track it as an issue. That said, I can see why an explicit skip instruction is more "robust". In using the explicit practice, how would you then track this "to do" item to un-skip the tests and either implement or remove them at a later point in time?

mfisher87: >I use the fact that it's "breaking a rule" (doesn't have test in the filename or needs an exception) as a reminder to myself that it's something that could be improved [...] In using the explicit practice, how would you then track this "to do" item to un-skip the tests and either implement or remove them at a later point in time?

I would use a GitHub issue for this, perhaps creating a special label for this type of issue so we can prioritize and have a place to discuss in the context of other project concerns. In my experience, the trouble with using breaking a rule as an indicator to come back to the thing is that it lacks context. It's one of those things I've cursed past-me for many times :laughing: I often forget why past-me did things. It's often not clear whether the rule was broken deliberately, and without that context, someone could either waste time working on something that's not ready to be worked, or they could decide that the deliberate rule-breakage was a personal preference instead of a reminder, and decide not to change (or worse, not discuss!) something that actually needs changing!

icepyx/core/granules.py
mfisher87: TODO: Force sort within sections https://pycqa.github.io/isort/docs/configuration/options.html#force-sort-within-sections

mfisher87: 3c22771

PR #550, opened 2024-08-12
Switch to set comparison to fix a test dependent on unreliable order

mfisher87:
> [!NOTE]
> I'm not sure this is the right decision! Should we instead change the setter so that it preserves order of parameters passed in?

### DIFF
--- a/icepyx/tests/test_quest_argo.py
+++ b/icepyx/tests/test_quest_argo.py
@@ -59,8 +59,8 @@ def test_param_setter(argo_quest_instance):
 
     reg_a.params = ["temperature", "salinity"]
 
-    exp = ["temperature", "salinity"]
-    assert reg_a.params == exp
+    exp = {"temperature", "salinity"}
+    assert set(reg_a.params) == exp
 
 
 def test_param_setter_invalid_inputs(argo_quest_instance):


weiji14:
> I'm not sure this is the right decision! Should we instead change the setter so that it preserves order of parameters passed in?

The setter is here:

https://github.com/icesat2py/icepyx/blob/a77b685c945ce16e9b39d06dda1797960b219104/icepyx/quest/dataset_scripts/argo.py#L79-L85

Suggestions:
1. The code is calling `list(set())` twice (once in the setter, and once in [`_validate_parameters`](https://github.com/icesat2py/icepyx/blob/a77b685c945ce16e9b39d06dda1797960b219104/icepyx/quest/dataset_scripts/argo.py#L214-L235)), we could at least remove one of those.
2. The [`_valid_params`](https://github.com/icesat2py/icepyx/blob/a77b685c945ce16e9b39d06dda1797960b219104/icepyx/quest/dataset_scripts/argo.py#L138-L211) list could be turned into a set for faster membership tests, see https://docs.astral.sh/ruff/rules/literal-membership.

mfisher87:
I'm up for doing this refactor! I will assume going forward that order doesn't matter because the setter has been historically throwing away order. What do you think about merging this PR? Since order doesn't matter, this test is now fixed.

weiji14 (approved):
Yep, ok to do the refactor later, and just fix the flaky test for now.

mfisher87:
Right on, merging :)

PR #554, opened 2024-08-13
[WIP] Order data with Harmony instead of ECS

mfisher87:
I'm starting out by adding a Mypy config that has no errors. Then I'm going to add some types to reflect the ECS API requests, and get the type checking passing with that. Then I'll migrate to a new set of Harmony query types.


mfisher87:
I'm going to break this up in smaller pieces.

PR #555, opened 2024-08-13
Keep GitHub Actions up to date with GitHub's Dependabot

cclauss:
Fixes software supply chain safety warnings like at the bottom right of
https://github.com/icesat2py/icepyx/actions/runs/10361169104

* [Keeping your actions up to date with Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)
* [Configuration options for the dependabot.yml file - package-ecosystem](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# Keep GitHub Actions up to date with GitHub's Dependabot...
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    groups:
+      github-actions:
+        patterns:
+          - "*"  # Group all Actions updates into a single larger pull request
+    schedule:
+      interval: monthly


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    groups:
+      github-actions:
+        patterns:
+          - "*"  # Group all Actions updates into a single larger pull request
+    schedule:
+      interval: weekly
```
mfisher87, line 13: Does this seem like an appropriate periodicity for this team's availability? I tend to pick monthly for my projects because I have a lot, but that's just me!

mfisher87 (approved):
:100: Great thinking!

.github/dependabot.yml
cclauss: I usually keep GHAs at the major version so they change VERY infrequently.  If you keep them at micro versions or commit hash versions then they may change more frequently.

cclauss, line 13:
```suggestion
      interval: monthly
```

weiji14: Yes, monthly would be preferable. I'm not a big fan of having the commit history cluttered with dependabot updates, so less is better.

cclauss:
Generated
* #568

Issue #558, opened 2024-08-13
Travis failing on all PRs due to EDL auth error


mfisher87:
```
FAILED icepyx/tests/test_auth.py::test_get_session - icepyx.core.auth.AuthenticationError: Earthdata authentication failed. Check output for error message

FAILED icepyx/tests/test_auth.py::test_get_s3login_credentials - icepyx.core.auth.AuthenticationError: Earthdata authentication failed. Check output for error message

FAILED icepyx/tests/test_auth.py::test_login_function - icepyx.core.auth.AuthenticationError: Earthdata authentication failed. Check output for error message
```

https://app.travis-ci.com/github/icesat2py/icepyx/builds/271856777

mfisher87:
@JessicaS11 @weiji14 just realized this is blocking merge of all PRs. :scream:

I'm trying to figure out what's going on and I don't see any change in the code that corresponds with the start of the failures. Is it possible that a credential was changed and Travis secrets weren't updated?

JessicaS11:
The place I've noticed the auth tests failing on Travis CI tests is the PRs submitted from forks. This is expected behavior, as noted in #272.

mfisher87:
You're so right. I was sure I'd also seen it on one of my PRs which was not from a fork, but now that I'm looking again, I was clearly wrong! I'm going to pin #272 to keep it more visible!

Issue #547, opened 2024-08-10
Consolidate CI from Travis -> GitHub Actions


mfisher87:
I'm not very familiar with Travis, but I'm pretty used to Actions and have tools (e.g. https://github.com/nektos/act) and knowledge I can't apply to Travis. I think it'd be better for contributors to only have to know one or the other. Thoughts?

JessicaS11:
I'm game to transition from Travis to GitHub actions. @mfisher87, I think you partially did this in setting up the v2 branch? Are you game to implement this change, with the manual trigger option we talked about for enabling tests needing auth to run (with approval) for PRs from forks?

I will add the task of setting up the necessary secrets for early next week.

mfisher87:
I'm on it :) But I think I need a refresher on the manual trigger thing. Where'd we talk about that? :thinking: :eyes: 

The v2 branch currently includes running the tests that don't require authentication in GitHub Actions, so I can open a PR with that workflow, and also add an "integration test" workflow that does require the creds.

PR #559, opened 2024-08-13
`test_auth.py`: Skip three pytests that fail on Travis CI

cclauss:
Travis CI limits outbound network requests.

https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

### DIFF
--- a/icepyx/tests/test_auth.py
+++ b/icepyx/tests/test_auth.py
@@ -8,7 +8,9 @@ import earthaccess
 from icepyx.core.auth import EarthdataAuthMixin
 
 # Are we running on Travis CI?
-TRAVIS = os.environ.get("TRAVIS") == "true"
+skip_if_travis_ci = pytest.mark.skipif(
+    os.environ.get("TRAVIS") == "true", reason="Skipping this test on Travis CI."
+)
 
 
 @pytest.fixture()
@@ -21,13 +23,13 @@ def auth_instance():
 
 
 # Test that .session creates a session
-@pytest.mark.skipif(TRAVIS, reason="Skipping this test on Travis CI.")
+@skip_if_travis_ci
 def test_get_session(auth_instance):
     assert isinstance(auth_instance.session, requests.sessions.Session)
 
 
 # Test that .s3login_credentials creates a dict with the correct keys
-@pytest.mark.skipif(TRAVIS, reason="Skipping this test on Travis CI.")
+@skip_if_travis_ci
 def test_get_s3login_credentials(auth_instance):
     assert isinstance(auth_instance.s3login_credentials, dict)
     expected_keys = set(
@@ -37,7 +39,7 @@ def test_get_s3login_credentials(auth_instance):
 
 
 # Test that earthdata_login generates an auth object
-@pytest.mark.skipif(TRAVIS, reason="Skipping this test on Travis CI.")
+@skip_if_travis_ci
 def test_login_function(auth_instance):
     assert isinstance(auth_instance.auth, earthaccess.auth.Auth)
     assert auth_instance.auth.authenticated
--- a/icepyx/tests/test_auth.py
+++ b/icepyx/tests/test_auth.py
@@ -7,6 +7,9 @@ import earthaccess
 
 from icepyx.core.auth import EarthdataAuthMixin
 
+# Are we running on Travis CI?
+TRAVIS = os.environ.get("TRAVIS") == "true"
+
 
 @pytest.fixture()
 def auth_instance():
@@ -18,16 +21,14 @@ def auth_instance():
 
 
 # Test that .session creates a session
-@pytest.mark.skipif(os.getenv("TRAVIS"), reason="Skipping this test on Travis CI.")
+@pytest.mark.skipif(TRAVIS, reason="Skipping this test on Travis CI.")
 def test_get_session(auth_instance):
     print(os.environ)
     assert isinstance(auth_instance.session, requests.sessions.Session)
 
 
 # Test that .s3login_credentials creates a dict with the correct keys
-@pytest.mark.skipif(
-    os.environ.get("TRAVIS") == "true", reason="Skipping this test on Travis CI."
-)
+@pytest.mark.skipif(TRAVIS, reason="Skipping this test on Travis CI.")
 def test_get_s3login_credentials(auth_instance):
     assert isinstance(auth_instance.s3login_credentials, dict)
     expected_keys = set(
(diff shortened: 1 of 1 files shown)


mfisher87:
>Travis CI limits outbound network requests.

Any idea why this just started happening?

We are currently separating out some tests that require EDL in the Travis config into "basic tests" and "behind Earthdata" tests. I feel we should only have one way of skipping Travis - either using Pytest and checking envvars, or using Travis config YML.

If the problem is Travis limiting outbound network requests, I'd expect `pytest icepyx/tests/test_behind_NSIDC_API_login.py` to also fail in Travis after we get past these tests. Should we just move off Travis so we can still run some of these tests (IMO they are integration tests, but that's another conversation ;) ) #547

icepyx/tests/test_auth.py
```diff
@@ -18,15 +18,15 @@ def auth_instance():
...
+@pytest.mark.skipif(os.getenv("TRAVIS"), reason="Skipping this test on Travis CI.")
 def test_get_session(auth_instance):
     print(os.environ)
     assert isinstance(auth_instance.session, requests.sessions.Session)
 
 
 # Test that .s3login_credentials creates a dict with the correct keys
 @pytest.mark.skipif(
-    os.environ.get("TRAVIS") == "true", "Skipping this test on Travis CI."
+    os.environ.get("TRAVIS") == "true", reason="Skipping this test on Travis CI."
```
mfisher87, line 29: Nitpick: Let's use either `os.environ.get` or `os.getenv` in the same way for each mark.

cclauss: 👍 Once I get it to work, I will do it the same way in all three places.

icepyx/tests/test_auth.py
```diff
@@ -1,10 +1,15 @@
...
+
 import pytest
 import requests
 
 import earthaccess
 
 from icepyx.core.auth import EarthdataAuthMixin
 
+# Are we running on Travis CI?
+TRAVIS = os.environ.get("TRAVIS") == "true"
```
mfisher87, line 11: What do you think of:

```python
skip_if_travis_ci = pytest.mark.skipif(TRAVIS, reason="Skipping this test on Travis CI.")


@skip_if_travis_ci
def test_something:
    ...
```

cclauss: I have never seen that before.  Does it work?

icepyx/tests/test_auth.py
mfisher87: Sorry if I'm getting ahead of things here :)

icepyx/tests/test_auth.py
cclauss:
```suggestion
skip_if_travis_ci = pytest.mark.skipif(
    os.environ.get("TRAVIS") == "true", reason="Skipping this test on Travis CI."
)
```

icepyx/tests/test_auth.py
```diff
@@ -16,11 +21,13 @@ def auth_instance():
 
 
 # Test that .session creates a session
+@pytest.mark.skipif(TRAVIS, reason="Skipping this test on Travis CI.")
```
cclauss, line 24:
```suggestion
@skip_if_travis_ci
```

icepyx/tests/test_auth.py
```diff
@@ -16,11 +21,13 @@ def auth_instance():
...
 def test_get_session(auth_instance):
     assert isinstance(auth_instance.session, requests.sessions.Session)
 
 
 # Test that .s3login_credentials creates a dict with the correct keys
+@pytest.mark.skipif(TRAVIS, reason="Skipping this test on Travis CI.")
```
cclauss, line 30:
```suggestion
@skip_if_travis_ci
```

icepyx/tests/test_auth.py
```diff
@@ -30,6 +37,7 @@ def test_get_s3login_credentials(auth_instance):
 
 
 # Test that earthdata_login generates an auth object
+@pytest.mark.skipif(TRAVIS, reason="Skipping this test on Travis CI.")
```
cclauss, line 40:
```suggestion
@skip_if_travis_ci
```

icepyx/tests/test_auth.py
```diff
@@ -1,10 +1,17 @@
+import os
+
 import pytest
 import requests
 
 import earthaccess
 
 from icepyx.core.auth import EarthdataAuthMixin
 
+# Are we running on Travis CI?
```
cclauss, line 10:
```suggestion
```

mfisher87:
> We are currently separating out some tests that require EDL in the Travis config into "basic tests" and "behind Earthdata" tests.
> 
> https://github.com/icesat2py/icepyx/blob/e0e1738737db77714e604035242da32bcd263ad2/.travis.yml#L41-L44C11
> 
> I feel we should only have one way of skipping Travis - either using Pytest and checking envvars, or using Travis config YML.
> 
> If the problem is Travis limiting outbound network requests, I'd expect `pytest icepyx/tests/test_behind_NSIDC_API_login.py` to also fail in Travis after we get past these tests. Should we just move off Travis so we can still run some of these tests (IMO they are integration tests, but that's another conversation ;) ) #547

Hey @cclauss what do you think of this? I'm concerned that when we merge, the "behind Earthdata" tests may fail!

cclauss:
I think we should merge this as is because it solves the current problem and that should get automated tests back to ✅.

It has been a long time since I worked with Travis CI so I cannot guarantee if/how they are limiting outbound network requests.  I would certainly support dropping Travis in favor of GHAs but that might be a bit of work given Conda, etc.

Perhaps `behind_NSIDC` will work just fine on Travis.

mfisher87:
I think I can transition us to GitHub actions pretty quickly, we'll just need the credentials set up! cc @JessicaS11

But we need to wait to merge until the release process is over.

mfisher87:
Since I'm new around here I'm not comfortable reducing test coverage without someone else giving the OK :) @weiji14 thoughts on #558 ?

cclauss:
I am completely lost on this maneuver.

https://github.com/icesat2py/icepyx/pull/559#issuecomment-2287152798
> I think we should merge this as is because it solves the current problem and that should get automated tests back to ✅.

This PR allowed Travis CI to pass.  My three other PRs all pass except for the failing Travis CI tests.  Why can't we get tests back to passing _before_ we break things further?

As I stated above, either get rid of Travis in favor of GitHub Actions or get it to pass so that other PRs are not failing.

mfisher87:
The problem isn't that Travis is failing on all PRs, rather that it's failing on PRs from forks only. Disabling those tests for Travis entirely will reduce the coverage for non-fork PRs, including tests running on `main`. #563 addresses this issue by moving these tests from the phase which runs for PRs to a phase which doesn't run for PRs. Does that seem reasonable to you?

PR #561, opened 2024-08-13
version 1.2.0 release

JessicaS11:


JessicaS11:
My only mild concern is #562.

weiji14:
> My only mild concern is #562.

So technically, the TestPyPI action should be ok when triggered on release because it will be pushing a `1.2.0` tag that doesn't exist on TestPyPI yet :crossed_fingers:

weiji14 (approved):
:rocket:

JessicaS11:
> So technically, the Publish to PyPI action should be ok when triggered on release because it will be pushing a 1.2.0 tag that doesn't exist on TestPyPI yet 🤞

Glad to have this confirmed (that's what I thought). I just have a penchant for ending up with the cases where what should happen isn't what happens... Thanks!

weiji14:
> > So technically, the Publish to PyPI action should be ok when triggered on release because it will be pushing a 1.2.0 tag that doesn't exist on TestPyPI yet 🤞
> 
> Glad to have this confirmed (that's what I thought). I just have a penchant for ending up with the cases where what should happen isn't what happens... Thanks!

Yes, Murphy's law tends to come up during releases (and you and I both know how many times release dramas have happened before), so always good to check!

PR #563, opened 2024-08-14
Avoid running tests that require creds on PRs

### LINKED ISSUE #272: PRs from forks fail Travis CI with `Earthdata authentication failed`

mfisher87:
Resolves #272 

Move test_auth.py into second stage which does not run on PRs

A better fix would be to control what runs based on whether the PR is from a fork. We can do that with GitHub Actions. See: #547 

I tested this by forking and opening a PR against this branch, and its CI passes: https://github.com/icesat2py/icepyx/pull/564

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ stages:
 jobs:
   include:
     - stage: basic tests
-      script: pytest icepyx/ --verbose --cov app --ignore icepyx/tests/test_behind_NSIDC_API_login.py
+      script: pytest icepyx/ --verbose --cov app --ignore=icepyx/tests/test_behind_NSIDC_API_login.py --ignore=icepyx/tests/test_auth.py
       # includes an integrity check of the uploader as recommended by CodeCov
       after_success:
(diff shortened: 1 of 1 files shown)


JessicaS11 (approved):
This seems like a reasonable short-term solution.

PR #565, opened 2024-08-14
Move dependencies from requirements files into project metadata

### LINKED ISSUE #552: Move `requirements*.txt` files into `pyproject.toml`?
I think it's more conventional to express these as extras groups in the project metadata. Then they can be installed as desired and in combination, e.g. with `pip install --editable .[dev,docs]`

mfisher87:
Resolves #552

I hope you don't mind, I took a pass at the contributing doc and made some formatting fixes and re-organized things a little bit in a way that I thought would make it more accessible. Would love some input!


pyproject.toml
```diff
@@ -5,7 +5,7 @@ license = {file = "LICENSE"}
 readme = "README.rst"
 
 requires-python = "~=3.7"
```
weiji14, line 7: Continuing from https://github.com/icesat2py/icepyx/pull/539#discussion_r1717852390, wanted to check if this `~=` pin was intentional? Just noticed it while updating the deps at https://github.com/conda-forge/icepyx-feedstock/pull/18

I know Python 4.0 won't be coming anytime soon, but an upper pin is a big pain in the packaging world (see https://prefix.dev/blog/the_python_packaging_debate), and I'd prefer to use `>=3.7` instead if that's ok

```suggestion
requires-python = ">=3.7"
```

weiji14 (commented):
> I hope you don't mind, I took a pass at the contributing doc and made some formatting fixes and re-organized things a little bit in a way that I thought would make it more accessible. Would love some input!

In this case, the documentation changes (which are quite substantial) should probably be a separate PR... One reason is so that we have 2 changelog entries, one under the the 'Documentation' section and one under 'Maintenance'.

mfisher87:
@weiji14 Some of the documentation changes are tightly coupled with this PR. I'm assuming you want those to stay, but the others to be in a different PR?

mfisher87:
:weary: Two hours ago, Ruff enabled formatting and checking of Jupyter Notebooks by default. New PR coming for that.

JessicaS11 (changes requested):
I haven't reviewed it line-by-line yet, but see my comment in #552.

mfisher87:
> > Yes, if the conda environment has pip installed! How are you currently installing requirements.txt into a conda environment?
> 
> This is a deal breaker for me. I want to be able to create my environment with all potential dependencies and only have to `pip install icepyx -e` (emphasis on the editable part). I've had enough headaches to last my lifetime with pip installing too many packages after creating a skeleton new conda environment. I'm not 100% closed off on this idea, but I need to do a lot more reading and learning before I'm comfortable going ahead with it.

I'm not sure what changes you're requesting here or what the dealbreaker is, could you provide some more info? We're still doing an editable install with pip in this proposal like in the current documentation -- having pip in the conda environment is not a new requirement. The main difference is representing the optional/extra dependencies in a [standard](https://peps.python.org/pep-0621/#dependencies-optional-dependencies) way in project metadata. The proposed workflow here will work the same with a python venv, with conda, or with another tool to manage the environment.

I think what you may be remembering is pip's unfortunate history before it was capable of dependency resolution. Back then, it would install the packages you requested in order, and if dependencies conflicted, it would choose the last one and blow away the needs of previous packages! But now it does truly resolve dependencies: https://pip.pypa.io/en/stable/topics/dependency-resolution/

PR #566, opened 2024-08-15
Tweak format and wording of contributor doc

mfisher87:
Split off from #565. Thank goodness for cherry-pick :)


doc/source/contributing/how_to_contribute.rst
```diff
@@ -3,62 +3,70 @@
...
 
 Contributing for the first time
 -------------------------------
 1. If you don't have one, sign up for a GitHub account (visit https://github.com/ and ‘sign up for GitHub account’).
 
 2. Clone the icepyx repo: Open a terminal window.
-Navigate to the folder on your computer where you want to store icepyx.
-For example,
+   Navigate to the folder on your computer where you want to store icepyx.
+   For example,
```
mfisher87, line 17: This text and code blocks, because they weren't indented, were not rendered as "part of" the numbered list element. Indenting them puts them inside.

JessicaS11: thanks for the explanation and cleaning up this formatting!

doc/source/contributing/how_to_contribute.rst
```diff
@@ -94,10 +102,10 @@ environment (from the home level of your local copy of the icepyx repo) with
 
     mamba env create --name icepyx-env --channel conda-forge -f requirements-dev.txt -f requirements.txt
```
JessicaS11, line 103: Any idea how easy it is to add the interactive "copy" button to the rendered version of this? Definitely not something we should hold up this PR for, but this caught my eye during review as a spot that would be nice to have.

JessicaS11 (approved):
Looks great - thanks Matt!

doc/source/contributing/how_to_contribute.rst
mfisher87: I think that's a pretty easy change with a plugin: https://sphinx-copybutton.readthedocs.io/en/latest/

mfisher87: #587

PR #570, opened 2024-08-18
Add ruff rules C4 to check comprehensions

cclauss:
https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4

% `ruff check --select=C4 --statistics | sort -k2`
```
1	C401	[*] unnecessary-generator-set
1	C405	[*] unnecessary-literal-set
2	C408	[*] unnecessary-collection-call
1	C413	[*] unnecessary-call-around-sorted
1	C416	[*] unnecessary-comprehension
1	C419	[*] unnecessary-comprehension-in-call
```
% `ruff check --select=C4 --fix --unsafe-fixes`
```
Found 7 errors (7 fixed, 0 remaining).
```
% `ruff rule C408`
# unnecessary-collection-call (C408)

Derived from the **flake8-comprehensions** linter.

Fix is always available.

## What it does
Checks for unnecessary `dict`, `list` or `tuple` calls that can be
rewritten as empty literals.

## Why is this bad?
It's unnecessary to call, e.g., `dict()` as opposed to using an empty
literal (`{}`). The former is slower because the name `dict` must be
looked up in the global scope in case it has been rebound.

## Examples
```python
dict()
dict(a=1, b=2)
list()
tuple()
```

Use instead:
```python
{}
{"a": 1, "b": 2}
[]
()
```

## Fix safety
This rule's fix is marked as unsafe, as it may occasionally drop comments
when rewriting the call. In most cases, though, comments will be preserved.

## Options
- `lint.flake8-comprehensions.allow-dict-calls-with-keyword-arguments`

### DIFF
--- a/doc/sphinxext/announce.py
+++ b/doc/sphinxext/announce.py
@@ -125,7 +125,7 @@ def get_authors(revision_range):
     #     pre.discard("Homu")
 
     # Append '+' to new authors.
-    authors = [s + " +" for s in cur - pre] + [s for s in cur & pre]
+    authors = [s + " +" for s in cur - pre] + list(cur & pre)
     authors.sort()
     return authors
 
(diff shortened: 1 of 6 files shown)


mfisher87 (approved):
Nice!

PR #571, opened 2024-08-19
Add ruff rules UP for pyupgrade

cclauss:
https://docs.astral.sh/ruff/rules/#pyupgrade-up

% `ruff check --select=UP --statistics | sort -k2`
```
 2	UP008	[*] super-call-with-parameters
 1	UP009	[*] utf8-encoding-declaration
 1	UP015	[*] redundant-open-modes
 2	UP027	[*] unpacked-list-comprehension
19	UP030	[*] format-literals
17	UP032	[*] f-string
```
% `ruff check --select=UP --fix --unsafe-fixes`
```
Found 28 errors (21 fixed, 7 remaining).
```
% `ruff rule UP032`
# f-string (UP032)

Derived from the **pyupgrade** linter.

Fix is sometimes available.

## What it does
Checks for `str.format` calls that can be replaced with f-strings.

## Why is this bad?
f-strings are more readable and generally preferred over `str.format`
calls.

## Example
```python
"{}".format(foo)
```

Use instead:
```python
f"{foo}"
```

## References
- [Python documentation: f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings)

### DIFF
--- a/icepyx/core/APIformatting.py
+++ b/icepyx/core/APIformatting.py
@@ -84,11 +84,11 @@ def _fmt_readable_granules(dset, **kwds):
                 # use single character wildcards "?" for date strings
                 # and ATLAS granule region number
                 if dset in ("ATL07", "ATL10", "ATL20", "ATL21"):
-                    granule_name = "{0}-??_{1}_{2}{3}??_*".format(dset, 14 * "?", t, c)
+                    granule_name = "{}-??_{}_{}{}??_*".format(dset, 14 * "?", t, c)
                 elif dset in ("ATL11",):
-                    granule_name = "{0}_{1}??_*".format(dset, t)
+                    granule_name = f"{dset}_{t}??_*"
                 else:
-                    granule_name = "{0}_{1}_{2}{3}??_*".format(dset, 14 * "?", t, c)
+                    granule_name = "{}_{}_{}{}??_*".format(dset, 14 * "?", t, c)
                 # append the granule
                 readable_granule_list.append(granule_name)
     # extend with explicitly named files (full or partial)
--- a/icepyx/core/granules.py
+++ b/icepyx/core/granules.py
@@ -498,7 +498,7 @@ class Granules(EarthdataAuthMixin):
         # (could create confusion depending on whether download was interrupted or kernel restarted)
         order_fn = ".order_restart"
         if os.path.exists(order_fn):
-            with open(order_fn, "r") as fid:
+            with open(order_fn) as fid:
                 order_dat = json.load(fid)
                 self.orderIDs = order_dat["orderIDs"]
 
--- a/icepyx/core/query.py
+++ b/icepyx/core/query.py
@@ -136,12 +136,7 @@ class GenQuery:
             self._temporal = tp.Temporal(date_range, start_time, end_time)
 
     def __str__(self):
-        str = "Extent type: {0} \nCoordinates: {1}\nDate range: ({2}, {3})".format(
-            self._spatial._ext_type,
-            self._spatial._spatial_ext,
-            self._temporal._start,
-            self._temporal._end,
-        )
+        str = f"Extent type: {self._spatial._ext_type} \nCoordinates: {self._spatial._spatial_ext}\nDate range: ({self._temporal._start}, {self._temporal._end})"
         return str
 
     # ----------------------------------------------------------------------
@@ -442,9 +437,7 @@ class Query(GenQuery, EarthdataAuthMixin):
     # Properties
 
     def __str__(self):
-        str = "Product {2} v{3}\n{0}\nDate range {1}".format(
-            self.spatial_extent, self.dates, self.product, self.product_version
-        )
+        str = f"Product {self.product} v{self.product_version}\n{self.spatial_extent}\nDate range {self.dates}"
         return str
 
     @property
--- a/icepyx/quest/dataset_scripts/argo.py
+++ b/icepyx/quest/dataset_scripts/argo.py
@@ -56,9 +56,9 @@ class Argo(DataSet):
             df = "\n" + str(self.argodata.head())
         s = (
             "---Argo---\n"
-            "Parameters: {0}\n"
-            "Pressure range: {1}\n"
-            "Dataframe head: {2}".format(self.params, prange, df)
+            f"Parameters: {self.params}\n"
+            f"Pressure range: {prange}\n"
+            f"Dataframe head: {df}"
         )
 
         return s
@@ -115,7 +115,7 @@ class Argo(DataSet):
         coordinates_array = np.asarray(gdf.geometry[0].exterior.coords)
         x = ""
         for i in coordinates_array:
-            coord = "[{0},{1}]".format(i[0], i[1])
+            coord = f"[{i[0]},{i[1]}]"
             if x == "":
                 x = coord
             else:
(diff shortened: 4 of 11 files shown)


icepyx/core/APIformatting.py
```diff
@@ -84,11 +84,11 @@ def _fmt_readable_granules(dset, **kwds):
...
                 # and ATLAS granule region number
                 if dset in ("ATL07", "ATL10", "ATL20", "ATL21"):
-                    granule_name = "{0}-??_{1}_{2}{3}??_*".format(dset, 14 * "?", t, c)
+                    granule_name = "{}-??_{}_{}{}??_*".format(dset, 14 * "?", t, c)
                 elif dset in ("ATL11",):
-                    granule_name = "{0}_{1}??_*".format(dset, t)
+                    granule_name = f"{dset}_{t}??_*"
                 else:
-                    granule_name = "{0}_{1}_{2}{3}??_*".format(dset, 14 * "?", t, c)
+                    granule_name = "{}_{}_{}{}??_*".format(dset, 14 * "?", t, c)
```
mfisher87, line 91: Why was only 1/3 of these converted to f-string? :thinking:

icepyx/core/query.py
```diff
@@ -136,12 +136,7 @@ def __init__(
             self._temporal = tp.Temporal(date_range, start_time, end_time)
 
     def __str__(self):
-        str = "Extent type: {0} \nCoordinates: {1}\nDate range: ({2}, {3})".format(
-            self._spatial._ext_type,
-            self._spatial._spatial_ext,
-            self._temporal._start,
-            self._temporal._end,
-        )
+        str = f"Extent type: {self._spatial._ext_type} \nCoordinates: {self._spatial._spatial_ext}\nDate range: ({self._temporal._start}, {self._temporal._end})"
```
mfisher87, line 139: I personally find this a bit less readable. What about:
```suggestion
        str = (
            f"Extent type: {self._spatial._ext_type}"
            f"\nCoordinates: {self._spatial._spatial_ext}"
            f"\nDate range: ({self._temporal._start}, {self._temporal._end})"
        )
```

icepyx/core/query.py
```diff
@@ -442,9 +437,7 @@ def __init__(
     # Properties
 
     def __str__(self):
-        str = "Product {2} v{3}\n{0}\nDate range {1}".format(
-            self.spatial_extent, self.dates, self.product, self.product_version
-        )
+        str = f"Product {self.product} v{self.product_version}\n{self.spatial_extent}\nDate range {self.dates}"
```
mfisher87, line 440: Similar comment about readability!

icepyx/quest/dataset_scripts/argo.py
```diff
@@ -56,9 +56,9 @@ def __str__(self):
             df = "\n" + str(self.argodata.head())
         s = (
             "---Argo---\n"
-            "Parameters: {0}\n"
-            "Pressure range: {1}\n"
-            "Dataframe head: {2}".format(self.params, prange, df)
+            f"Parameters: {self.params}\n"
+            f"Pressure range: {prange}\n"
+            f"Dataframe head: {df}"
```
mfisher87, line 61: So much more readable. The implicit concatenation combined with `.format()` just weirds me out :)

icepyx/quest/dataset_scripts/argo.py
```diff
@@ -363,7 +361,7 @@ def _download_profile(
 
         # Consider any status other than 2xx an error
         if not resp.status_code // 100 == 2:
```
mfisher87, line 363: Thinking out loud, this just jumped out at me, but it's not part of the PR. I'd like to try to use `resp.ok` instead of this pattern eventually.

icepyx/core/query.py
cclauss: Funny that a repo that fails to set a ruff `line-length` wants to wrap lines.

icepyx/quest/dataset_scripts/argo.py
cclauss: This `if` covers 200, 201, 202, 299... -- https://requests.readthedocs.io/en/latest/api/#status-code-lookup

mfisher87: Oh, yeah, good point! Thanks.

icepyx/core/APIformatting.py
cclauss: If the conversion would make the line more difficult to read then the conversion will not be made.

icepyx/core/query.py
cclauss: `E501` is the ruff rule that enables `line-length` for the linter.  It is universally ignored in this repo.

icepyx/core/APIformatting.py
mfisher87: Personally, I think it'd be easier to read this one as an f-string! Just my opinion though :) I'm not requesting any changes though, just curious about why it did what it did.

```python
"{}_{}_{}{}??_*".format(dset, 14 * "?", t, c)
f"{dset}_{14 * '?'}_{t}{c}??_*"
f"{dset}_??????????????_{t}{c}??_*"  # I think I like this best haha
```

icepyx/core/query.py
mfisher87: That's not quite a complete description of the situation. `E501` is the "line too long rule", true. However, Ruff is a linter and formatter, and uses `line-length` for both, but `E501` only affects linting behavior. The reason why E501 was disabled is explained in the config file in a co-located comment. Please try for yourself! If you add a long line, the formatter will work as described above! This is the formatting that was applied for me:

```
-print("This is a very very very very very very very very very very very long line" + "And I want it to be formatted")
+print(
+    "This is a very very very very very very very very very very very long line"
+    + "And I want it to be formatted"
+)
```

icepyx/core/APIformatting.py
JessicaS11: > If the conversion would make the line more difficult to read then the conversion will not be made.

Any idea what these criteria are? I'm with Matt that I think the f-strings are clearer.

icepyx/core/granules.py
```diff
@@ -498,7 +498,7 @@ def download(self, verbose, path, restart=False):
         # (could create confusion depending on whether download was interrupted or kernel restarted)
         order_fn = ".order_restart"
         if os.path.exists(order_fn):
-            with open(order_fn, "r") as fid:
+            with open(order_fn) as fid:
```
JessicaS11, line 501: Why is being explicit not the preference here (even if "r" is default)?

mfisher87: I think it's considered well-known or intuitive that read mode is the default? I don't think I fully agree with this. "explicit is better than implicit" seems like a good enough reason to disable this rule, but this is pretty low-stakes IMO, so I don't feel strongly.

The rule docs (https://docs.astral.sh/ruff/rules/redundant-open-modes/) say:

> ## Why is this bad?
> Redundant `open` mode parameters are unnecessary and should be removed to avoid confusion.

Whenever I see "it's unnecessary" used as a justification, it's a yellow-flag in my mind. We do lots of things that are unnecessary to make our work more accessible or provide other value.

mfisher87:
@cclauss  I'm still interested in this PR, are you open to continuing?

mfisher87:
Related conversation happening here: https://github.com/icesat2py/icepyx/pull/572#issuecomment-2301441004

PR #572, opened 2024-08-19
Add ruff rules SIM to simplify Python code

cclauss:
https://docs.astral.sh/ruff/rules/#flake8-simplify-sim

% `ruff check --select=SIM --statistics`
```
26	SIM118	[*] in-dict-keys
 5	SIM105	[ ] suppressible-exception
 5	SIM108	[*] if-else-block-instead-of-if-exp
 2	SIM114	[*] if-with-same-arms
 2	SIM201	[*] negate-equal-op
 1	SIM101	[*] duplicate-isinstance-call
 1	SIM103	[*] needless-bool
```
% `ruff rule SIM118`
# in-dict-keys (SIM118)

Derived from the **flake8-simplify** linter.

Fix is always available.

## What it does
Checks for key-existence checks against `dict.keys()` calls.

## Why is this bad?
When checking for the existence of a key in a given dictionary, using
`key in dict` is more readable and efficient than `key in dict.keys()`,
while having the same semantics.

## Example
```python
key in foo.keys()
```

Use instead:
```python
key in foo
```

## Fix safety
Given `key in obj.keys()`, `obj` _could_ be a dictionary, or it could be
another type that defines a `.keys()` method. In the latter case, removing
the `.keys()` attribute could lead to a runtime error. The fix is marked
as safe when the type of `obj` is known to be a dictionary; otherwise, it
is marked as unsafe.

## References
- [Python documentation: Mapping Types](https://docs.python.org/3/library/stdtypes.html#mapping-types-dict)

[preview]: https://docs.astral.sh/ruff/preview/

### DIFF
--- a/icepyx/core/APIformatting.py
+++ b/icepyx/core/APIformatting.py
@@ -73,7 +73,7 @@ def _fmt_readable_granules(dset, **kwds):
     # list of readable granule names
     readable_granule_list = []
     # if querying either by 91-day orbital cycle or RGT
-    if "cycles" in kwargs.keys() or "tracks" in kwargs.keys():
+    if "cycles" in kwargs or "tracks" in kwargs:
         # default character wildcards for cycles and tracks
         kwargs.setdefault("cycles", ["??"])
         kwargs.setdefault("tracks", ["????"])
@@ -110,7 +110,7 @@ def _fmt_var_subset_list(vdict):
     """
 
     subcover = ""
-    for vn in vdict.keys():
+    for vn in vdict:
         vpaths = vdict[vn]
         for vpath in vpaths:
             subcover += "/" + vpath + ","
@@ -300,7 +300,7 @@ class Parameters:
 
         val_list = list({val for lis in self.poss_keys.values() for val in lis})
 
-        for key in self.fmted_keys.keys():
+        for key in self.fmted_keys:
             assert key in val_list, (
                 "An invalid key (" + key + ") was passed. Please remove it using `del`"
             )
@@ -317,7 +317,7 @@ class Parameters:
         ), "You cannot call this function for your parameter type"
         reqkeys = self.poss_keys[self._reqtype]
 
-        if all(keys in self.fmted_keys.keys() for keys in reqkeys):
+        if all(keys in self.fmted_keys for keys in reqkeys):
             assert all(
                 self.fmted_keys.get(key, -9999) != -9999 for key in reqkeys
             ), "One of your formatted parameters is missing a value"
@@ -337,7 +337,7 @@ class Parameters:
         spatial_keys = self.poss_keys["spatial"]
 
         # not the most robust check, but better than nothing...
-        if any(keys in self._fmted_keys.keys() for keys in spatial_keys):
+        if any(keys in self._fmted_keys for keys in spatial_keys):
             assert any(
                 self.fmted_keys.get(key, -9999) != -9999 for key in spatial_keys
             ), "One of your formatted parameters is missing a value"
@@ -410,13 +410,13 @@ class Parameters:
                 opt_keys = self.poss_keys["optional"]
 
                 for key in opt_keys:
-                    if key == "Coverage" and key in kwargs.keys():
+                    if key == "Coverage" and key in kwargs:
                         # DevGoal: make there be an option along the lines of Coverage=default, which will get the default variables for that product without the user having to input is2obj.build_wanted_wanted_var_list as their input value for using the Coverage kwarg
                         self._fmted_keys.update(
                             {key: _fmt_var_subset_list(kwargs[key])}
                         )
                     elif (key == "temporal" or key == "time") and (
-                        "start" in kwargs.keys() and "end" in kwargs.keys()
+                        "start" in kwargs and "end" in kwargs
                     ):
                         self._fmted_keys.update(
                             _fmt_temporal(kwargs["start"], kwargs["end"], key)
--- a/icepyx/core/granules.py
+++ b/icepyx/core/granules.py
@@ -75,10 +75,9 @@ def gran_IDs(grans, ids=False, cycles=False, tracks=False, dates=False, cloud=Fa
         if cloud is True:
             try:
                 for link in gran["links"]:
-                    if link["href"].startswith("s3") and link["href"].endswith(
-                        (".h5", "nc")
-                    ):
-                        gran_s3urls.append(link["href"])
+                    href = link["href"]
+                    if href.startswith("s3") and href.endswith((".h5", "nc")):
+                        gran_s3urls.append(href)
             except KeyError:
                 pass
 
--- a/icepyx/core/query.py
+++ b/icepyx/core/query.py
@@ -126,7 +126,7 @@ class GenQuery:
         **kwargs,
     ):
         # validate & init spatial extent
-        if "xdateline" in kwargs.keys():
+        if "xdateline" in kwargs:
             self._spatial = spat.Spatial(spatial_extent, xdateline=kwargs["xdateline"])
         else:
             self._spatial = spat.Spatial(spatial_extent)
@@ -637,7 +637,7 @@ class Query(GenQuery, EarthdataAuthMixin):
         else:
             # If the user has supplied a subset list of variables, append the
             # icepyx required variables to the Coverage dict
-            if "Coverage" in kwargs.keys():
+            if "Coverage" in kwargs:
                 var_list = [
                     "orbit_info/sc_orient",
                     "orbit_info/sc_orient_time",
@@ -653,7 +653,7 @@ class Query(GenQuery, EarthdataAuthMixin):
                 ]
                 # Add any variables from var_list to Coverage that are not already included
                 for var in var_list:
-                    if var not in kwargs["Coverage"].keys():
+                    if var not in kwargs["Coverage"]:
                         kwargs["Coverage"][var.split("/")[-1]] = [var]
 
             if self._subsetparams is None:
@@ -741,9 +741,7 @@ class Query(GenQuery, EarthdataAuthMixin):
         <icepyx.core.granules.Granules at [location]>
         """
 
-        if not hasattr(self, "_granules"):
-            self._granules = Granules()
-        elif self._granules is None:
+        if not hasattr(self, "_granules") or self._granules is None:
             self._granules = Granules()
 
         return self._granules
@@ -869,7 +867,7 @@ class Query(GenQuery, EarthdataAuthMixin):
         ]
 
         try:
-            all(key in self._cust_options.keys() for key in keys)
+            all(key in self._cust_options for key in keys)
         except (AttributeError, KeyError):
             self._cust_options = is2ref._get_custom_options(
                 self.session, self.product, self._version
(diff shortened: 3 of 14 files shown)


icepyx/core/granules.py
```diff
@@ -73,16 +74,13 @@ def gran_IDs(grans, ids=False, cycles=False, tracks=False, dates=False, cloud=Fa
         gran_ids.append(producer_granule_id)
 
         if cloud is True:
-            try:
+            with contextlib.suppress(KeyError):
```
mfisher87, line 77: Not a strong opinion, but I feel this is less readable despite having fewer lines.

icepyx/core/query.py
```diff
@@ -741,9 +741,7 @@ def granules(self):
         <icepyx.core.granules.Granules at [location]>
         """
 
-        if not hasattr(self, "_granules"):
-            self._granules = Granules()
-        elif self._granules is None:
+        if not hasattr(self, "_granules") or self._granules is None:
```
mfisher87, line 744: Huge value from this type of simplification; it would be easy to introduce a bug by changing only one branch of the existing conditional and forgetting the other.

icepyx/quest/dataset_scripts/argo.py
```diff
@@ -297,7 +294,7 @@ def search_data(self, params=None, presRange=None, printURL=False) -> str:
         selectionProfiles = resp.json()
 
         # Consider any status other than 2xx an error
-        if not resp.status_code // 100 == 2:
+        if resp.status_code // 100 != 2:
```
mfisher87, line 297: :100:

mfisher87 (changes requested):
Love it! Except, and this is just my personal opinion, but I think the `contextlib.suppress` change is a (thin) obfuscation of the underlying `try`/`except` behavior and reduces readability. I can probably be convinced otherwise, but I've read a lot of Python and had to go to the docs to figure out what it did! [It doesn't seem uncommon to single this rule out for disabling](https://github.com/jupyter/jupyter_core/blob/fa513c1550bbd1ebcc14a4a79eb8c5d95e3e23c9/pyproject.toml#L183).

icepyx/core/granules.py
cclauss: It just takes getting used to it is easier to read once you do.

mfisher87: I think I'd need a bit more convincing on this :) Apologies, since I originally posted, my opinion has gotten a bit stronger. In my view, "taking some getting used to" is a really big cost! Code is read at least one order of magnitude more than it's written, so I care deeply about optimizing for folks who haven't gotten used to it yet.

cclauss:
The suppressible-exception have been suppressed.

mfisher87 (approved):
LGTM! I'll leave some time for discussion before merging.

icepyx/core/APIformatting.py
```diff
@@ -73,7 +73,7 @@ def _fmt_readable_granules(dset, **kwds):
     # list of readable granule names
     readable_granule_list = []
     # if querying either by 91-day orbital cycle or RGT
-    if "cycles" in kwargs.keys() or "tracks" in kwargs.keys():
+    if "cycles" in kwargs or "tracks" in kwargs:
```
JessicaS11, line 76: I can see the argument for using this simplification (thanks for including the notes about it with the PR). However, I am concerned that this has potential to be confusing to folks newer to Python and ultimately make the code less clear/readable to them. It assumes an awareness that Python will automagically include `.keys()` if this is a dict, which is news to me as of this PR. What's the incentive for using this simplification over explicitly including `.keys()`?

mfisher87: I think readability is paramount, and the performance concern is negligible, but want to mention it anyway. Calling `.keys()` is causing some extra overhead to create a [view](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects) of the dictionary's keys before checking.

While I do agree that calling `.keys()` here is more explicit, and [explicit is better than implicit](https://peps.python.org/pep-0020/), "readability counts". Sometimes the most readable thing is the way everyone else is doing it, and I think this is one of those cases. The official Python docs endorse this approach:

>To check whether a single key is in the dictionary, use the [in](https://docs.python.org/3/reference/expressions.html#in) keyword.

https://docs.python.org/3/tutorial/datastructures.html

cclauss: Membership test operations (`item in container`) works for dicts, lists, sets, tuples, and all other classes that implement `.__contains__()`
* https://docs.python.org/3/reference/expressions.html#in
* [`https://docs.python.org/3/reference/datamodel.html#object.__contains__`](https://docs.python.org/3/reference/datamodel.html#object.__contains__)

mfisher87: IMO this is a great use case for ChatGPT ("comprehension", in the "list-comprehension" sense, not the cognition sense, of documentation or community coding convention) so I threw this question at it, and it did kind of lie by misquoting the documentation, but it came pretty close to what I consider a good answer.

https://chatgpt.com/share/de22c27b-3370-4051-9d58-b5eeb37b9e09

mfisher87:
>What I try to avoid is making complex manual changes in a pull request that is supposed to be based on automatic transforms. These make the PR too difficult to review and can introduce bugs that are not based on applying the rules of the tool. I know that the tool and its rules are imperfect but in general their transformations are useful.

:100: Agreed! But - when a rule does a thing that the project team doesn't universally agree on, IMO, that PR is the place to hash it out, and that's what I feel was happening in #571. I'm OK with addressing the concerns in #571 by disabling a rule and leaving a comment explaining why we disabled that rule (e.g. "automatic f-string conversion on this codebase makes some string interpolations harder to read") and opening an issue to track the desired steps. I'm also OK with mixing concerns a little bit (and what I mean by "a little bit" is based on vibes) - a PR that re-enabled that rule and also manually adjusted some strings would be fine, especially if those two things were in two separate commits I could review separately.

I think this PR is ready to merge if @JessicaS11 is now comfortable with our conversation above. If not, I think we should disable the rule which converts `... in foo.keys()` to `... in foo`, merge, and come back to it later.

> [!IMPORTANT]
>
> Going forward, let's please keep discussions about a PR associated with said PR!

icepyx/core/APIformatting.py
JessicaS11: Thanks @cclauss and @mfisher87 for helping me understand this convention!

icepyx/core/variables.py
```diff
@@ -111,11 +108,7 @@ def __init__(
 
     @property
     def path(self):
-        if self._path:
-            path = self._path
-        else:
-            path = None
-        return path
+        return self._path or None
```
JessicaS11, line 111: Is this another case where there are "hidden" assumptions? Why doesn't this simplify to `return self._path if self._path else None`? To me, `return self._path or None` is like saying "yes" to "do you want an apple or an orange?" It's not obvious which one I should give you (and I can't give you both).

JessicaS11:
>I think this PR is ready to merge if @JessicaS11 is now comfortable with our conversation above. If not, I think we should disable the rule which converts ... in foo.keys() to ... in foo, merge, and come back to it later.

Agreed. As a newbie to ruff - how do these "extras" fit into the Python formatting ecosystem? Some of these ruff rules are improving PEP compliance (yay!), some are aiming to make code a bit shorter (overall good), and some seem to be enforcing personal preferences that make the code shorter but I would argue less intuitive to read if you don't know the underlying assumptions. As a library that tries to be super-beginner friendly, I am trying to discern where it's acceptable to have a slightly "longer and less optimized" code base that is readable to a broader subset of scientific Python contributors. If I have to guess at what's happening under the hood (e.g. in the `return self._path or None` case), I can only imagine it will be the case for other developers who don't have "software" somewhere in their job title. This takes the ruff rule from "helpful" to "harmful" in my view.

icepyx/core/variables.py
cclauss: There are no assumptions as Truth Value Testing with `or` and `and` is defined quite clearly at the beginning of the Python Standard Library docs before `int`, `list`, `dict`, etc. are defined.

icepyx/core/APIformatting.py
mfisher87: Happy to!

icepyx/core/variables.py
mfisher87: We can simplify this statement in two passes to make it a bit clearer:

```python
if self._path:
    path = self._path
else:
    path = None
return path
```

->

```python
if self._path:
    return self._path
else:
    return None
```

->

```python
return self._path or None
```

The `or` here is doing the lifting of the `else` condition -- if `self._path` is "Falsy", we move along to the next thing in the `or` operation.

cclauss:
```suggestion
        return self._path if self._path else None
```

cclauss:
`return self._path if self._path else None` wins.

mfisher87:
>how do these "extras" fit into the Python formatting ecosystem? Some of these ruff rules are improving PEP compliance (yay!), [...]

They largely come from the flake8 linting ecosystem. Flake8 has a huge number of plugins, and after a while I stopped thinking of it only in terms of PEP8 despite the name :)

> [...] some are aiming to make code a bit shorter (overall good), [...]

I try to be cautious with this, because sometimes shorter is better (more concise) and other times shorter introduces special concepts (like SIM105).

> [...] some seem to be enforcing personal preferences that make the code shorter but I would argue less intuitive to read if you don't know the underlying assumptions

Absolutely. The SIM105 rule is, to me, definitely on this side of the equation. In my experience, there are very few rules like this in the default set, but there are some.

**There's also a really important 4th category**: Things that aren't PEP8, aren't shorter, but are common pitfalls for Python developers. One of the most important rulesets, IMO, is the [flake8-bugbear](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b) set. It detects, for example, attempting to clear environment vars by assigning to `os.environ` (B003). Or using a mutable object (B006) or calling a function (B008) to set a default argument. Things that look like will do what you want, but really are a bug or misleading. Then there's [flake8-unused-arguments](https://docs.astral.sh/ruff/rules/#flake8-unused-arguments-arg) which detects when you write a function with args but you don't use some of them in the function body, falsely advertising to users of the function that all of the args passed in are significant.

There are so many ways to write code that looks right, but isn't.

Type checking, to me, is another vital tool for detecting situations like this :point_up: and I'd like to get in to that space for v2. This is very valuable both for detecting bugs, but also for showing autocompletions in developers IDEs.

> If I have to guess at what's happening under the hood (e.g. in the `return self._path or None` case), I can only imagine it will be the case for other developers who don't have "software" somewhere in their job title. This takes the ruff rule from "helpful" to "harmful" in my view.

The functionality of `or`, I personally feel, is really important for beginner developers to understand; ["short-circuiting"](https://en.wikipedia.org/wiki/Short-circuit_evaluation) is very handy behavior. This is a "Pythonic" pattern that readers will definitely come across again, Ruff or no Ruff. That said, `return self._path if self._path else None` is trivially "less-Pythonic" in my view than `return self._path or None`.

JessicaS11:
@all-contributors
please add @cclauss for maintenance
please add @mfisher87 for bug, code, doc, maintenance, test, infra
please add @jrenrut for bug, code

JessicaS11:
Thanks for this great response! I'm learning so much.

> The functionality of or, I personally feel, is really important for beginner developers to understand; ["short-circuiting"](https://en.wikipedia.org/wiki/Short-circuit_evaluation) is very handy behavior.

I just taught a bunch of folks about this concept, so it was exciting to learn with them. I agree it's a good feature to know about (and @RomiP provided some helpful insight on its ability to make code more efficient by not running unnecessary bits).

JessicaS11 (approved):
Thanks for this PR and all the great discussion!

mfisher87:
@all-contributors
please add @cclauss for review

mfisher87:
@all-contributors
please add @jrenrut for bug, code

PR #573, opened 2024-08-19
Enable currently-ignored linting rules E721, E722, F403

cclauss:
All three of these are discussed on [PEP8](https://peps.python.org/pep-0008) and should not be ignored.

E722 is https://realpython.com/the-most-diabolical-python-antipattern

```diff
[tool.ruff.lint]
ignore = [
-  # TODO: remove ignores below this line
-  # comparison syntax in tests
-  "E721",
-  # bare except
-  "E722",
-  # unable to detect undefined names
-  "F403",
]
```

### DIFF
--- a/icepyx/tests/test_quest.py
+++ b/icepyx/tests/test_quest.py
@@ -26,9 +26,9 @@ def test_add_is2(quest_instance):
 
     obs = quest_instance.datasets
 
-    assert type(obs) == dict
+    assert isinstance(obs, dict)
     assert exp_key in obs.keys()
-    assert type(obs[exp_key]) == exp_type
+    assert isinstance(obs[exp_key], exp_type)
     assert quest_instance.datasets[exp_key].product == prod
 
 
@@ -40,9 +40,9 @@ def test_add_argo(quest_instance):
 
     obs = quest_instance.datasets
 
-    assert type(obs) == dict
+    assert isinstance(obs, dict)
     assert exp_key in obs.keys()
-    assert type(obs[exp_key]) == exp_type
+    assert isinstance(obs[exp_key], exp_type)
     assert set(quest_instance.datasets[exp_key].params) == set(params)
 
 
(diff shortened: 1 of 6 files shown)


icepyx/tests/test_quest.py
```diff
@@ -40,9 +40,9 @@ def test_add_argo(quest_instance):
 
     obs = quest_instance.datasets
 
-    assert type(obs) == dict
+    assert isinstance(obs, dict)
     assert exp_key in obs.keys()
-    assert type(obs[exp_key]) == exp_type
+    assert isinstance(obs[exp_key], exp_type)
```
mfisher87, line 45: Note: this is a test so we have automatic confirmation it's fine, but for anyone not aware, there is a difference in behavior between `type(...) == ...` and `isintance(..., ...)` - `isinstance` evaluates `True` with inheritance:


```python
>>> class Foo(str):
...     ...
... 
>>> type(Foo()) is str
False
>>> type(Foo()) is Foo
True
>>> isinstance(Foo(), str)
True
>>> isinstance(Foo(), Foo)
True
```

mfisher87 (approved):

icepyx/tests/test_quest.py
cclauss: Agreed.  The `==` --> `is` transition would be good enough to silence this ruff rule.  The "comparing types" thing is frowned on in PEP8 and there is a different ruff rule that covers that.

mfisher87: The more I think about it, the more I think we should go back to exact type comparison with `type(...) is ...`. Since this is a test, loosening the comparison enables the underlying code to change more without breaking the test. We an always come back to this. My little mind is just thinking about the "foolish consistency" part of PEP8.

mfisher87 (changes requested):
Apologies for all the mind-changing I'm doing today @cclauss. Could you address this before merge? https://github.com/icesat2py/icepyx/pull/573/files#r1722071874

mfisher87 (approved):
Lovely! Thank you again :bow:

PR #580, opened 2024-08-21
Migrate unit and integration tests from TravisCI to GitHub Actions

### LINKED ISSUE #547: Consolidate CI from Travis -> GitHub Actions

mfisher87:
Closes #547 

This PR migrates the current TravisCI behavior into GitHub Actions.

**New feature**: Unit tests run on multiple versions of Python.

TODO:

- [x] @JessicaS11 add credentials currently being used for TravisCI tests to GitHub repository secrets
- [x] @JessicaS11 make new unit and integration test workflows "required" in GH repository settings
- [x] @JessicaS11 clean up required Travis check  in repository settings 
- [x] @JessicaS11 or @mfisher87 update CI badges on readme (currently point to Travis)

> [!NOTE]
>
> This **does not** attempt to implement an approval process for PRs from forks. We don't have that now, so we're not losing anything by dropping Travis in this PR. This is challenging. I'll open another draft PR with that stuff, and we can hopefully figure it out.
>
> See #588


mfisher87:
I got a bit frustrated trying to get the integration test "secure approval" process working the way I want it to. I'm going to spend some time on the Harmony -> ECS work today to give myself a break. I made a post in Openscapes Slack asking if anyone has ideas what's going wrong in the meantime :) Any tips greatly appreciated!

mfisher87:
I think I'm going to split this PR into two pieces: (1) migrate Travis -> GHA as-is; (2) the approval thing.

(1) alone provides value by reducing the number of unique automation systems we depend on.

.github/workflows/integration_test.yml
```diff
@@ -0,0 +1,40 @@
...
+          fetch-depth: 0
+
+      - uses: "./.github/actions/install-icepyx"
+        with:
+          python-version: "3.12"
+
+      - name: "Run tests"
+        env:
+          EARTHDATA_USERNAME: "${{ secrets.EARTHDATA_USERNAME }}"
+          EARTHDATA_PASSWORD: "${{ secrets.EARTHDATA_PASSWORD }}"
```
mfisher87, line 31: @JessicaS11 these are the secrets we need added to this repository :point_up:

.github/actions/install-icepyx/action.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Install icepyx'
+description: 'Install icepyx, and dev dependencies'
```
JessicaS11, line 2:
```suggestion
description: 'Install icepyx and dev dependencies'
```

.github/workflows/integration_test.yml
```diff
@@ -0,0 +1,40 @@
+name: "Integration test"
+# NOTE: We're just running the tests that require earthdata login here; we
+#       don't distinguish yet between unit and integration tests yet.
```
JessicaS11, line 3:
```suggestion
# NOTE: We're just running the tests that require earthdata login here; we
#       don't distinguish between unit and integration tests yet.
```

Is your thought that we'll eventually transition to unit + integration, so it makes sense to rename to those now?

.github/workflows/integration_test.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - "main"
+      - "development"
```
JessicaS11, line 10: Given these are slightly "expensive" tests since they download files (and running them for every commit could ultimately influence the icepyx metrics), what are our options for restricting how often they run for a given PR?

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,39 @@
...
+      - "development"
+
+
+jobs:
+  test:
+    runs-on: "ubuntu-latest"
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.9", "3.12"]
```
JessicaS11, line 19: Learning moment (and I know there's an open issue too): why these two versions?

.github/workflows/integration_test.yml
mfisher87: Yes, that's my thinking. It's common terminology that is recognizable at a glance to those familiar with it, and also searchable. I'm not attached to this naming, however!

I would like to, if there's time, eventually split our tests up in to "unit" and "integration" directories.

.github/workflows/integration_test.yml
mfisher87: These triggers (`branches:`) would be for every merge, not for every commit on PRs. We can customize the `pull_request:` triggers. By default, they run on `open` and `synchronize` (effectively, when the PR is pushed to, not on every commit). The full list is here: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

I think the best way we can limit these runs is by adding `paths:` filters. That would enable e.g. docs changes to not trigger the expensive tests!

mfisher87: Should we consider this part of this PR, or create an issue to track the work?

.github/workflows/unit_test.yml
mfisher87: I chose the lowest and highest versions of Python that we support. We can test every version we support, but I was trying to balance exhaustiveness with environmental impact. I personally can't think of a problem that would be caught by running our tests for 3.11, but that may just be my ignorance :)

.github/workflows/integration_test.yml
mfisher87: It auto-resolved this when I accepted your suggestion, so I un-resolved. You can mark resolved if you feel that's the case :)

.github/workflows/integration_test.yml
JessicaS11: I added EARTHDATA_PASSWORD and CODECOV_TOKEN to the org/repo secrets. EARTHDATA_USERNAME is included as part of a session fixture within `test_behind_NSIDC_API_login.py`.

.github/workflows/integration_test.yml
JessicaS11: > (effectively, when the PR is pushed to, not on every commit).

Right - forgot this subtlety - but a PR can have a lot of pushes too!

> Should we consider this part of this PR, or create an issue to track the work?

I would definitely consider it part of this PR. I don't want to have a setup running that will trigger so many runs.

.github/workflows/unit_test.yml
JessicaS11: Works for me (and makes complete sense).

```suggestion
        python-version: ["3.9", "3.12"] #NOTE: min and max Python versions supported by icepyx
```

JessicaS11:
> I got a bit frustrated trying to get the integration test "secure approval" process working the way I want it to. The event isn't triggering, and I don't know how (of if it's possible) to debug that.

Is it possible that it's because the action isn't yet in development?

.github/workflows/integration_test.yml
JessicaS11: I'm fine with the rename, I just wasn't sure if they were "accurate" titles given we haven't separated the tests yet.

.github/workflows/integration_test.yml
mfisher87: >I would definitely consider it part of this PR. I don't want to have a setup running that will trigger so many runs.

Don't want to push back too hard, because I am fine doing it however, but I do want to raise that Travis CI behaves this way already:

https://github.com/icesat2py/icepyx/blob/development/.travis.yml#L21-L24

I would lean towards breaking this work into two PRs: (1) migrate Travis -> GH; (2) more efficient. When those PRs are merged into `development`, they will produce two clearer commits and entries in the changelog. I'm OK with immediately following this PR up with step 2 if you're OK with dividing it!

.github/workflows/integration_test.yml
mfisher87: There are definitely some things I would call "integration tests" that will run in the unit tests job. I didn't want to make this PR huge by actually digging in to that though :)

mfisher87:
>Is it possible that it's because the action isn't yet in development?

I did run in to that initially. I've been testing on my fork, and I did merge into `development` over there, and still not seeing the events I expected!

.github/workflows/integration_test.yml
JessicaS11: > There are definitely some things I would call "integration tests" that will run in the unit tests job. I didn't want to make this PR huge by actually digging in to that though :)

That's why I was wondering aloud if it's a misnomer to rename before we've done the separation. I don't have strong feelings about it but didn't want to create a more-confusing situation for someone else in the future.

mfisher87:
@JessicaS11 I stealth-edited some TODOs in the description of this PR. Just realized Travis will block merge if it doesn't run :)

image

.github/workflows/integration_test.yml
JessicaS11: > Travis CI behaves this way already:

Not quite...

>  - name: behind Earthdata
>  if: branch = main OR commit_message =~ nsidc_tests OR type = cron

This isn't running the behind-earthdata tests on the development branch unless "nsidc_tests" is in the commit message. As a result, it only runs for the cron or before releases (when the PR is to main).

.github/workflows/integration_test.yml
mfisher87: How do you feel about the priority of separating the tests? If we're not going to do it anytime soon, perhaps different names are appropriate. Perhaps just "Tests" and "Behind EDL tests"? :shrug:

.github/workflows/integration_test.yml
mfisher87: Good catch, I was being too hasty in my reading! Would you prefer me to make this PR more reflect the Travis setup? What was the cron interval? It's not shown in the config.

.github/workflows/integration_test.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          fetch-depth: 0
+
+      - uses: "./.github/actions/install-icepyx"
+        with:
+          python-version: "3.12"
+
+      - name: "Run tests"
+        env:
+          EARTHDATA_USERNAME: "${{ secrets.EARTHDATA_USERNAME }}"
```
mfisher87, line 30:
```suggestion
```

Remove username envvar; it's provided as a fixture already.

.github/workflows/integration_test.yml
JessicaS11: Cron was weekly. I'm okay changing the config (especially if GH Actions has options that allow us to take a better approach). As we're conversing, I'm recalling that the cron was an effort to catch anything that got messed up through work going into development without having these tests run for every push to every PR to that branch. So I'm happy to nix the cron and have some filters that run it more selectively instead (for PRs to development; we'll probably want to keep it as is for main).

.github/workflows/integration_test.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+  push:
+    branches:
+      - "main"
+      - "development"
+
+
+jobs:
+  test:
+    name: "Test"
```
JessicaS11, line 14:
```suggestion
    name: "Integration Tests"
```

For some reason this is the name GH uses in the actions interfaces... so it would mean we have a bunch of "test" options in the branch rules, which will mean I will have to forever cross reference.

.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+  push:
+    branches:
+      - "main"
+      - "development"
+
+
+jobs:
+  test:
+    runs-on: "ubuntu-latest"
```
JessicaS11, line 15:
```suggestion
  test:
    name: "Unit test"
    runs-on: "ubuntu-latest"
```

.github/workflows/integration_test.yml
mfisher87: What do you think of this filter? https://github.com/icesat2py/icepyx/pull/580/commits/098bb985bbb6b551159ddd23409a3adc900a8d37

.github/workflows/integration_test.yml
JessicaS11: > How do you feel about the priority of separating the tests?

Medium. Honestly, I think I'd prefer to have more tests/better test coverage. I'm guessing that ultimately we might end up with three categories: behind EDL, unit, and integration. So renaming as you suggest "tests" and "behind EDL tests" lets us move forward with later separating out things in "tests" as needed.

.github/workflows/integration_test.yml
JessicaS11:
```suggestion
    name: "behind EDL test"
```

mfisher87: Not sure I follow. I'm seeing this, and I was trying to avoid "Integration tests/Integration tests":

image

.github/workflows/integration_test.yml
mfisher87: I've been thinking of "behind EDL" as a subset of "integration", and using for example [skipif](https://docs.pytest.org/en/stable/how-to/skipping.html#skipif) to control whether those tests run in certain environments. What do you think?

mfisher87:
I was confused why the tests weren't passing, and it turns out they also depend on an envvar named `$NSIDC_LOGIN`. When they are both populated with the password, things go!

.github/workflows/integration_test.yml
JessicaS11: It's inconsistent. But viewed in the action tab, it only shows "Test". And when I tried to add the tests as branch rules, all I can see is like 5 versions of "test" (with varying capitals and [3.9] or [3.12]). The ones with python versions become the only names I can logically link to a specific test.

image

mfisher87: Yuck! Thanks, GitHub :laughing:

mfisher87: 7bb6cb5

.github/workflows/integration_test.yml
JessicaS11: It still feels to me like it will run a lot (and a lot more than necessary)... I'd like to revisit with fresh eyes tomorrow, but what are your thoughts on something like:
```
on:
  push: # these are ultimately the runs that inform our README status badges and provide the "current status"
    branches:
      - "main"
      - "development"
  pull_request: # we want to know that we're not going to break anything with the release
    branches:
      - "main"
 pull_request_review:
    branches:
      - "development"
    paths:
      (list from above)
```

If I've got it right, we run: (1) once for each release (the push to main); (1) one (or more, but hopefully only one) time when we open a release PR from development to main (the pull_request trigger); and (1-2) once or twice for each PR to development (possibly once after it's approved if a relevant path is touched, the pull_request_review trigger, and once on the development branch for current status (the push to development)). By my calculations, this keeps us under ~50 downloads/month, probably with the exception of the next couple!

README.rst
```diff
@@ -38,11 +38,11 @@ Current development version (development branch): |Docs Status dev| |Travis dev
...
 
-.. |Travis main Build Status| image:: https://app.travis-ci.com/icesat2py/icepyx.svg?branch=main
-    :target: https://app.travis-ci.com/icesat2py/icepyx
+.. |Tests Status (main)| image:: https://github.com/icesat2py/icepyx/actions/workflows/integration_test.yml/badge.svg?branch=main
+    :target: https://github.com/icesat2py/icepyx/actions/workflows/integration_test.yml
 
-.. |Travis dev Build Status| image:: https://app.travis-ci.com/icesat2py/icepyx.svg?branch=development
-    :target: https://app.travis-ci.com/icesat2py/icepyx
+.. |Tests Status (dev)| image:: https://github.com/icesat2py/icepyx/actions/workflows/integration_test.yml/badge.svg?branch=development
+    :target: https://github.com/icesat2py/icepyx/actions/workflows/integration_test.yml
```
JessicaS11, line 45: I don't suppose there's a magical way for this to check both action statuses (i.e. we don't want it to show as passing if the non-integration tests fail)?

JessicaS11: There is! https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/adding-a-workflow-status-badge#using-the-event-parameter

.github/workflows/integration_test.yml
mfisher87: I've never used `pull_request_review` event before. I think we will only want to trigger on the "submitted" type for that event. Let's try it out! NOTE: This still won't fix our fork problem :)

JessicaS11: > I think we will only want to trigger on the "submitted" type for that event. Let's try it out! 

Or "approved"? I think this is how I set up the uml diagram action so it only runs once per PR.

> NOTE: This still won't fix our fork problem :)

Wouldn't that be nice...

mfisher87: 4183ece look OK?

README.rst
mfisher87: There isn't a way that I know of. The document you linked is just for filtering by event type. So we could show a badge for the status of the last integration test caused by a push event to the main branch.

If we want to show a combined badge for integration tests and unit tests, we will need to combine the workflows, to the best of my knowledge.

.github/workflows/integration_test.yml
mfisher87: >Or "approved"? I think this is how I set up the uml diagram action so it only runs once per PR.

There is no approved event type. That workflow checks the current state of the PR after triggering the workflow, and if it's not "approved", it sets the workflow to "skipped": https://github.com/icesat2py/icepyx/blob/a52bc378ab5817db38cd2b47b5ab0c5352128061/.github/workflows/uml_action.yml#L10

If there was an approved event type, then we could avoid the triggering in the first place :(

But since we have so many other possible triggers we would need something like `${{ github.event.action != "pull_request_review" || github.event.review.state == "approved" }}`. Thoughts?

README.rst
mfisher87: What about a separate badge for each class of tests?

.github/workflows/integration_test.yml
JessicaS11: Those sound like reasonable approaches (and we can adjust the triggers for the integration tests more broadly then).

.github/workflows/integration_test.yml
JessicaS11: > There is no approved event type.

Guess that was wishful thinking on my part then.

> Since we have other possible triggers we would need something like ${{ github.event.action != "pull_request_review" || github.event.review.state == "approved" }}. Otherwise, integration tests would be skipped on our release PRs until they were approved, and if we want that behavior we can remove the pull_request trigger. Thoughts?

I'm happy with the latest edits versus this. I don't think we tend to have a lot of official "reviews" until we're getting pretty close to merging, and it would be great to have the tests run on release PRs right away.

README.rst
JessicaS11: I read the docs too literally, leading me to my not-quite-right interpretation!

> If we want to show a combined badge for integration tests and unit tests, we will need to combine the workflows, to the best of my knowledge. With the complexity of our triggers, and the future need to approve integration tests from forks, I don't want to do that 😆

100%. We should not combine the test workflows for badging purposes!

> What about a separate badge for each class of tests?

I think that's probably the way to go.

README.rst
```diff
@@ -38,11 +38,11 @@ Current development version (development branch): |Docs Status dev| |Travis dev
 .. |Docs Status dev| image:: https://readthedocs.org/projects/icepyx/badge/?version=development
    :target: https://icepyx.readthedocs.io/en/development
 
-.. |Travis main Build Status| image:: https://app.travis-ci.com/icesat2py/icepyx.svg?branch=main
-    :target: https://app.travis-ci.com/icesat2py/icepyx
+.. |Tests Status (main)| image:: https://github.com/icesat2py/icepyx/actions/workflows/integration_test.yml/badge.svg?branch=main
+    :target: https://github.com/icesat2py/icepyx/actions/workflows/integration_test.yml
```
JessicaS11, line 43:
```suggestion
.. |Unit Tests Status (main)| image:: https://github.com/icesat2py/icepyx/actions/workflows/unit_test.yml/badge.svg?branch=main
    :target: https://github.com/icesat2py/icepyx/actions/workflows/unit_test.yml

.. |Integration Tests Status (main)| image:: https://github.com/icesat2py/icepyx/actions/workflows/integration_test.yml/badge.svg?branch=main
    :target: https://github.com/icesat2py/icepyx/actions/workflows/integration_test.yml
```

README.rst
JessicaS11: Lines 10 and 12:
```
Latest release (main branch): |Docs Status main| |Unit Tests Status (main)| |Integration Tests Status (main)| |Code Coverage main|
Current development version (development branch): |Docs Status dev| |Unit Tests Status (dev)| |Integration Tests Status (dev)| |Code Coverage dev| |Pre-commit dev|
```

.github/workflows/integration_test.yml
JessicaS11: > I'm happy with the latest edits versus this.  I don't think we tend to have a lot of official "reviews" until we're getting pretty close to merging, and it would be great to have the tests run on release PRs right away.

I might need to rescind this. I'm noticing the integration test is running, which makes me think that GH considers a comment (regardless of approval) to be a "review", thus triggering the action for any comment made on the PR.

JessicaS11: I'm thinking the least unpalatable option at this point is your suggestion to remove the PR trigger and add something like `${{ github.event.action != "pull_request_review" || github.event.review.state == "approved" }}`.

mfisher87: Let's try it on for size!

README.rst
mfisher87: I think we've got this now, please unresolve if you disagree :)

JessicaS11 (approved):
Thanks for all the problem solving and changes on this one! I'm excited to get CI moved onto GH, so thanks for making it happen, Matt.

mfisher87:
Any time! :rocket:

PR #581, opened 2024-08-22
Reinstate test_download_granules_without_subsetting

weiji14:
Trying to increase code coverage of `ipx.Query` class's `order_granules` and `download_granules` methods, by ensuring that granules can be ordered from NSIDC and downloaded with the `subset=False` option.

TODO:
- [x] Add test which should fail due to not capturing stdout and stderr messages
- [x] Regex match on stdout and stderr so test will pass
- [x] Check that downloaded files are actually not subsetted (kinda)

Adjacent stuff done in this PR

- [x] Fix a possible race condition in case the NSIDC order status if completed right at the start (e.g. when order was cached/done already).
~- [x] Upload code coverage for 'behind Earthdata' tests via Travis CI~


icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -49,8 +54,33 @@ def test_download_granules_with_subsetting(reg, session):
...
+        "request_mode": "async",
+        "short_name": "ATL06",
+        "version": "006",
+    }
+    assert len(reg.granules.orderIDs) == 2
+    assert int(reg.granules.orderIDs[0]) >= 5_000_000_000_000
+
+    reg.download_granules(path=path)
+    assert len(glob.glob(pathname=f"{path}/processed_ATL06_201902*.h5")) == 3
+    # check that the max extent of the downloaded granules isn't subsetted
```
JessicaS11, line 86: > assert len(glob.glob(pathname=f"{path}/processed_ATL06_201902*.h5")) == 3

If the granules are NOT subset, would they still have `processed_` preceding the filename? I thought that wasn't the case, but would need to verify.

>   # check that the max extent of the downloaded granules isn't subsetted

Is checking the (approximate) file size a viable way to do this? I have a vague notion that I started writing this test because I noticed that all files requested via icepyx were getting subset, even when `subset=False` was set by the user. I think we'd need to check the non-setting of `reg.subsetparams` to ensure none of those are being passed, but I can't recall offhand if it's the use of CMR kwargs (vs subset kwargs) for space and time or something else that triggers in the API whether or not subsetting happens.

weiji14: Yeah, I thought they shouldn't have a `processed_` prefix too, wondering if it's because it's cached the previous (non-subsetted) order and so this is what is returned (seems like a bug)? And yes, good idea on checking the filesize, I didn't want to open the file to check the extent of the point cloud, and totally forgot about just checking the filesize :laughing:

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -49,8 +54,40 @@ def test_download_granules_with_subsetting(reg, session):
...
+    reg.download_granules(path=path)
+    # check that the max extent of the downloaded granules isn't subsetted
+    assert len(glob.glob(pathname=f"{path}/ATL06_201902*.iso.xml")) == 3
+    h5_paths = glob.glob(pathname=f"{path}/ATL06_201902*.h5")
+    assert len(h5_paths) == 3
+    assert [os.path.getsize(filename=p) for p in h5_paths] == [
+        65120027,  # 62.1 MiB
+        53228429,  # 50.8 MiB
+        49749227,  # 47.4 MiB
+    ]
```
weiji14, line 93: Checking for the non-subsetted 'ATL06_*.h5' files here now. Let me know if there's a better way to check for non-subsetted files instead of just the filesize here.

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -49,8 +54,40 @@ def test_download_granules_with_subsetting(reg, session):
...
+
+    reg.order_granules(verbose=False, subset=False, email=False)
+    out, err = capsys.readouterr()  # capture stdout and stderr
+    assert out.startswith(
+        "Total number of data order requests is  1  for  3  granules.\n"
+        "Data request  1  of  1  is submitting to NSIDC\n"
+    )
+    assert err == ""
+
+    assert reg.reqparams == {
```
JessicaS11, line 72: Anyone know why codecov gives "not covered by tests" warnings for the tests?

weiji14: Hmm, the tests should be covered if they ran. I can try again later when changing from travis to gh actions and see if it changes things.

weiji14: Ah, I think the codecov report wasn't uploaded because the travis test failed (one of the assert tests wasn't deterministic, but I fixed that in 82fa916d176619476668749902e15674b10c5fd1). I've switched to GitHub Actions and am re-running the integration test again.

weiji14:
Got the `test_download_granules_without_subsetting` test to pass when I triggered it at https://github.com/icesat2py/icepyx/actions/runs/10631387124/job/29472168206#step:4:25.

icepyx/tests/test_behind_NSIDC_API_login.py
JessicaS11: @mfisher87 Any ideas other than file size?

icepyx/tests/test_behind_NSIDC_API_login.py
```diff
@@ -49,8 +54,40 @@ def test_download_granules_with_subsetting(reg, session):
...
+        "page_size": 2000,
+        "request_mode": "async",
+        "short_name": "ATL06",
+        "version": "006",
+    }
+    assert len(reg.granules.orderIDs) == 2
+    assert int(reg.granules.orderIDs[0]) >= 5_000_000_000_000
+
+    reg.download_granules(path=path)
+    # check that the max extent of the downloaded granules isn't subsetted
```
JessicaS11, line 85:
```suggestion
    # check that there are the right number of files of the correct size
```

JessicaS11 (approved):
Thanks, @weiji14! I think this looks great.

JessicaS11:
> Got the test_download_granules_without_subsetting test to pass when I triggered it

Out of curiousity, how did you trigger it? It doesn't have a manual trigger, and should only run with an "approved" PR review status (but we're having some issues with this, clearly, so trying to learn how you were able to make it go).

weiji14:
> > Got the test_download_granules_without_subsetting test to pass when I triggered it
> 
> Out of curiousity, how did you trigger it? It doesn't have a manual trigger, and should only run with an "approved" PR review status (but we're having some issues with this, clearly, so trying to learn how you were able to make it go).

I triggered it at commit d3a0b8a39719ce2519c9a686348932cf209b78aa by modifying `.github/workflows/integration_test.yml` to run on all branches for pull requests (not just those that target the `main` branch). Downside with this trick is that you'll need to have a commit to revert the changes later.

JessicaS11:
> > > Got the test_download_granules_without_subsetting test to pass when I triggered it
> > 
> > 
> > Out of curiousity, how did you trigger it? It doesn't have a manual trigger, and should only run with an "approved" PR review status (but we're having some issues with this, clearly, so trying to learn how you were able to make it go).
> 
> I triggered it at commit [d3a0b8a](https://github.com/icesat2py/icepyx/commit/d3a0b8a39719ce2519c9a686348932cf209b78aa) by modifying `.github/workflows/integration_test.yml` to run on all branches for pull requests (not just those that target the `main` branch). Downside with this trick is that you'll need to have a commit to revert the changes later.

Sneaky! Would that work from a fork too?

JessicaS11 (approved):
Re-approving to trigger integration test.

weiji14:
> Sneaky! Would that work from a fork too?

Not sure :sweat_smile: I had a look at the [docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request) and couldn't figure it out. Understandably, this sounds like a security concern, but secrets shouldn't be accesible from forks, and there should be an 'Approve to run' button that is set for new contributors according to https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

icepyx/tests/test_behind_NSIDC_API_login.py
mfisher87: Woops, missed this. I don't have any other ideas right now!

PR #588, opened 2024-08-26
Workflow to approve integration tests from forks

mfisher87:
Follow up on #580.

> [!IMPORTANT]
>
> [Preventing Pwn Requests is important](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/).
> 
> Secrets exfiltration is a risk if we try to allow integration tests that use secrets to run on PRs from forks.

I took a stab and [ran in to a wall](https://github.com/mfisher87/icepyx/pull/3). Taking a pause on this to focus on other high priorities. I may be entirely misguided in my approach. Any expertise would be appreciated :)


mfisher87:
Superceded by #617

Issue #569, opened 2024-08-18
Python version support policy


mfisher87:
Hey all, I noticed in a previous post [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) was used justification for dropping Python 3.6. It's been a couple years, and the community (including numpy) is now coalescing around this very similar standard, SPEC0: https://scientific-python.org/specs/spec-0000/

Under SPEC0 and NEP29, we can drop support for 3.7, 3.8, and 3.9. Do we have an official policy? Should it wait for v2?

JessicaS11:
> Do we have an official policy?

We do not.

> Also, are we following semantic versioning? 

It's buried in the [release guide](https://icepyx.readthedocs.io/en/latest/contributing/release_guide.html), but we do!

> We have a v2 branch in progress, and I'm fairly sure dropping support for a Python version justifies a major bump if we are. At least that's what I've seen in practice!

That makes sense and seems like a good opportunity to do this bit of housekeeping.

mfisher87:
Thanks, Jessica! Does the new issue title seem like accurate and complete?

weiji14:
Seems like we can drop Python 3.10 as well, so let's drop Python 3.9 and 3.10 (#655) after Python 3.13 support is added in #622

JessicaS11:
Should we add a note to the docs around using SPEC0 as our standard policy for dropping support of older versions of Python? It sounds like there's general agreement around that and this way it's written down for future devs.

mfisher87:
Definitely :) That's what I had in mind with the "documentation" label I added, but I should have probably been more explicit!!

PR #593, opened 2024-08-27
Typecheck some API interfaces

mfisher87:
I'm going to be changing a lot in the refactor to use Harmony. The typechecker is going to help find sneaky bugs along the way. This is a start, but I'd really like feedback.

My main goal here is to avoid refactoring the code more than needed and set the stage for future cleanups and refactors. E.g. I think long-term, we will want to split up the `Parameters` class instead of having it handle all the kinds of parameters.

I'm using pyright here for the first time, and I'm pleased with it compared to Mypy so far. The reason I switched to Pyright: https://github.com/python/mypy/issues/12098


icepyx/core/APIformatting.py
```diff
@@ -181,12 +184,56 @@ def to_string(params):
...
+class _FmtedKeysDescriptor:
+    """Enable the Parameters class' fmted_keys property to be typechecked correctly.
+
+    See: https://github.com/microsoft/pyright/issues/3071#issuecomment-1043978070
+    """
+
+    @overload
+    def __get__(  # type: ignore
+        self,
+        instance: 'Parameters[Literal["CMR"]]',
```
mfisher87, line 201: This allows the return type of `fmted_keys` method to depend on the `partype` the class is initialized with

mfisher87: Woops, commented one line too low

requirements-dev.txt
```diff
@@ -3,3 +3,8 @@ pypistats
 pytest>=4.6
 pytest-cov
 responses
+pyright
+pandas-stubs
+types-docutils
+types-requests
+types-tqdm
```
mfisher87, line 10: Some libraries don't come with type annotations, but awesome contributors have packaged them separately. Sometimes these are not complete. The official place for external type annotations is [typeshed](https://github.com/python/typeshed). Those are the ones prefixed with `types-` in this list.

icepyx/core/APIformatting.py
mfisher87: Actually... it shows up as one line too low in the diff view, but it shows up in the right place in the conversation view. Anyone else seeing that?

.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install package and test dependencies
+        run: |
+          python -m pip install .[complete]
```
mfisher87, line 26: We need the geoviews library to be installed to typecheck things we import from it

icepyx/core/APIformatting.py
```diff
@@ -181,12 +184,56 @@ def to_string(params):
     return "&".join(param_list)
 
 
+ParameterType = Literal["CMR", "required", "subset"]
+# DevGoal: When Python 3.12 is minimum supported version, migrate to PEP695 style
+T = TypeVar("T", bound=ParameterType)
```
mfisher87, line 189: This is the grossest thing to me about the Python type system. The generics are much better in 3.12. Basically, we're going to use this to mark the `Parameter` function as varying on the `partype` parameter.

icepyx/core/types.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import annotations
+
+from typing import Literal, TypedDict
+from typing_extensions import NotRequired
+
+
+CMRParamsBase = TypedDict(
```
mfisher87, line 7: I'm shooting for annotating the code as-is to enable deeper future refactors with less mistakes. I would prefer to pass around dataclasses :)

Here we're using the "old" way of defining a TypedDict because the keys we want to use are not valid in the new way. The new way is to define them as a class.

icepyx/core/types.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class CMRParamsWithPolygon(CMRParamsBase):
+    polygon: str
+
+
+CMRParams = CMRParamsWithBbox | CMRParamsWithPolygon
+
+
+class EGISpecificParamsBase(TypedDict):
```
mfisher87, line 29: I'm like sure I've got stuff wrong here, but it's right enough to pass static analysis :laughing:

pyproject.toml
```diff
@@ -102,3 +102,31 @@ ignore = [
...
+  "icepyx",
+]
+exclude = [
+  "**/__pycache__",
+  "icepyx/tests",
+]
+# DevGoal: Remove all ignores
+ignore = [
+  "icepyx/quest/*",
+  "icepyx/core/APIformatting.py",
```
mfisher87, line 121: I want to remove this ignore first, but I didn't want this PR to get to huge. But I can also include that change if requested :)

.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
```
JessicaS11, line 22: why not v3.12?

icepyx/core/visualization.py
```diff
@@ -466,7 +466,7 @@ def parallel_request_OA(self) -> da.array:
             OA_data_da = da.concatenate(requested_OA_data, axis=0)
             return OA_data_da
 
-    def viz_elevation(self) -> (hv.DynamicMap, hv.Layout):
+    def viz_elevation(self) -> tuple[hv.DynamicMap, hv.Layout]:
```
JessicaS11, line 469: Aren't these the same because `()` is by definition a tuple? Or is there something internally that happens differently?

icepyx/core/granules.py
```diff
@@ -7,12 +9,15 @@
...
-from xml.etree import ElementTree as ET
 import zipfile
+from typing import Final
+from requests.compat import unquote
+from xml.etree import ElementTree as ET
 
 import icepyx.core.APIformatting as apifmt
-from icepyx.core.auth import EarthdataAuthMixin
 import icepyx.core.exceptions
+from icepyx.core.auth import EarthdataAuthMixin
```
JessicaS11, line 19: Why is this reordering happening? I thought we'd addressed it (though I can't quickly find in which ruff/reformatting PR).

icepyx/core/granules.py
```diff
@@ -361,7 +370,7 @@ def place_order(
             request.raise_for_status()
             esir_root = ET.fromstring(request.content)
             if verbose is True:
-                print("Order request URL: ", requests.utils.unquote(request.url))
+                print("Order request URL: ", unquote(request.url))
```
JessicaS11, line 373: We now import `requests.compat.unquote` - is this just an alias?

icepyx/core/types.py
```diff
@@ -0,0 +1,79 @@
...
+
+CMRParamsBase = TypedDict(
+    "CMRParamsBase",
+    {
+        "temporal": NotRequired[str],
+        "options[readable_granule_name][pattern]": NotRequired[str],
+        "options[spatial][or]": NotRequired[str],
+        "readable_granule_name[]": NotRequired[str],
+    },
+)
```
JessicaS11, line 15: I'm not sure I fully understand why:

```
class CMRParamsBase(TypedDict):
        temporal: NotRequired[str],
        options[readable_granule_name][pattern]: NotRequired[str],
        options[spatial][or]: NotRequired[str],
        readable_granule_name[]: NotRequired[str],
```

(the "new" way) isn't an option?

> I'm shooting for annotating the code as-is to enable deeper future refactors with less mistakes. I would prefer to pass around dataclasses :)

Totally makes sense. Still learning about dataclasses so can't agree with you yet!

pyproject.toml
JessicaS11: Does pyright ignore places there aren't type annotations? Without having run pyright yet, I'm trying to understand why these would currently fail (and need to be ignored).

icepyx/core/visualization.py
mfisher87: These are actually different; note the square brackets instead of parenthesis. `(a, b)` is equivalent to `tuple([a, b])`; these both create a runtime tuple object. When writing type annotations, the brackets indicate "arguments" to the tuple _type_. Things on the right-hand side of the `->` are always types and don't affect the runtime behavior of the function (though they _can_ be read as metadata intentionally at runtime!), and same for things on the right-hand side of the `:` in function argument signatures.

What we're saying here is that the function returns an object of type `tuple` with two elements: first element is an object of type `hv.DynamicMap`, and the second is an object of type `hv.Layout`.

icepyx/core/granules.py
mfisher87: I thought so too. :thinking:

icepyx/core/types.py
mfisher87: Good question! Python will try to access objects named `options` (to call `__getitem__` IIRC), then `readable_granule_name`, and `pattern` to evaluate that code and you'll get an error:

```python
>>> class Foo(TypedDict):
...   quz[qux]: int
... 
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in Foo
NameError: name 'quz' is not defined
```

Class attributes must follow naming rules, e.g. can't start with a number or symbol, can't use certain symbols, etc. I'm struggling to find the doc in the time I have right now :X

icepyx/core/granules.py
mfisher87: isn't not an alias, i.e. we're not using `as` in the import; we're importing `unquote` with its original name:

```
from requests.compat import unquote
```

This change was prompted by the typechecker. I didn't dig deep in to it since the change seemed fine and the queue of typechecker errors was pretty long :)

.github/workflows/typecheck.yml
mfisher87: I want to test with annotations that are compatible with the oldest Python version we support. There's a similar setting in `pyproject.toml` for pyright itself!

pyproject.toml
mfisher87: No, where there are no annotations, Pyright or Mypy will do inference. It's capable of finding bugs that way, too!

For example, `typing_test.py`:

```python
foo = "bar"
foo / 2
```

```bash
$ mypy typing_test.py 
typing_test.py:2: error: Unsupported operand types for / ("str" and "int")  [operator]
Found 1 error in 1 file (checked 1 source file)

$ pyright typing_test.py 
WARNING: there is a new pyright version available (v1.1.377 -> v1.1.378).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`

/tmp/typing_test.py
  /tmp/typing_test.py:2:1 - error: Operator "/" not supported for types "Literal['bar']" and "Literal[2]" (reportOperatorIssue)
  /tmp/typing_test.py:2:1 - warning: Expression value is unused (reportUnusedExpression)
1 error, 1 warning, 0 informations 
```

icepyx/core/granules.py
JessicaS11: The part that was confusing to me was that we went from `requests.utils.unquote` to `requests.compat.unquote`. They both lead to a function called `unquote` but in two different places, so I wanted to make sure they were doing the same thing. I was wondering if `requests.compat.unquote` was an alias for `requests.utils.unquote`.

icepyx/core/types.py
JessicaS11: Got it - thanks! [Docs](https://docs.python.org/3.12/library/typing.html#typing.TypedDict) (turns out I had them open).

pyproject.toml
JessicaS11: Got it - I ran it and now I'm seeing all the inferred issues it's catching.

pyproject.toml
```diff
@@ -102,3 +102,31 @@ ignore = [
...
+  "icepyx/core/auth.py",
+  "icepyx/core/exceptions.py",
+  "icepyx/core/icesat2data.py",
+  "icepyx/core/is2ref.py",
+  "icepyx/core/read.py",
+  "icepyx/core/spatial.py",
+  "icepyx/core/temporal.py",
+  "icepyx/core/validate_inputs.py",
```
JessicaS11, line 129:
```suggestion
  "icepyx/core/is2ref.py",
  "icepyx/core/read.py",
  "icepyx/core/spatial.py",
```

FWIW, it looks like we could remove these few already!

JessicaS11: Sorry that rendered so obnoxiously...

icepyx/core/types.py
```diff
@@ -0,0 +1,79 @@
...
+class EGISpecificParamsSearch(EGISpecificParamsBase):
+    """Parameters for searching through EGI."""
+
+
+class EGISpecificParamsOrder(EGISpecificParamsBase):
+    """Parameters for ordering through EGI."""
+
+    # TODO: Does this type need page_* attributes?
+
+
```
JessicaS11, line 60:
```suggestion
```

order = download as far as parameters are concerned (the parameters are submitted as an order, and download just grabs the files staged using the order IDs.

icepyx/core/types.py
```diff
@@ -0,0 +1,79 @@
...
+    See: https://wiki.earthdata.nasa.gov/display/SDPSDOCS/EGI+Programmatic+Access+Documentation
+
+    EGI shares parameters with CMR, so this data is used in conjunction with CMRParams
+    to build EGI requests.
+
+    TODO: Validate more strongly (with Pydantic and its annotated types?
+    https://docs.pydantic.dev/latest/concepts/types/#composing-types-via-annotated):
+
+    * short_name is `ATL##` (or Literal list of values?)
+    * version is 1-3 digits
```
JessicaS11, line 41: there's a list of valid products in `is2ref._validate_product`. Some have the format `ATL##QL`.

I believe the version is always 3 digits (e.g. `006` vs `6`).

icepyx/core/types.py
JessicaS11: It looks like we don't have any parameters set up for subsetting. This especially comes into play because EGI takes different parameters than CMR for time and space (`time` vs `temporal` and `bbox` versus `bounding_box`).

icepyx/core/APIformatting.py
```diff
@@ -181,12 +184,56 @@ def to_string(params):
...
+        instance: 'Parameters[Literal["CMR"]]',
+        owner: Any,
+    ) -> CMRParams: ...
+
+    @overload
+    def __get__(
+        self,
+        instance: 'Parameters[Literal["required"]]',
+        owner: Any,
+    ) -> EGISpecificParams: ...
```
JessicaS11, line 210: the "required" type maps to either `EGISpecificParamsSearch` or `EGISpecificParamsDownload` but NOT ``EGISpecificParamsSubset` (all three are included in `EGISpecificParams`).

icepyx/core/APIformatting.py
```diff
@@ -181,12 +184,56 @@ def to_string(params):
...
+        return instance._fmted_keys
+
+
 # ----------------------------------------------------------------------
 # DevNote: Currently, this class is not tested!!
 # DevGoal: this could be expanded, similar to the variables class, to provide users with valid options if need be
 # DevGoal: currently this does not do much by way of checking/formatting of other subsetting options (reprojection or formats)
 # it would be great to incorporate that so that people can't just feed any keywords in...
-class Parameters:
+class Parameters(Generic[T]):
```
JessicaS11, line 236: I do not 100% understand the changes here (i.e. I could not write code in this way), but with the exception of the notes I've left, they seem to make sense.

icepyx/core/granules.py
```diff
@@ -168,14 +173,19 @@ def __init__(
     # ----------------------------------------------------------------------
     # Methods
 
-    def get_avail(self, CMRparams, reqparams, cloud=False):
+    def get_avail(
+        self,
+        CMRparams: CMRParams | None,
+        reqparams: EGISpecificParams | None,
```
JessicaS11, line 179: see related comment above on `EGISpecificParams` being too inclusive.

icepyx/core/granules.py
```diff
@@ -276,11 +286,11 @@ def place_order(
...
         Parameters
         ----------
-        CMRparams : dictionary
+        CMRparams :
             Dictionary of properly formatted CMR search parameters.
-        reqparams : dictionary
+        reqparams :
             Dictionary of properly formatted parameters required for searching, ordering,
-            or downloading from NSIDC.
+            or downloading from EGI.
```
JessicaS11, line 293:
```suggestion
            or downloading from NSIDC (via their EGI system).
```

Nobody outside NSIDC has ever heard of EGI...

icepyx/core/query.py
```diff
@@ -392,6 +394,10 @@ class Query(GenQuery, EarthdataAuthMixin):
     GenQuery
     """
 
+    _CMRparams: apifmt.CMRParameters
+    _reqparams: apifmt.RequiredParameters
+    _subsetparams: apifmt.SubsetParameters | None
```
JessicaS11, line 399: What is this doing?

icepyx/core/granules.py
mfisher87: Ah. Sorry, my mistake! `requests` is just re-importing the `unquote` function in the `utils` module; it's original source is the `compat` module:

https://github.com/psf/requests/blob/f12ccbef6d6b95564da8d22e280d28c39d53f0e9/src/requests/utils.py#L44-L52

icepyx/core/APIformatting.py
mfisher87: Thank you for this clarity! :bow:

icepyx/core/APIformatting.py
mfisher87: I want to in particular emphasize how much I dislike the `TypeVar` style of annotating generic functions. It's much better in Python 3.12:  https://peps.python.org/pep-0695/

icepyx/core/granules.py
mfisher87: Personally, I think it's a bit confusing to associate NSIDC and EGI like this (granted, `nsidc` is in the EGI url as it's running on hardware we manage, but I think that's largely irrelevant to the user). All public information about it is under the NASA wiki (https://wiki.earthdata.nasa.gov/display/SDPSDOCS/EGI+Programmatic+Access+Documentation). There are 3 DAACs which are "compatible" with this system: NSIDC, LP, and ASDC. That's just my opinion, I'm fine with whatever you prefer :) Especially since EGI is going away, none of this matters :laughing: I'm hyperfocusing :laughing:

icepyx/core/query.py
mfisher87: This is specifying the type of these class attributes. So if we were do to `self._reqparams = 5`, the typechecker would yell at us for violating this type specification.

icepyx/core/granules.py
JessicaS11: Gotcha - that totally makes sense to switch then!

icepyx/core/query.py
JessicaS11: > This is specifying the type of these class attributes. 

I thought this was happening below in (e.g. line 540) `def CMRparams(self) -> CMRParams:`, but missed the distinction between typing the public and private versions of `CMRparams`. Thanks for clarifying!

icepyx/core/granules.py
JessicaS11: > Personally, I think it's a bit confusing to associate NSIDC and EGI like this

Then perhaps keep it at its original "downloading from NSIDC"? I think it would be more confusing to users if they somehow think they're not downloading their data from NSIDC because we're naming the specific endpoint being hit (and underlying tool being used). After the update we can try to explain this clearly in the tutorials (if it's not already).

icepyx/core/granules.py
mfisher87: Rebased, this should be resolved.

pyproject.toml
mfisher87: 27f4641

icepyx/core/granules.py
mfisher87: I do think your suggestion is an improvement over the original, because at least then there's a way to Google for the term!

icepyx/core/types.py
mfisher87: Used that explicit list to exactly type `short_name`: 0440fa9

icepyx/core/APIformatting.py
mfisher87: Addressed here df78ca2  what do you think? I changed the names in a way that I feel is a bit clearer too.

icepyx/core/granules.py
mfisher87: df78ca2

icepyx/core/types.py
mfisher87: df78ca2

icepyx/core/types.py
mfisher87: df78ca2

icepyx/core/types.py
```diff
@@ -0,0 +1,111 @@
+from __future__ import annotations
+
+from typing import Literal, TypedDict, Union
+
+from typing_extensions import NotRequired
+
+IcesatProductShortName = Literal[
+    "ATL01",
```
JessicaS11, line 8: Got any suggested resources or keywords for reading up on type-checking versus input validation? I don't have the vocabulary to google beyond basics, and while I understand it's useful to check more than "is it a string?", to me this sort of type declaration should effectively be replacing the `is2ref._validate_product()` function (or better yet, turning that function into this plus a helpful error message - can you add custom messages to type checking?) rather than maintaining the list of literals in multiple spots.

icepyx/core/types.py
```diff
@@ -0,0 +1,111 @@
+from __future__ import annotations
+
+from typing import Literal, TypedDict, Union
+
+from typing_extensions import NotRequired
+
+IcesatProductShortName = Literal[
```
JessicaS11, line 7:
```suggestion
Icesat2ProductShortName = Literal[
```

icepyx/core/APIformatting.py
JessicaS11: As I'm starting to look through these updates, I can't help but wonder why we're putting so much effort into perfecting type checking for code we're going to remove or substantially alter... given that the changes will be breaking anyway and as I learn more about when and how type checkers do (or don't!) actively "run" with code, it's not clear to me how having this in place helps us detect bugs for anything that's EGI specific.

icepyx/core/types.py
```diff
@@ -0,0 +1,111 @@
...
+    * 0 < page_size <= 2000
+    """
+
+    short_name: IcesatProductShortName  # alias: "product"
+    version: str
+    page_size: int  # default 2000
+    page_num: int  # default 0
+
+
+class EGISpecificParamsSearch(EGISpecificParamsBase):
```
JessicaS11, line 75: It's pedantic and not worth spending time massaging since Harmony parameter names are better aligned with CMR, but technically these are CMR required search parameters/EGI required download parameters (but not EGI specific parameters). Functionally it doesn't matter, but see your statements about "naming" as a top challenge in Python. 😆

icepyx/core/types.py
```diff
@@ -0,0 +1,111 @@
...
+    projection_parameters: NotRequired[str]
+    Coverage: NotRequired[str]
+
+
+class EGIParamsSubsetBbox(EGIParamsSubsetBase):
+    bbox: str
+
+
+class EGIParamsSubsetBoundingShape(EGIParamsSubsetBase):
+    Boundingshape: str
```
JessicaS11, line 106:
```suggestion
    bbox: NotRequired[str]


class EGIParamsSubsetBoundingShape(EGIParamsSubsetBase):
    Boundingshape: NotRequired[str]
```

icepyx/core/granules.py
```diff
@@ -170,14 +175,19 @@ def __init__(
     # ----------------------------------------------------------------------
     # Methods
 
-    def get_avail(self, CMRparams, reqparams, cloud=False):
+    def get_avail(
+        self,
+        CMRparams: Optional[CMRParams],
+        reqparams: Optional[EGISpecificRequiredParams],
```
JessicaS11, line 181: None is not a valid input for these, so I don't think `Optional` is a valid type for them.

JessicaS11: I think we could actually limit `reqparams` further to `EGISpecificParamsSearch` too.

icepyx/core/granules.py
```diff
@@ -170,14 +175,19 @@ def __init__(
...
+        cloud=False,
+    ):
         """
         Get a list of available granules for the query object's parameters.
         Generates the `avail` attribute of the granules object.
 
         Parameters
         ----------
-        CMRparams : dictionary
+        CMRparams :
```
JessicaS11, line 190: I find this aspect of documentation incredibly helpful (even if they have custom types defined) and get irritated when I have no idea what flavor of input I should be supplying. Is the thing to do here have a link to the particular flavor of dictionary allowed (`CMRParams`) rather than nothing?

icepyx/core/types.py
mfisher87: Good call - would you suggest replacing the word "Specific" with "Required" in these names?

icepyx/core/granules.py
mfisher87: What I'd like to happen is for the documentation generator to pick this information up from the type hints instead of duplicating that information in the docstring. Since the docstring type information isn't actually checked, we run the risk of it becoming wrong!

We do this on _earthaccess_, but we're using mkdocs there. I have to check to make sure this is working with our doc generator.

- [ ] Check auto-generated docs include correct type information

icepyx/core/granules.py
mfisher87: Ooo, nice call, thank you!

icepyx/core/APIformatting.py
```diff
@@ -242,15 +300,7 @@ def poss_keys(self):
...
     #     return self._wanted
 
-    @property
-    def fmted_keys(self):
-        """
-        Returns the dictionary of formatted keys associated with the
-        parameter object.
-        """
-        return self._fmted_keys
-
```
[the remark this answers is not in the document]
mfisher87: It's not causing any failures because it isn't actually removed :) I moved the logic into a "descriptor":

https://github.com/icesat2py/icepyx/blob/434c28a64dcd7648684f4b30b170fde67146b843/icepyx/core/APIformatting.py#L198-L260

I did this because the `@overload` (which I used to represent the different return type possibilities based on `self.partype`) and `@property` decorators don't work together as expected.

icepyx/core/types.py
mfisher87: Neither are required? I thought it was one or the other!

icepyx/core/types.py
mfisher87: >Got any suggested resources or keywords for reading up on type-checking versus input validation?

I don't have a resource off the top of my head, but here's my quick explanation, and I'm always happy to chat about type checking :)

Type checking occurs without running the code. A type checker is like you or me reading the code and saying "That doesn't look right, isn't that supposed to be a string?" Except we tell the type checker explicitly "This is supposed to be a string" and let it read the code for us to look for instances where we're doing something we're not supposed to do with that thing we told the typechecker is supposed to be a string. Like using the division operator on it! This type will prevent us from writing code that uses an invalid short name, but won't catch users passing in an invalid short name at runtime.

Input validation will catch the latter case, but of course requires dilligence on our part to write the input validations and maintain them. We can get the best of both worlds with awesome tools like [pydantic's validation decorator](https://docs.pydantic.dev/latest/concepts/validation_decorator/) (beartype and other libraries can do this too!). Because type annotations add metadata to Python objects, Pydantic can inspect that metadata at runtime and generate input validation code for us! :exploding_head: 

>to me this sort of type declaration should effectively be replacing the is2ref._validate_product() function (or better yet, turning that function into this plus a helpful error message ...

:100: I think this is a wonderful use case for `@validate_call` or similar :) 

> ... can you add custom messages to type checking?

The pydantic decorator shown above doesn't seem to allow customizing the error messages it generates from runtime validations, but the error messages are very helpful in my experience. But please note that this different than type checking, it's runtime input validation based on type annotation metadata :)

.github/workflows/typecheck.yml
weiji14: Shouldn't we be setting this to Python 3.7, looking at https://github.com/icesat2py/icepyx/blob/1cf650486c6220b355bc1f0350f41e4167be4a61/pyproject.toml#L7

mfisher87: Probably yeah for 1.x, and then update everything to reflect a new version policy in 2.x? Related discussion here: https://github.com/icesat2py/icepyx/issues/569

weiji14: Cool, I'm just asking because I saw you using `Union[..., ...]` in places, and was wondering if we could use `... | ...` instead, but that's only available with Python 3.10 - https://docs.python.org/3.10/whatsnew/3.10.html#pep-604-new-type-union-operator

mfisher87: Yeah, I was originally doing that and that caused our 3.9 unit tests to fail :laughing:  So in either case, `3.10` is not correct here. We can run Ruff's pyupgrade ruleset to automatically migrate these :)

icepyx/core/APIformatting.py
mfisher87: Woops, forgot to post here after we chatted on Zoom. To greatly simplify, since the conversation isn't fresh in my mind anymore, type annotations are really helpful to me in refactoring; as I remove code to replace it with something new, the typechecker tells me about how calling code expects to work with the new code I need to write. The errors sort of act as a guide during refactors of any size. I'm largely targeting the API interfaces here, but the same benefits apply to other refactors in these modules or code which uses them.

Also, IDEs (and I assume also LLMs?) will give better hints with type annotations!

icepyx/core/types.py
mfisher87: 2247b2c

icepyx/core/granules.py
mfisher87: a8996d1

.github/workflows/typecheck.yml
mfisher87: I think it may not make sense to set to 3.7; I can't install from `requirements.txt` and `requirements-dev.txt`  in Python 3.7 using conda or pip, and can only install in Python 3.8 with pip, not conda. 3.9 installs just fine. Setting to 3.9!

dfe4c09

doc/source/conf.py
```diff
@@ -79,6 +83,31 @@
 nb_execution_mode = "off"
 suppress_warnings = ["myst.header"]  # suppress non-consecutive header warning
 
+
+# -- Options for Napoleon docstrign parsing ----------------------------------
```
JessicaS11, line 87:
```suggestion
# -- Options for Napoleon docstring parsing ----------------------------------
```

icepyx/core/granules.py
mfisher87: The auto-generated docs were not set up to parse your docstrings, which are in the numpy style. Instead, they were being rendered as straight RST. Now, the descriptions in the docstrings will be merged with the type annotations to produce a richer doc.

## New

image


## Old

image


TODO: 

- [ ] Generate docs for the types module, ensure linkage

requirements-docs.txt
```diff
@@ -6,6 +6,7 @@ numpydoc
 pybtex
 pygithub
 sphinx>=4.3
-sphinx-panels
+sphinx-autodoc-typehints>=2.0
+sphinx-design
```
mfisher87, line 10: I wasn't able to upgrade to get the needed docstring parsing without dropping `sphinx-panels`. I could probably figure out how with enough time but the migration was simple enough.

icepyx/core/APIformatting.py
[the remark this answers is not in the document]
JessicaS11: Okay... I hadn't connected all the pieces of logic to trace out where the descriptor returns the instance. Thanks!

icepyx/core/granules.py
```diff
@@ -261,13 +273,13 @@ def get_avail(self, CMRparams, reqparams, cloud=False):
     # DevGoal: add kwargs to allow subsetting and more control over request options.
     def place_order(
         self,
-        CMRparams,
-        reqparams,
+        CMRparams: CMRParams,
+        reqparams: EGISpecificParamsSearch,
```
JessicaS11, line 277:
```suggestion
        reqparams: EGISpecificParamsDownload,
```

icepyx/core/types.py
JessicaS11: That's great!

POI for future naming: we're currently in some water-muddying induced by which parameters are required in a given situation. The current list of EGISpecificParamsBase is actually required parameters for a CMR search AND EGI download (and there's no such thing as an EGI Search - EGI is only hit for order/download). But again, neither here nor there to make forward progress since it will all change anyway and there's better streamlining of parameter names across CMR + Harmony.

icepyx/core/types.py
JessicaS11: Technically not for subsetting. A spatial parameter is required for searching, which I think is why the CMR `bounding_box` or `polygon` gets passed but the EGI subsetting spatial parameter doesn't necessarily. But I'd also noticed some funny stuff recently with both sets of parameters being submitted and subsetting always happening even when the user selected `subset=False`, so it's possible the parameter submissions have gotten muddled at some point and aren't behaving as they should/we think they are. Since we know there must be some form of spatial info submitted, we suggest we leave them as required and continue identifying/working out any existing bugs moving forward rather than negotiating if they exist or not here!

doc/source/index.rst
```diff
@@ -31,86 +31,99 @@ ICESat-2 datasets to enable scientific discovery.
 To further enhance data discovery, we have developed the QUEST module to facilitate querying of ICESat-2 data and complimentary Argo oceanographic data, with additional dataset support expected in the future.
 
 
-.. panels::
-    :card: + intro-card text-center
-    :column: col-lg-4 col-md-4 col-sm-6 col-xs-12 p-2
-    :img-top-cls: pl-2 pr-2 pt-2 pb-2
```
JessicaS11, line 37: Can we re-add the title centering (and perhaps title text background)? The updated version looks slightly less polished to me and I think that's why...

.github/workflows/typecheck.yml
JessicaS11: Should we change this to 3.9 too?

icepyx/core/APIformatting.py
JessicaS11: > Woops, forgot to post here after we chatted on Zoom.

No worries - I asked on Zoom because I thought GH ate my comment (I refreshed multiple times and couldn't find it, so I assumed it was lost in a draft box somewhere).

requirements-docs.txt
JessicaS11: Ah - I was wondering what motivated this and GH didn't give me this review comment over in the diff view. 🤦‍♀️

icepyx/core/granules.py
mfisher87: > > What I'd like to happen is for the documentation generator to pick this information up from the type hints instead of duplicating that information in the docstring.
> 
> Awesome! Thanks for making that happen. How do I, as a user, now know what `CMRParamswithBbox` is? Is there a setting that will link this to the type module (so I can still know, for instance, that it's SOME type of dictionary)?

Yes, exactly. That's my intent with the TODO just above your comment. The user should be able to, once I'm done, see exactly the structure of the dict that the function expects!

>One of my documentation pet peeves is not knowing each library's custom types and having to dig into the source code to figure it out myself.

:100: Same!

doc/source/conf.py
mfisher87: Nice catch, thanks!

requirements-docs.txt
mfisher87: So many little annoying things like that with GitHub PRs that I expected to be fixed by now!

doc/source/index.rst
mfisher87: I agree, it does, good call. I will give it a shot!

.github/workflows/typecheck.yml
mfisher87: Do you mean the `requires-python` setting in `pyproject.toml`? I don't have a strong feeling on it, so I erred on the side of not touching it to minimize the scope of the PR. Happy to do it whatever way makes sense to everyone else!

icepyx/core/granules.py
mfisher87: 7f60849

image

There are some classes which couldn't be autodocumented because they contain symbols in the keys. There's an open issue on the sphinx repo and I linked to it in a comment!

icepyx/core/granules.py
mfisher87: is `Query.reqparams` always of type `EGISpecificParamsDownload`? Or only sometimes (this is my understanding)?

mfisher87: bb96a5f

I think this is probably the wrong way to handle things. We're telling the typechecker "Ignore your eyes and ears, just trust us" with the `cast()` call. Instead, we can add a "typeguard" which is a runtime check that asserts a narrower type, e.g. a literal `assert isinstance(...)` or `if isinstance(...): raise`

doc/source/index.rst
mfisher87: 7a423a8

I think the only difference I haven't accounted for is the grey background on the headers of the cards, but thought perhaps that wouldn't be worth the time. Let me know what you think!

icepyx/core/types.py
mfisher87: 6a4acd9

.github/workflows/typecheck.yml
weiji14: > I think we should also change from `~=` compatibility pinning to `>=3.9`.

Yes, this needs to go into release v1.3.0 !!

mfisher87: In that case, opened a separate PR so we can be sure that happens :) https://github.com/icesat2py/icepyx/pull/608

doc/source/index.rst
JessicaS11: Thanks! It looks great.

> I think the only difference I haven't accounted for is the grey background on the headers of the cards, but thought perhaps that wouldn't be worth the time.

Probably true. Not a problem for today.

doc/source/user_guide/documentation/types.rst
```diff
@@ -0,0 +1,11 @@
...
+=====
+
+.. automodule:: icepyx.core.types
+   :members:
+   :undoc-members:
+   :exclude-members: CMRParamsBase,CMRParamsWithBbox,CMRParamsWithPolygon
+
+.. COMMENT. `exclude-members` specified above is required because those models
+   contain symbols ('[', ']') in some keys, which sphinx doesn't like.
+   See: https://github.com/sphinx-doc/sphinx/issues/11039
```
JessicaS11, line 11: "[" in a key strikes again!

icepyx/core/granules.py
JessicaS11: Thanks for all of this - the additions and the explanations. All sounds good.

doc/source/user_guide/documentation/types.rst
mfisher87: Yes! Perhaps we can address that in v2 by switching from dictionaries to Pydantic objects (which can be read in from dicts or converted to dicts or query strings that look the way we need)!

icepyx/core/granules.py
JessicaS11: > is `Query.reqparams` always of type `EGISpecificParamsDownload`? Or only sometimes (this is my understanding)?

The latter - it's only sometimes of type `EGISpecificParamsDownload` (when `Query._reqparams._reqtype = "download"). Otherwise it's of type "search", which maps to our `EGIRequiredParamsSearch`.

> but perhaps we do know for sure that reqparams is always EGISpecificParamsDownload at the point we're cast()ing.

I think this is true.

icepyx/core/granules.py
```diff
@@ -261,13 +277,13 @@ def get_avail(self, CMRparams, reqparams, cloud=False):
     # DevGoal: add kwargs to allow subsetting and more control over request options.
     def place_order(
         self,
-        CMRparams,
-        reqparams,
+        CMRparams: CMRParams,
+        reqparams: EGIRequiredParamsDownload,
         subsetparams,
```
JessicaS11, line 282:
```suggestion
        subsetparams: Union[EGIRequiredParamsSubset, dict[Never, Never]],
```

I might have the syntax wrong. In some cases we specify the type from `types.py`, and in others from `APIformatting.py` (e.g. in `query.py`).

icepyx/core/types.py
JessicaS11: Thanks for this explanation. So it sounds like we can only maintain the list in one spot, it just needs to live on the type-checking side (plus a decorator for validating) rather than the validation side (plus a decorator for type checking). Looking forward to making this simplification in a future PR!

doc/source/user_guide/documentation/types.rst
JessicaS11: Or the folks on your end could stop creating keys with symbols in them. 🤣 I'm intrigued by these Pydantic objects and how they could help with this challenge.

JessicaS11 (approved):
Approving this for merging now that we've released the last of v1 (v1.3.0). I think we've worked out most of the kinks and figure any others will be ironed out as we continue moving towards the refactor in v2. I'll be on PTO Tuesday and I know in our discussion today you said this was a critical next step for you for making progress.

doc/source/user_guide/documentation/types.rst
mfisher87: >Or the folks on your end could stop creating keys with symbols in them.

Hah, I don't think I have that much pull ;)

>I'm intrigued by these Pydantic objects and how they could help with this challenge.

Pydantic is super cool! We can use Pydantic models to describe the objects we pass around in the code, and control [serialization](https://docs.pydantic.dev/latest/concepts/serialization/), and also to do automated runtime validation based on type hints!

icepyx/core/granules.py
mfisher87: `dict[Never, Never]` means an empty dict, and this change doesn't pass type checking. In what case would `reqparams={}` be passed?

mfisher87:
Thanks, @JessicaS11 !

icepyx/core/granules.py
mfisher87: Woops, I meant `subsetparams` :)

PR #595, opened 2024-08-30
Replace integration test review trigger with manual trigger

mfisher87:
@JessicaS11 I think we should remove the `pull_request_review` trigger and logic. I don't know what we need to get this right, so I replaced it with a manual trigger. This will help us get PRs moving again. What do you think?

TODO:

- [x] @JessicaS11 remove Integration Tests from workflows required to merge PRs


JessicaS11:
> I don't know what we need to get this right, so I replaced it with a manual trigger. 

I was looking at our setup, and I think the funny business might be coming from our if statement: `if: "${{ !github.event.pull_request.head.repo.fork && (github.event.action != 'pull_request_review' || github.event.review.state == 'approved') }}"`

My plain text of this logic is to run the test if (1) it's not a fork AND (2) it's (a) not a PR review OR (b) it's approved. I'm not sure why condition (a) is in there, since we're using the `pull_request_review` trigger but only want it to run if it's approved (and not on a fork). I know I don't have an intuition yet for saving computation by not evaluating past the first condition, but I can't make the logic of this one work out.

JessicaS11:
@mfisher87 Any further thoughts on this?

mfisher87:
Maybe we should close the PR? Things seem to be working well anyway right now. Can always come back to it :)

JessicaS11:
> Things seem to be working well anyway right now. Can always come back to it :)

Strong disagree. The Integration Tests are being triggered far more often than they should be.

mfisher87:
When are they being triggered that you don't want them triggered? I've been struggling with them not being triggered when I do want them to be triggered. Do you think this manual triggering approach is the way forward?

mfisher87:
We decided today to avoid the rabbit hole of GitHub Actions debugging and disable the integration test merge requirement and stop running integration tests on approval for now. I'm sure we can figure it out eventually :)

JessicaS11 (approved):

JessicaS11 (approved):

PR #598, opened 2024-09-05
Partially typecheck `APIformatting` module

mfisher87:
Add type annotations and some refactors to the APIformatting module.


icepyx/core/APIformatting.py
```diff
@@ -36,10 +36,6 @@ def _fmt_temporal(start, end, key):
 
     assert isinstance(start, dt.datetime)
     assert isinstance(end, dt.datetime)
-    assert key in [
-        "time",
-        "temporal",
-    ], "An invalid time key was submitted for formatting."
```
mfisher87, line 42: This logic (is `key` `"time"` or `"temporal"`?) is repeated just below, so I moved the error statement into that conditional.

In general, assertions should not be used in "the code" (use only in the tests) unless you specifically want the assertions to be controllable at runtime. `python -O` will disable all assertions like `-W` disables warnings. So when I moved this logic I also changed it to raise a `RuntimeError`.

icepyx/core/APIformatting.py
```diff
@@ -282,31 +283,14 @@ def __init__(
         self._fmted_keys = values if values is not None else {}
 
     @property
-    def poss_keys(self):
+    def poss_keys(self) -> dict[str, list[str]]:
```
mfisher87, line 286: This property returns static data, from what I can tell, so I simplified it down, and removed `_poss_keys` attribute.

icepyx/core/APIformatting.py
```diff
@@ -341,8 +325,17 @@ def _get_possible_keys(self) -> dict[str, list[str]]:
                     "Coverage",
                 ],
             }
+        else:
+            raise RuntimeError("Programmer error!")
```
mfisher87, line 329: The typechecker doesn't seem to be good at knowing whether a conditional is "exhaustive", so we're making that explicit here.

JessicaS11: It would be good to have something more explicit (e.g. "exhaustive list error", since that seems to be a typechecker weakness) that makes it more obvious why it go there to begin with.

icepyx/core/APIformatting.py
JessicaS11: I think I had it this way because (1) I'd learned about properties and was probably overusing them; (2) I wanted it to be private enough the user understood they could look but shouldn't touch (so an invalid parameter wasn't submitted). Overall a fan of the simplification though.

JessicaS11 (approved):

icepyx/core/APIformatting.py
mfisher87: Love the idea to go really specific with the name. :100:

icepyx/core/APIformatting.py
mfisher87: >(1) I'd learned about properties and was probably overusing them;

Who can blame you, properties are awesome!

>(2) I wanted it to be private enough the user understood they could look but shouldn't touch

Do you mean "can't set it" by "shouldn't touch"?

icepyx/core/APIformatting.py
mfisher87: ea4a0c5

mfisher87:
Integration tests: https://github.com/icesat2py/icepyx/actions/runs/11041992788

mfisher87:
Not passing. #598 fixes bug introduced on `development` branch from my last-merged PR.

mfisher87:
In the interest of time, I'm going to merge this PR with failing integration tests, having tested the solution locally and gotten this branch to pass. This will simplify #616 into a one-step change and reduce the amount of review cycles. Apologies if this is objectionable! Please let me know if so and I will avoid a repeat in the future.

icepyx/core/APIformatting.py
JessicaS11: > Do you mean "can't set it" by "shouldn't touch"?

I suppose so.

mfisher87: Gotcha. That's already covered by the `@property` decorator :)

```python
>>> class Foo:
...   @property
...   def foo(self):
...     return "foo"
... 
>>> Foo().foo
'foo'
>>> Foo().foo = "bar"
  File "<stdin>", line 1, in <module>
AttributeError: can't set attribute 'foo'
```

PR #601, opened 2024-09-09
bump sangonzal/repository-traffic-action version to latest working one (fixes #596)

JessicaS11:
#596 bumped the version to an invalid version number (v1), causing the action to fail.

### DIFF
--- a/.github/workflows/traffic_action.yml
+++ b/.github/workflows/traffic_action.yml
@@ -23,7 +23,7 @@ jobs:
 
     # Calculates traffic and clones and stores in CSV file
     - name: GitHub traffic
-      uses: sangonzal/repository-traffic-action@v1
+      uses: sangonzal/repository-traffic-action@v0.1.6
       env:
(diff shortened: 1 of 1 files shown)


mfisher87:
Interesting. The `v1` tag exists on that repo, and dependabot's script is calculating it as the latest version, even though it's not.

https://github.com/sangonzal/repository-traffic-action/tags

I wonder if we can get dependabot to stop trying to update this action.

JessicaS11:
>I wonder if we can get dependabot to stop trying to update this action.

Perhaps with an [ignore](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) for that action+version tag combo? Only way it'd be an issue is if they remove the original tag and then reuse it, which seems unlikely.

mfisher87:
> Perhaps with an [ignore](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore) for that action+version tag combo?

Oo nice, TIL!

PR #603, opened 2024-09-09
Add v1.x deprecation warning

### LINKED ISSUE #545: Add deprecation warning to icepyx v1
Notify users that a future change will break icepyx v1 through a deprecation warning. They will need to take action to install v2. They can do this as soon as v2 is available.

See #544 for more info

Should be thrown at import time

mfisher87:
Closes #545


icepyx/__init__.py
```diff
@@ -1,3 +1,15 @@
+from warnings import warn
+
+deprecation_msg = """icepyx v1.x is deprecated; the back-end systems on which is relies
+are being shut down. Please upgrade to icepyx v2.x which uses the new NASA Harmony
+back-end.
```
JessicaS11, line 5: Could we say "is going to be deprecated"? As written, this suggests v2 is already available for upgrading to...
```suggestion
deprecation_msg = """icepyx v1.x is going to be deprecated; the back-end systems on which it relies
are being shut down. Please upgrade to icepyx v2.x which uses the new NASA Harmony
back-end.
```

mfisher87: I agree we should clarify that the 2.x upgrade is not yet available.

However, "is going to be deprecated" is IMO confusing terminology. "Deprecated" colloquially means "scheduled for decommissioning or removal" ([see Python's "deprecations" page for examples](https://docs.python.org/3/deprecations/index.html); all of the things there are already considered "deprecated", but will be removed in future releases). So I would argue that the message "1.x is deprecated" is correct, but the wording around it could be clearer.

mfisher87: What do you think? 08db71d

icepyx/__init__.py
```diff
@@ -1,3 +1,17 @@
...
+deprecation_msg = """icepyx v1.x is deprecated; the back-end systems on which it relies
+will be shut down as of late 2024. At that time, upgrade to icepyx v2.x, which uses the
+new NASA Harmony back-end, will be required. Please see
+<https://icepyx.readthedocs.io/en/latest/user_guide/changelog/v1.3.0.html> for more
+information!
+"""
+# IMPORTANT: This is being done before the other icepyx imports because the imported
+# code changes warning filters. If this is done after the imports, the warning won't
+# work.
+warn(deprecation_msg, DeprecationWarning, stacklevel=2)
```
weiji14, line 12: Sure about using [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) instead of [`PendingDeprecationWarning`](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning)? There's also [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning) that is more intended for end users :upside_down_face:

mfisher87: Good question, not completely sure! RE: PendingDeprecationWarning:

>This class is rarely used as emitting a warning about a possible upcoming deprecation is unusual, and [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) is preferred for already active deprecations.

From PEP387:

>`PendingDeprecationWarning` may be used in special cases where the old and new versions of the API will coexist for many releases

I guess that provides a bit more clarity!

I'm still not sure about `FutureWarning`. Any examples of how other major projects handle similar scenarios?

icepyx/__init__.py
```diff
@@ -1,3 +1,17 @@
+from warnings import warn
+
+deprecation_msg = """icepyx v1.x is deprecated; the back-end systems on which it relies
```
JessicaS11, line 3:
```suggestion
deprecation_msg = """icepyx v1.x is being deprecated; the back-end systems on which it relies
```

Adding the link was a great idea!

JessicaS11 (approved):

mfisher87:
I really wish I had more time to get the integration test workflow all smoothed out. Not sure what's going on :(

JessicaS11 (approved):
Re-reviewing after merges from development to see if it triggers UML+integration tests

PR #604, opened 2024-09-10
Type annotate the temporal module

mfisher87:
> [!IMPORTANT]
>
>  To be merged only after merging #598 and then rebasing

Annotating this module was pretty quick! No typeguards or ignores needed.


icepyx/core/temporal.py
```diff
@@ -150,12 +158,10 @@ def validate_date_range_datestr(date_range, start_time=None, end_time=None):
 
     Parameters
     ----------
-    date_range: list(str)
+    date_range:
         A date range provided in the form of a list of strings
             Strings must be of formats accepted by validate_inputs_temporal.convert_string_to_date().
             List must be of length 2.
-    start_time: string, datetime.time, None
-    end_time:  string, datetime.time, None
```
JessicaS11, line 158: Why remove the `start_time` and `end_time` parameters? Shouldn't we just remove the type?

mfisher87: Since they're part of the function signature, they will be documented automatically;  we only need to mention things in the docstring that aren't already represented in the function signature. One less thing to keep in sync :)

icepyx/core/temporal.py
```diff
@@ -261,14 +271,18 @@ def validate_date_range_date(date_range, start_time=None, end_time=None):
     return _start, _end
 
 
-def validate_date_range_dict(date_range, start_time=None, end_time=None):
+def validate_date_range_dict(
+    date_range: dict[str, dt.datetime],
```
JessicaS11, line 275:
```suggestion
    date_range: dict[str, dt.datetime, dt.date],
```

icepyx/core/query.py
```diff
@@ -254,7 +257,7 @@ def spatial_extent(self):
         return (self._spatial._ext_type, self._spatial._spatial_ext)
 
     @property
-    def dates(self):
+    def dates(self) -> Union[list[str], list[dt.time]]:
```
JessicaS11, line 260:
```suggestion
    def dates(self) -> list[str]:
```

icepyx/core/temporal.py
```diff
@@ -1,12 +1,13 @@
...
-import warnings
-
 """
 Helper functions for validation of dates
 """
 
+import datetime as dt
+from typing import Union
+import warnings
+
```
JessicaS11, line 8: I know this reordering is a ruff thing, but it doesn't make sense in this case because the long comment isn't describing the whole module, only the first portion of it, while the imports apply to the whole thing.

icepyx/core/temporal.py
mfisher87: dates and datetimes are related, so I think we can equivalently annotate `dict[str, dt.date]`:

```python
>>> import datetime as dt
>>> datetime = dt.datetime.now()
>>> date = dt.datetime.now().date()
>>> datetime
datetime.datetime(2024, 9, 25, 12, 47, 38, 861853)
>>> date
datetime.date(2024, 9, 25)
>>> isinstance(datetime, dt.date)
True
>>> isinstance(datetime, dt.datetime)
True
>>> isinstance(date, dt.date)
True
>>> isinstance(date, dt.datetime)
False
```

icepyx/core/query.py
```diff
@@ -279,7 +282,7 @@ def dates(self):
             ]  # could also use self._start.date()
 
     @property
-    def start_time(self):
+    def start_time(self) -> Union[list[str], str]:
```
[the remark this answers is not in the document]
mfisher87: I'm trying not to change any underlying behavior here. This function, along with several others, returns error messages as a list containing a single string:

```python
["No temporal parameters set"]
```

I'm not sure why this is the case, but I don't want to mess anything up that depends on that behavior!

icepyx/core/query.py
mfisher87: Another good catch, thank you!

icepyx/core/temporal.py
mfisher87: Ah, this was my mistake then, I mistook that for a module-level docstring that had gotten moved to the wrong place over time. This makes much more sense. Are you OK with me changing this to a `#` comment? The fact that it was in triple-quoted style is what stimulated my jump to that conclusion!

mfisher87: 7751e26

icepyx/core/temporal.py
mfisher87: d037e11

icepyx/core/query.py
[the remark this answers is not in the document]
JessicaS11: Missed that - thanks!

icepyx/core/temporal.py
JessicaS11: Based on your description, it should be:
`dict[str, Union[str, dt.date]]` since str is also a valid input for the values.

mfisher87: I see, thank you!

mfisher87: 97a5544

icepyx/core/temporal.py
```diff
@@ -261,14 +269,18 @@ def validate_date_range_date(date_range, start_time=None, end_time=None):
     return _start, _end
 
 
-def validate_date_range_dict(date_range, start_time=None, end_time=None):
+def validate_date_range_dict(
+    date_range: dict[str, Union[str, dt.datetime]],
```
JessicaS11, line 273:
```suggestion
    date_range: dict[str, Union[str, dt.date]],
```
Per the previous thread, acceptable input types are str, dt.datetime, and dt.date, so I think using dt.date captures the latter two.

icepyx/core/temporal.py
JessicaS11: https://github.com/icesat2py/icepyx/pull/604/files#r1777677428

icepyx/core/temporal.py
mfisher87: Thank you again, Jessica :bow:

JessicaS11 (approved):
thanks for adding types to this module, @mfisher87!

mfisher87:
Thanks for the thorough review :bow:

JessicaS11:
> I don't understand why the unit tests aren't running here. Re-opening PR to trigger.
#618

I should have noted I was intending to use this as a test case after #619 (and that's why I hadn't closed and reopened the PR!)

PR #605, opened 2024-09-10
Typecheck spatial module

mfisher87:


icepyx/core/spatial.py
```diff
@@ -47,17 +60,25 @@ def geodataframe(extent_type, spatial_extent, file=False, xdateline=None):
     >>> reg_a = ipx.Query('ATL06',[-55, 68, -48, 71],['2019-02-20','2019-02-28'])
     >>> gdf = geodataframe(reg_a.spatial.extent_type, reg_a.spatial.extent)
     >>> gdf.geometry
-    0    POLYGON ((-48 68, -48 71, -55 71, -55 68, -48 ...
+    0    POLYGON ((-48 68, -48 71, -55 71, -55 68, -48 ...))
```
mfisher87, line 63: My editor was negatively impacted by the unclosed parentheses :X

icepyx/core/spatial.py
```diff
@@ -47,17 +60,25 @@ def geodataframe(extent_type, spatial_extent, file=False, xdateline=None):
...
+    # If extent_type is a polygon AND from a file, create a geopandas geodataframe from it
+    # DevGoal: Currently this branch isn't tested...
+    if file is True:
+        if extent_type == "polygon":
+            return gpd.read_file(spatial_extent)
+        else:
+            raise TypeError("When 'file' is True, 'extent_type' must be 'polygon'")
+
+    if isinstance(spatial_extent, str):
+        raise TypeError(f"Expected list of floats, received {spatial_extent=}")
```
mfisher87, line 76: Moved the handling for the file case all to one place

icepyx/core/spatial.py
```diff
@@ -67,17 +88,29 @@ def geodataframe(extent_type, spatial_extent, file=False, xdateline=None):
                 for pair in zip(cartesian_lons, spatial_extent[1::2])
                 for item in pair
             ]
-            bbox = box(*cartesian_spatial_extent)
+            bbox = box(
+                cartesian_spatial_extent[0],
+                cartesian_spatial_extent[1],
+                cartesian_spatial_extent[2],
+                cartesian_spatial_extent[3],
+            )
```
mfisher87, line 96: The typechecker needs to know there are only 4 things being passed, since the 5th argument to `box` must be a `bool`. The typechecker doesn't understand `*cartesian_spatial_extent[0:4]` to mean the same thing unfortunately.

icepyx/core/spatial.py
```diff
@@ -67,17 +88,29 @@ def geodataframe(extent_type, spatial_extent, file=False, xdateline=None):
...
+                spatial_extent[1],
+                spatial_extent[2],
+                spatial_extent[3],
+            )
 
         # TODO: test case that ensures gdf is constructed as expected (correct coords, order, etc.)
-        gdf = gpd.GeoDataFrame(geometry=[bbox], crs="epsg:4326")
+        # HACK: Disabled Pyright due to issue
+        #       https://github.com/geopandas/geopandas/issues/3115
+        return gpd.GeoDataFrame(geometry=[bbox], crs="epsg:4326")  # pyright: ignore[reportCallIssue]
```
mfisher87, line 108: Avoiding use of state where we don't need it!

mfisher87:
Rebased this PR (and #612 and #613)  and fixed merge conflicts!

icepyx/core/spatial.py
```diff
@@ -1,24 +1,34 @@
...
 # DevGoal: need to update the spatial_extent docstring to describe coordinate order for input
 
 
-def geodataframe(extent_type, spatial_extent, file=False, xdateline=None):
+ExtentType = Literal["bounding_box", "polygon"]
+
+
+def geodataframe(
+    extent_type: ExtentType,
+    spatial_extent: Union[str, list[float]],
```
JessicaS11, line 19:
```suggestion
    spatial_extent: Union[str, list[float], Polygon],
```

icepyx/core/spatial.py
```diff
@@ -1,24 +1,34 @@
...
     Return a geodataframe of the spatial extent
 
     Parameters
     ----------
-    extent_type : string
+    extent_type :
         One of 'bounding_box' or 'polygon', indicating what type of input the spatial extent is
 
-    spatial_extent : string or list
+    spatial_extent :
```
JessicaS11, line 31: Need to update this docstring to include Polygon as an input type (since we handle it in the code!)

icepyx/core/spatial.py
```diff
@@ -269,7 +310,12 @@ def validate_polygon_pairs(spatial_extent):
     return "polygon", polygon, None
 
 
-def validate_polygon_list(spatial_extent):
+def validate_polygon_list(
+    spatial_extent: Union[
+        list[Union[float, str]],
```
JessicaS11, line 315:
```suggestion
        list[float],
```
Check me on this please, but I think we're not taking strings for coordinate lists.

icepyx/core/spatial.py
```diff
@@ -364,7 +410,22 @@ def validate_polygon_file(spatial_extent):
 
 
 class Spatial:
-    def __init__(self, spatial_extent, **kwarg):
+    _ext_type: ExtentType
+    _spatial_ext: list[float]
+    _geom_file: Optional[str]
```
JessicaS11, line 415:
```suggestion
    _ext_type: ExtentType
    _geom_file: Optional[str]
    _spatial_ext: list[float]
```
I've been making them abc order - is there a best practice on this?

icepyx/core/spatial.py
```diff
@@ -364,7 +410,22 @@ def validate_polygon_file(spatial_extent):
...
+
+    def __init__(
+        self,
+        spatial_extent: Union[
+            str,  # Filepath
+            list[str],  # Bounding box or polygon
```
JessicaS11, line 421:
```suggestion
```
List of strings isn't an option we handle.

icepyx/core/spatial.py
```diff
@@ -435,34 +496,56 @@ def __init__(self, spatial_extent, **kwarg):
...
                 (
                     self._ext_type,
                     self._spatial_ext,
                     self._geom_file,
-                ) = validate_bounding_box(spatial_extent)
+                ) = validate_bounding_box(
+                    # HACK: Unfortunately, the typechecker can't narrow based on the
+                    # above conditional expressions. Tell the typechecker, "trust us"!
+                    cast(
+                        Union[list[float], NDArray[np.floating]],
```
JessicaS11, line 510:
```suggestion
                    cast(
                        Union[list[float, tuple[float,float]], NDArray[np.floating]],
```
I think I've been inconsistent in how I've added tuples to this typing, but didn't want to miss any!

JessicaS11: Nevermind - missed the len=4 condition!

icepyx/core/spatial.py
```diff
@@ -435,34 +496,56 @@ def __init__(self, spatial_extent, **kwarg):
...
                 (
                     self._ext_type,
                     self._spatial_ext,
                     self._geom_file,
-                ) = validate_polygon_list(spatial_extent)
+                ) = validate_polygon_list(
+                    # HACK: Unfortunately, the typechecker can't narrow based on the
+                    # above conditional expressions. Tell the typechecker, "trust us"!
+                    cast(
+                        Union[list[Union[str, float]], NDArray[np.floating]],
```
JessicaS11, line 545:
```suggestion
                        Union[list[float], NDArray[np.floating]],
```

icepyx/core/spatial.py
```diff
@@ -646,9 +729,12 @@ def fmt_for_CMR(self):
 
             cmr_extent = ",".join(map(str, extent))
 
+        else:
+            raise RuntimeError("Programmer error!")
```
JessicaS11, line 733:
```suggestion
            raise RuntimeError("Programmer error!")
```
Note to change this to the exhaustive options error we created.

icepyx/core/spatial.py
```diff
@@ -674,4 +760,7 @@ def fmt_for_EGI(self):
             egi_extent = gpd.GeoSeries(poly).to_json()
             egi_extent = egi_extent.replace(" ", "")  # remove spaces for API call
 
+        else:
+            raise RuntimeError("Programmer error!")
```
JessicaS11, line 764: as above

JessicaS11:
> Rebased this PR (and https://github.com/icesat2py/icepyx/pull/612 and https://github.com/icesat2py/icepyx/pull/613) and fixed merge conflicts!

Thanks, @mfisher87! This PR is pretty close - a few suggestions where we may need to make type adjustments (would love a second set of eyes on these) and some notes (which I can tackle later but have to run now) for some docstring and error message edits.

trey-stafford:
@JessicaS11 just a heads-up that I am beginning to review and work on the Harmony integration tasks that @mfisher87 started, beginning with this PR. I'll be looking over things here and will try to respond to your comments this afternoon!

icepyx/core/spatial.py
trey-stafford: Just to confirm, should this also accept a `Polygon`?

Assuming that's the case, I think this is what we would want:

```suggestion
    spatial_extent: Union[str, list[float], list[tuple[float, ...]], Polygon],
```

This accepts either a string, a list of floats, a list of tuples containing floats, or a Polygon.

icepyx/core/spatial.py
trey-stafford: The `Spatial` class is typed to accept a list of strings representing a bounding box or polygon. The docstring for that states:

> list of coordinates
>             (stored in a list of strings, list of numerics, list of tuples, OR np.ndarray)

trey-stafford: Below, the docstring for this function also indicates `spatial_extent` takes a list or np.ndarray of strings or numerics.

icepyx/core/spatial.py
trey-stafford: 0ca4485 reorders these attrs. 

In this case, however, I think we want `_spatial_ext` typed as only being a list of floats. Functions like `validate_polygon_list` take the user-supplied `spatial_extent` and convert it into a list of floats for consistency.

icepyx/core/spatial.py
trey-stafford: Ok, I think the docstrings and type annotations here are misleading and I can see some places where the list of strings would not work.

It looks like maybe this was something that was supported in the past, or had been planned but not all-the-way completed? In any case, I think I'll create another branch to address these issues.

trey-stafford: The changes were not as extensive as I thought they might be, so I made a new commit with the change to remove typing of `list[str]`: b035695

icepyx/core/spatial.py
trey-stafford: I'm tackling this incrementally. Added support for `Polygon` input with: 1c1c490

trey-stafford: And this adds support for the `list[tuple[float, float]]` case: 52037e9. I also add unit tests for each of these new cases.

trey-stafford:
Not sure about the failing `codecov/project` check but I have added some additional unit tests to cover branches identified by codecov as not being covered by this PR.

Aside from that issue, this should be ready for re-review!

icepyx/core/spatial.py
JessicaS11: Hmmm... I'm surprised this resulted in so many errors (I haven't gotten fully up to speed on pyright, so I was just trying to make the typing match what the code already did). It should have already been handling Polygons (now circa line 137) and lists (of either floats or tuples, now circa line 140). I wonder if pyright is choking on the fact that which types are valid for `spatial_extent` depends on the input for `extent_type`?

Thanks for adding those tests!

icepyx/core/spatial.py
JessicaS11: > Ok, I think the docstrings and type annotations here are misleading and I can see some places where the list of strings would not work.

This is reminding me that there were a few outstanding updates to the Spatial module docstrings that never got completed by the student who created the module... since this PR is adding the typing I definitely want to make sure we get that right (and was probably not looking at the docstrings when I checked the types)! Thanks for fixing the docstrings to match.

icepyx/tests/unit/test_spatial.py
```diff
@@ -6,6 +6,7 @@
 import pytest
 from shapely.geometry import Polygon
 
+import icepyx
```
JessicaS11, line 9:
```suggestion
import icepyx as ipx
```
For consistency...

icepyx/tests/unit/test_spatial.py
```diff
@@ -498,6 +564,20 @@ def test_bbox_fmt():
     assert obs == exp
 
 
+def test_fmt_for_cmr_fails_unknown_extent_type():
+    bbox = spat.Spatial([-55, 68, -48, 71])
+    bbox._ext_type = "Unknown_user_override"
+    with pytest.raises(icepyx.core.exceptions.ExhaustiveTypeGuardException):
```
JessicaS11, line 570:
```suggestion
    with pytest.raises(ipx.core.exceptions.ExhaustiveTypeGuardException):
```

icepyx/tests/unit/test_spatial.py
```diff
@@ -498,6 +564,20 @@ def test_bbox_fmt():
...
+        bbox.fmt_for_CMR()
+
+
+def test_fmt_for_egi_fails_unknown_extent_type():
+    bbox = spat.Spatial([-55, 68, -48, 71])
+    bbox._ext_type = "Unknown_user_override"
+    with pytest.raises(icepyx.core.exceptions.ExhaustiveTypeGuardException):
```
JessicaS11, line 577:
```suggestion
    with pytest.raises(ipx.core.exceptions.ExhaustiveTypeGuardException):
```

icepyx/core/spatial.py
trey-stafford: Part of the reason was subsequent function calls also being typed without support for Polygons and lists of tuples w/ floats. Not as difficult to resolve as I initially feared when making the change!

Happy to add the tests - I'll try to add more where I can as I'm working through this harmony transition work!

JessicaS11: > Part of the reason was subsequent function calls also being typed without support for Polygons and lists of tuples w/ floats. Not as difficult to resolve as I initially feared when making the change!

Ah - that makes sense (and is certainly far less challenging to deal with). Thanks for doing so!

> Happy to add the tests - I'll try to add more where I can as I'm working through this harmony transition work!

You would be my hero!

I think the only thing still outstanding for this PR is we still need to resolve the duplicate functionality for handling Polygons (lines 85 + 135) and lists (92+140) that was added in your above linked commits. I think the original ordering got confusing when @mfisher87 moved the `if file is True` bit to the top, so it transitioned from handling all `extent_type==bounding_box` and then all `extent_type==polygon` cases to handling `extent_type==polygon` if it's from a file, `extent_type==bounding_box`, and then the rest of the `extent_type==polygon` cases. I don't think it particularly matters what order we do things in, so perhaps moving the rest of the polygon handling above the bounding_box handling is the clearest thing to do?

trey-stafford: > I think the only thing still outstanding for this PR is we still need to resolve the duplicate functionality for handling Polygons (lines 85 + 135) and lists (92+140) that was added in your above linked commits. I think the original ordering got confusing when @mfisher87 moved the if file is True bit to the top, so it transitioned from handling all extent_type==bounding_box and then all extent_type==polygon cases to handling extent_type==polygon if it's from a file, extent_type==bounding_box, and then the rest of the extent_type==polygon cases. I don't think it particularly matters what order we do things in, so perhaps moving the rest of the polygon handling above the bounding_box handling is the clearest thing to do?

I'll start taking a look at this now! I'll be working on getting as much done with icepyx/harmony integration between now and our discussion Thursday as possible. Look forward to chatting then :)

trey-stafford: I think another part of the confusion here is that the `extent_type` and what's allowed by `spatial_extent` are closely tied together. Adding typechecking has found some code paths that are possible, but do not technically work.

For example, the `check_dateline` function:

* If `extent_type` is `"bounding_box"`, then `spatial_extent` must be a list of floats (`list[float]`).
* If `extent_type` is `polygon`, `spatial_extent` must be either a list or a tuple `list | tuple` and assumes that the list/tuple contains pairs of lon/lat.

This function does not support some of the other input types that have been typed on the `geodataframe`. There we have `spatial_extent` accepting a string, a list of floats, a list of tuples, or a `Polygon`. So we need to do some pre-processing to make the input `spatial_extent` match what called functions expect (like `check_dateline`) or we need to update those functions to accept the full list of possible `extent_type`s.

trey-stafford: Ok, here's a commit that refactors the `geodataframe` function to more clearly show what is and is not supported: c89df21. Note that I removed the "redundant" handling I added in previous commits. 

  _Most_ of the code assumes that the `spatial_extent` is a list of floats. 

`spatial_extent` can be a shapely `Polygon` only if the `extent_type` is `"polygon"` and `xdateline` is passed in as a parameter. If `None` is given (the default) for `xdatetline`, the function will fail.

Otherwise `spatial_extent` must be a string  (the file case) or a list of floats. A list of tuples or a list of lists (e.g., `[(lon, lat), (lon, lat)]` is not supported by this code and will require additional effort to support. I'll see what I can do, but may need to put this down in favor of focusing on Harmony adoption soon.

trey-stafford: @JessicaS11 Ok, I added a function that does the conversion to `list[float]` and track that as a separate variable that gets used where a list of floats is  expected. I think things are working as expected now! 

Note that I did have to update some unit tests because test data contained lists of `int`. I converted those to `list[float]`. Maybe we should support `list[float]` and `list[int]`? Currently, the code will fail if a list of ints is passed instead of the list of floats, as expected.

JessicaS11: > I think another part of the confusion here is that the extent_type and what's allowed by spatial_extent are closely tied together. Adding typechecking has found some code paths that are possible, but do not technically work.

Makes sense. I think another [major] part of the issues we're running into (which I didn't realize fully until today) was how much of the context of when the functions are actually called isn't being accounted for. As a result, we're trying to type `geodataframe` according to ways it wasn't intended to be used. Some of the docstrings throughout this module were definitely not-quite-right, but the `geodataframe` one was almost spot on (with the critical caveats you need to interpret it correctly based on your above statement, which is definitely NOT ideal, and it was missing that we handle `shapely.Polygon` types too). For instance, if `extent_type == 'polygon'`, then the only options for `spatial_extent` are string (if it's a filename), list (of int, float, np.int64, if it's a list of coordinates), or a valid Polygon object. The conversion of the list of tuples or list of lists has already happened in `validate_polygon_pairs` and `validate_polygon_list`, which are called during the `__init__`.

> Otherwise spatial_extent must be a string (the file case) or a list of floats. A list of tuples or a list of lists (e.g., [(lon, lat), (lon, lat)] is not supported by this code and will require additional effort to support.

> Ok, I added a function that does the conversion to list[float] and track that as a separate variable that gets used where a list of floats is expected. I think things are working as expected now!

As I alluded to above, at a glance some of these additions are now duplicating the functionality that's in the `validate_...` set of functions, which are set up to handle all of the possible user inputs and turn them into one common format (a list of floats, in the case of a polygon input) that is what would actually be passed into `geodataframe`.

> I'll see what I can do, but may need to put this down in favor of focusing on Harmony adoption soon.

Please don't hesitate to do so. This PR has turned into a way bigger effort than anticipated, and has brought up some interesting discussions more broadly (e.g. related to https://github.com/nsidc/earthaccess/issues/804) about what inputs we/earthaccess should/shouldn't accept and/or check for the user. The spatial module was our first effort towards isolating handling the various input types and returning something uniform that icepyx could latch on to for any further spatial formatting needs. It clearly needs some streamlining and updating (and my hope is to move some revised version of it upstream to earthaccess, which I'm starting conversations about there).

PR #609, opened 2024-09-14
Mark tests which depend on ECS as expected fail; separate integration and unit tests

mfisher87:


JessicaS11 (dismissed):
shouldn't the __init__.py and conftest.py files remain at the icepyx/tests directory level? Otherwise the integration tests won't pick up the auth info from conftest.

mfisher87:
Yes, excellent catch! I thought I had the `conftest.py` in the integration directory, but clearly not! Both types of tests turn out to require fixtures from `conftest.py`. I'd recommend having one `conftest.py` for unit tests and another for integration tests, but didn't want to increase the scope of the PR :) The `__init__.py` seems to not be needed.

b773376

mfisher87:
Integration tests run:

https://github.com/icesat2py/icepyx/actions/runs/11040360996/job/30668167255

The tests marked `xfail` have status `XPASS` (unexpected pass), which is considered OK by default.

JessicaS11:
Something funny happened with the timing and I can't get the unit tests to run and report a status based on https://github.com/icesat2py/icepyx/pull/609/commits/9a9d67a5fb4345b7dfcef1b298cb9dfe2c11eeb0. But they were passing before the UML updates so I'm going to use admin to merge. See #618.

Issue #575, opened 2024-08-20
Delegate full granule query to earthaccess, keep smart subsetting logic in icepyx


weiji14:
To reduce the confusion of ICESat-2 users in choosing whether to use `icepyx` or [`earthaccess`](https://github.com/nsidc/earthaccess) to search for ATLAS products, this issue attempts to discuss a path forward to have proper separation of concerns in both libraries. Talking to @JessicaS11 and others at the 2024 UW Hackweek, we propose to change the 'search' workflow to:

1. For 80% of the users who wants an easy entrpoint -> Delegate full granule search and download to [`earthaccess`](https://github.com/nsidc/earthaccess). Specifically use `earthaccess` to return the Granule IDs from the NSDIC DAAC, and connect/open to that (ideally cloud-hosted) dataset via `fsspec`.
2. For 20% of the users who want to do more advanced search -> Use `earthaccess` to look for the Granule IDs, and `icepyx` (as a plugin/extension) handles the smart `open_subset` logic.
  - a) Transition phase `icepyx=v1.x` - Have the variable subsetting logic remain in `icepyx` only, without integration into `earthaccess`. Users use this interface with a mix of `earthaccess` and `icepyx`
  - b) Breaking phase `icepyx=v2.x` - Allow users to install `icepyx` as an extension of `earthaccess`. Users will interface directly using `earthaccess` <link to relevant earthaccess issue here>.

Taking a stab this week as part of https://github.com/ICESAT-2HackWeek/icepyx/issues/5

mfisher87:
:100:

weiji14:
Implementation-wise, I'm gonna try to increase code coverage on `query.py` and `granules.py` in #581 first, and then refactor `ipx.Query`'s internal logic to use `earthaccess`'s functions/methods to get the granule list. Intent is to try to retain backwards compatibility as much as possible in icepyx=1.x, and then make any breaking changes in icepyx=2.x.

Issue #530, opened 2024-07-03
transition to Harmony API from on-prem EGI


JessicaS11:
NSIDC on-premises subsetting services for ICESat-2 data are currently handled by the EGI system. This system is being sunset in favor of the Harmony subsetter. The timeline for this transition is summer into fall 2024, with the goal to turn off EGI by the end of 2024. icepyx currently submits user data orders to EGI and must therefore be transitioned to use Harmony as well. (On a side note, we also aim to use this transition to set ourselves up for future improved integrations with [earthaccess]() and [SlideRule]() as part of our collective broader efforts to streamline the user experience and the way these tools interact).

Some key notes on how this transition will impact subsetting of ICESat-2 data and icepyx:
- the two-step "search + filter" followed by "order + download with subsetting" workflow will remain unchanged
- the search and filter step to identify granules of interest will still use CMR*
- temporal and spatial subsetting will be available for [data product Levels](https://nsidc.org/data/icesat-2/products) 2-3a; Level 3b data (ATL11, >=ATL14) will have no subsetting options available
- variable subsetting will no longer be available.

*Harmony filters using CMR, so we could combine this into one step, but we still think the [search, find services, order/download workflow is best](https://github.com/nsidc/earthaccess/discussions/611#discussioncomment-9875046)

This parent issue aims to track specific issues required for icepyx to transition to Harmony:

PR #611, opened 2024-09-16
release v1.3.0

JessicaS11:
This is anticipated to be the last release in v1 of icepyx.

NSIDC (the NASA DAAC that hosts ICESat-2 data) is transitioning their API endpoint from the on-premises EGI system to the cloud-based Harmony services. icepyx is being updated to access these new services; the only change users will experience concerns services that are no longer available (e.g. reformatting). (#530)

The developers plan to use this version change to implement additional breaking changes concerning the Query module, transitioning it's functionality to use the earthaccess library instead. (#575)

Find more details on all these changes, including their progress, on GitHub issues and discussions.


mfisher87 (approved):
Looks good to me!

PR #612, opened 2024-09-18
Low hanging refactors

mfisher87:
> [!IMPORTANT]
> To be merged only after merging #605 and rebasing


icepyx/core/is2ref.py
```diff
@@ -20,38 +21,38 @@ def _validate_product(product):
...
-            "ATL15",
-            "ATL16",
-            "ATL17",
-            "ATL19",
-            "ATL20",
-            "ATL21",
-            "ATL23",
-        ], error_msg
-    else:
+    if not isinstance(product, str):
```
mfisher87, line 24: I find the style of:

```python
if not valid:
  raise ...

# main logic here
```

more readable than:

```python
if valid:
  # main logic here
else:
  raise ...

File: icepyx/core/is2ref.py [hunk 2]
```diff
@@ -330,13 +332,13 @@ def gt2spot(gt, sc_orient):
             elif gr_lr == "r":
                 spot = 6
 
-    if "spot" not in locals():
```
mfisher87, line 333: Looking at `locals()` is a smell to me. I feel it may throw off some other code-readers.

icepyx/core/is2ref.py
```diff
@@ -423,30 +425,33 @@ def extract_version(filepath, auth=None):
         f = h5py.File(filepath, "r")
 
     # Read the version information
+    version_str: str
     try:
-        version = f["METADATA"]["DatasetIdentification"].attrs["VersionID"]
+        version = f["METADATA"]["DatasetIdentification"].attrs["VersionID"]  # pyright: ignore[reportIndexIssue]
```
mfisher87, line 430: `h5py` isn't annotated to allow accessing metadata with subscript notation. Just telling the typechecker "it's fine" for now :)

icepyx/core/query.py
```diff
@@ -753,20 +754,18 @@ def granules(self):
...
+                                       71],['2019-02-20','2019-02-28']) #
+        +SKIP
         >>> reg_a.granules # doctest: +SKIP
         <icepyx.core.granules.Granules at [location]>
         """
-
-        if not hasattr(self, "_granules") or self._granules is None:
-            self._granules = Granules()
-
-        return self._granules
```
mfisher87, line 764: Looks to me like `self.granules` is static and could be set at init-time as a class attribute, but then we wouldn't get this awesome docstring. I changed the method to a `cached_property` to, IMO, get the best qualities of both.

icepyx/core/spatial.py
```diff
@@ -698,10 +698,10 @@ def fmt_for_CMR(self) -> str:
 
         """
         # CMR keywords: ['bounding_box', 'polygon']
-        if self._ext_type == "bounding_box":
-            cmr_extent = ",".join(map(str, self._spatial_ext))
+        if self.extent_type == "bounding_box":
+            return ",".join(map(str, self._spatial_ext))
```
mfisher87, line 702: When there's an assignment like this, that tells me that something may be done to this variable after the conditional ends. Instead, `return` tells the reader right away that nothing further will happen.

icepyx/core/is2ref.py
JessicaS11: To confirm my understanding: new line 299 `spot: Optional[int] = None` will leave `spot` and it's value alone if it exists, otherwise create it and set it to `None`?

trey-stafford: `spot: Optional[int] = None` sets `spot` to `None` and annotates it as being optional (can take the value of `None`, which it is initialized to) or `int`. 

https://docs.python.org/3/library/typing.html#typing.Optional

> `Optional[X]` is equivalent to `X | None` (or `Union[X, None]`).

In our case, the equivalent would be `int | None`.

trey-stafford:
Looks like this PR is ready-to-go pending #605

icepyx/core/is2ref.py
JessicaS11: Thanks for the link and explanation. My concern here then is that if there is already a variable named `spot`, we're overwriting it here so it will always be None. We'll need to move this line above the functionality that's trying to fill in the spot value.

trey-stafford: Yes, in this case we initialize `spot` to `None` (line 299), and then try to set it (lines 300-333). If, after that, `spot` is still `None` (line 335), we raise a `ValueError` (line 336). 

It boils down to this:

```
spot = None
<code to attempt setting spot to an integer>
if spot is None:  # `spot` is still `None` , the intervening code failed
    raise ValueError()
```

JessicaS11: Facepalm (and thanks for kindly re-explaining). Apparently when I last looked at this I missed that this conversation is happening a few lines later than the `spot: Optional[int] = None` line and instead thought the setting `spot=None` was happening after our `<code to attempt setting spot to an integer>`. Sorry about that!

pyproject.toml
```diff
@@ -141,7 +141,7 @@ exclude = [
 ignore = [
   "icepyx/quest/*",
   "icepyx/core/auth.py",
-  "icepyx/core/is2ref.py",
+  # "icepyx/core/is2ref.py",
```
JessicaS11, line 144:
```suggestion

```
Looks like we've just removed the other ignores rather than commenting them.

mfisher87: Not sure why I may have done that :laughing: :shrug:

icepyx/core/spatial.py
```diff
@@ -759,10 +759,10 @@ def fmt_for_CMR(self) -> str:
...
 
-        elif self._ext_type == "polygon":
+        elif self.extent_type == "polygon":
```
JessicaS11, line 765: Silly question: isn't this more computationally expensive since it adds another "step" to look at the property?

JessicaS11:
> Looks like this PR is ready-to-go pending https://github.com/icesat2py/icepyx/pull/605

Any idea why this shouldn't go directly into development?

icepyx/core/spatial.py
trey-stafford: I don't believe so - at least not in any significant way. `self.extent_type` just returns `self._ext_type`. I think I did this because it seemed more clear to me. Happy to change it back though!

trey-stafford:
> > Looks like this PR is ready-to-go pending #605
> 
> Any idea why this shouldn't go directly into development?

This work is based on what's in #605 . We might be able to pull out the changes from this PR independently if we decide not to move forward with #605 though.

JessicaS11:
> > > Looks like this PR is ready-to-go pending #605
> > 
> > 
> > Any idea why this shouldn't go directly into development?
> 
> This work is based on what's in #605 . We might be able to pull out the changes from this PR independently if we decide not to move forward with #605 though.

@trey-stafford @mfisher87 if either of you are able to separate out the changes in the PR that rely on #605 (maybe we can add those directly to #605?) from those that don't, it would be super to get this merged in!

mfisher87:
I might be able to take a look in a couple weeks, but can't promise at the moment!

PR #624, opened 2024-10-29
add dependabot ignore for non-semantic v1 of repository-traffic-action

JessicaS11:
Adds ignore to prevent recurrence of #596 with need for #601.


mfisher87 (approved):
Seems right! :)

PR #627, opened 2024-10-31
Add `harmony-py` as a dependency

trey-stafford:
WIP to add harmony-py and remove references to UAT. IS2 collections are now enabled for harmony in prod - we just haven't started advertising this widely yet.


icepyx/core/harmony.py
```diff
@@ -1,13 +1,14 @@
...
 def get_capabilities(concept_id: str) -> dict[str, Any]:
-    response = requests.get(
-        CAPABILITIES_BASE_URL,
-        params={"collectionId": concept_id},
-    )
-    return response.json()
+    capabilities_request = harmony.CapabilitiesRequest(concept_id=concept_id)
+    # TODO: This will work if the user has a .netrc file available but the other
+    # auth options might fail. We might need to add harmony client auth to the
+    # icepyx auth package.
```
JessicaS11, line 10: icepyx.auth is a mixin for earthaccess, so it is easily made available wherever credentials are needed. I poked around harmony-py a bit and it looks like they don't allow you to just pass in a valid session, but I think earthaccess lets you just grab the username and password, so if we have the existing auth object we should be able to authenticate with harmony.Client that way.

icepyx/core/harmony.py
```diff
@@ -1,13 +1,62 @@
...
+        temporal: Union[HarmonyTemporal, None] = None,
+    ) -> str:
+        """Places a Harmony order with the given parameters.
+
+        Return a string representing a job ID.
+
+        TODO/Question: it looks like this code will always use the provided
+        parameters to do subsetting. Are there cases where we just want the data
+        downloaded as whole granules? If so, we may need to use another API to
+        do so?
```
JessicaS11, line 45: Interesting. The current icepyx default is to subset automatically for spatial and temporal (and then users could add any other available subsetting/reformatting options - it's probably best explained in the example notebooks for access and subsetting). In ipx.query, the order/download functions accept `subset=False` (would want to verify I have the kwarg right), so that would probably be the best way to trigger a non-subsetted download. I'm looking at transitioning that functionality over to earthaccess (with the ultimate goal of retiring ipx.Query entirely in favor of earthaccess), so we can work this case in there.

trey-stafford: Thanks for the feedback! I'm planning on using earthaccess to fill the gaps that harmony can't handle (non-subset orders).

LOTS more changes incoming :) Note that you don't need to keep up with this PR. I'm really just "going at it" and learning/breaking a lot as I go. My primary goal right now is assessing how much work it is going to be for us to implement harmony support, and we may need to revisit (or completely re-write) large chunks later.

PR #628, opened 2024-10-31
Remove references to UAT

trey-stafford:
Created this PR to identify UAT-specific code that Matt added. This temporary support for UAT is no longer needed, as IS2 collections are now supported in production harmony.


trey-stafford:
I'm going to go-ahead with merging this into `harmony`. `harmony` is already failing its tests because there is a lot of work-in-progress that needs to be resolved. Subsequent PRs will focus on implementing the harmony support necessary to resolve the failing tests.

PR #652, opened 2025-02-04
Harmony integration

betolink:
This PR will integrate Harmony functionality to the Query class, it breaks the query file into base_query and the query itself. Right now it's a draft and WIP. I still need to implement full CMR parameters in the Harmony query call to be able to use cycles and tracks.


betolink:
I think we can start reviewing the changes, @JessicaS11 @asteiker some tests fail because they need the secrets from the repo to be injected as credentials, once merged into icepyx they should pass, they pass locally if we export EARTHDATA_USERNAME and EARTHDATA_PASSWORD. 

This PR wires Icepyx to harmony, the 2 notebooks I used for reference are the one that subsets data and the one that access the data. I fixed a few things along the way and will write more comments on Monday.

betolink:
I just noticed that we may need to merge Wei Ji's branch, I dropped python 3.9 in favor of 3.10 and his branch goes to 3.11 which is better and I missed some files too.

doc/HARMONY_MIGRATION_NOTES.md
```diff
@@ -165,7 +165,7 @@ See this thread on EOSDIS Slack for more details:
 
 ## Remaining tasks
 
-Remaining tasks for "take2" development incldue:
+Remaining tasks for "take2" development includue:
```
betolink, line 168: haha still a typo.

PR #657, opened 2025-02-21
Harmony integration: the story continues.

betolink:
This PR replaces NSIDC on-prem subsetter with the cloud-based Harmony service for ICESat-2 data. Note: the Harmony trajectory subsetter does not support variable selection for IS2 yet. The rest of the functionality in Icepyx was preserved but this PR when merged and released will contain minor breaking changes.


betolink:
TODO: We need to expose pause, restart and "skip preview" endpoints in the order class.

.github/workflows/unit_test.yml
```diff
@@ -39,9 +39,12 @@ jobs:
           python-version: "${{ matrix.python-version }}"
 
       - name: "Run tests"
+        # env:
+          # EARTHDATA_PASSWORD: "${{ secrets.EARTHDATA_PASSWORD }}"
+          # EARTHDATA_USERNAME: ${{ secrets.EARTHDATA_USERNAME }}
+          # NSIDC_LOGIN: "${{ secrets.EARTHDATA_PASSWORD }}" # remove this
```
JessicaS11, line 45:
```suggestion
```

icepyx/core/is2ref.py
```diff
@@ -12,6 +12,14 @@
...
 
 
+def _get_concept_id(product, version):
+    collections = earthaccess.search_datasets(
+        short_name=product, version=version, cloud_hosted=True
+    )
+    if collections:
+        return collections[0].concept_id()
+
+
```
JessicaS11, line 22: this function is duplicated (it also exists in class Query, line 139). Suggest removing this version.

icepyx/core/orders.py
```diff
@@ -0,0 +1,203 @@
...
+            A list of downloaded granules
+        """
+        path = Path(path)
+        path.mkdir(parents=True, exist_ok=True)
+        if self.type == "subset":
+            return self.harmony_api.download_granules(
+                download_dir=str(path), overwrite=overwrite
+            )
+        else:
+            return earthaccess.download(self.granules, local_path=path)
```
JessicaS11, line 203: aren't these (`download_granules` and `download`) effectively duplicate functions?

icepyx/core/query.py
```diff
@@ -531,6 +136,41 @@ def tracks(self):
...
+            return None
+
+    @property
+    def order_vars(self) -> Union[list[str], None]:
+        """This used to print the list of vasriables for subsetting, Harmony doesn't provide that for IS2 datasets.
+        we do need to implement a class that gets the variables even if it'sm only for listing.
+        """
+        if self.product:
+            return Variables(product=self.product).avail()  # type: ignore[no-any-return]
+        return []
```
JessicaS11, line 155: I'm working on keeping it possible for the user to still see what variables are available for the product, even though they can't subset by them. I'm thinking this is a good place for a deprecation notice that "variable subsetting during order/download" is no longer available, and we should add a new `product_variables` property or something that lets the user just see the list of what would be available before they download. Thoughts?

icepyx/core/orders.py
betolink: they are equivalent, I think of them as aliases because the name of the function in the Query class is `download_granules` so if I user wants to map that to this class it would work the same.

icepyx/core/query.py
betolink: I like `product_variables` or `collection_variables`, agree, renaming the method and removing "order" from the references will make things clearer.

icepyx/core/orders.py
JessicaS11: Currently the user interacts with the icepyx Query class; the future goal is that they will use an earthaccess Results class object (that's currently being created). In both cases, they would not interact directly with the Orders object, so  what is the use case for an `order.download_granules` alias versus mapping `query.download_granules` to `order.download`?

icepyx/core/query.py
```diff
@@ -531,6 +140,54 @@ def tracks(self):
...
+        """
+        Display customization/subsetting options available for this product.
+
+        Parameters
+        ----------
+        dictview : boolean, default False
+            Show the variable portion of the custom options list as a dictionary with key:value
+            pairs representing variable:paths-to-variable rather than as a long list of full
+            variable paths.
+
```
JessicaS11, line 184:
```suggestion
```

JessicaS11: We might be able to grab some of the examples from the docstring of the original function too.

betolink:
Looks like `EARTHDATA_USERNAME` is not a secret in the repo settings? I tried splitting the conftest.py into integration and unit tests versions but for some reason the integration test version is not setting the default user name. Locally works just fine. @JessicaS11  I removed more unused code, I think we could have a workable version tomorrow.

icepyx/core/orders.py
betolink: I think `earthaccess` will use an order's class as well, because it's more intuitive to control orders on its own abstraction. In this case I was switching between the Query class and the DataOrder, and noticed that I had named `download()` in the order, my preference would be that Query would adopt that but it would be another breaking change.

icepyx/core/query.py
```diff
@@ -1,445 +1,56 @@
...
+This module is a work-in-progress and re-implements the core functionality of
+`icepyx.Query` using `earthaccess` and `harmony-py` in preparation for NSIDC's
+decommissioning of the ECS-based ordering system. See the
+`HARMONY_MIGRATION_NOTES` for more details.
+
+The intent is to have the `QueryV2` class replace the existing, v1 `Query`
+class. Once core functionality is implemented, relevant "V2" classes/functions
+can be renamed and the v1 versions removed.
+"""
```
JessicaS11, line 11: notes need updating

icepyx/core/query.py
```diff
@@ -1,445 +1,56 @@
...
+            self._temporal = None  # type: ignore[reportIncompatibleVariableOverride]
 
-        self._version = val.prod_version(is2ref.latest_version(self._prod), version)
+        logging.basicConfig(level=logging.WARNING)
 
-        # build list of available CMR parameters if reducing by cycle or RGT
-        # or a list of explicitly named files (full or partial names)
-        # DevGoal: add file name search to optional queries
+        cycles = kwargs.get("cycles")
+        tracks = kwargs.get("tracks")
```
JessicaS11, line 53: not sure where it's happening during execution, but I found the issue while reviewing the init so I'm noting it here: the search by cycle and tracks is not returning the correct results. It's returning far more granules than it should (and with cycles and tracks that are not specified as inputs).

icepyx/core/query.py
```diff
@@ -449,10 +60,6 @@ def __init__(
                 self._prod, cycles=self.cycles, tracks=self.tracks
             )
 
-        # initialize authentication properties
-        EarthdataAuthMixin.__init__(self)
-
-    # ----------------------------------------------------------------------
```
JessicaS11, line 455: I can't tag the lines specifically, but why do we need these properties (`product`, `product_version`, `__str__`) in both Query and BaseQuery? Could we have them in just one or the other location?

icepyx/core/query.py
```diff
@@ -932,233 +297,218 @@ def avail_granules(self, ids=False, cycles=False, tracks=False, cloud=False):
...
+            return gran_IDs(
                 self.granules.avail,
                 ids=ids,
                 cycles=cycles,
                 tracks=tracks,
                 cloud=cloud,
             )
         else:
-            return granules.info(self.granules.avail)
+            return self.granules.avail
```
JessicaS11, line 312: when I run this in a notebook, I get a list of granules with detailed info about each, rather than the summary previously provided by `avail_granules` (and as noted in the docstring)

icepyx/core/query.py
```diff
@@ -932,233 +297,218 @@ def avail_granules(self, ids=False, cycles=False, tracks=False, cloud=False):
...
+            if status["status"] == "PREVIEW":
+                return self.last_order.resume()
 
-    # DevGoal: put back in the kwargs here so that people can just call download granules with subset=False!
-    def download_granules(
-        self, path, verbose=False, subset=True, restart=False, **kwargs
-    ):  # , extract=False):
+    def order_granules(
+        self, subset: bool = True, skip_preview: bool = False, path: str = "./"
+    ) -> DataOrder:
```
JessicaS11, line 423: to me it is confusing that this function "orders" granules if they're being subset, but "downloads" granules if they're not, and then there's another `download_granules` function that only "downloads" ordered+subset granules (but not the whole granules). Is there a functional reason not to just move the else steps from here to the `download_granules` function and do nothing if a user calls `order_granules` with `subset=False`?

icepyx/core/query.py
```diff
@@ -1,445 +1,56 @@
...
-    ('polygon', [-55.0, 68.0, -55.0, 71.0, -48.0, 71.0, -48.0, 68.0, -55.0, 68.0])
-    Date range ['2019-02-22', '2019-02-28']
-
-    See Also
-    --------
-    GenQuery
-    """
 
+class Query(BaseQuery):
+    _temporal: Union[tp.Temporal, None]
```
JessicaS11, line 37: As I went to update the API docs for this module, I realized I'm still struggling to see how BaseQuery fits into the bigger picture. My current understanding is that it was a useful distinction while we were working on migrating functions to highlight which ones needed to be touched/updated, but is no longer a useful division of functions. Fundamentally, nothing was replaced by "direct" earthaccess calls (e.g. the user still needs to create a `Query` object), and if anything the things in query.py should ultimately be the IS2 specific stuff that builds on the earthaccess GenQuery/BaseQuery-type object that would be created by earthaccess, not icepyx (thinking ahead to the plugin architecture). Thoughts?

icepyx/core/orders.py
JessicaS11: > because it's more intuitive to control orders on its own abstraction

Agreed!

> noticed that I had named download() in the order, my preference would be that Query would adopt that but it would be another breaking change.

Thanks for this explanation - that totally makes sense. Would this be a good case for a deprecation warning? I agree it would be nice if we could not break this fundamental, but if we're pretty certain that `download()` will be the future function name that aligns with earthaccess/plugin architecture, should we use it as an opportunity to suggest users update their code for that eventuality?

icepyx/core/query.py
JessicaS11: see also: https://github.com/icesat2py/icepyx/pull/657#discussion_r2029360744

icepyx/core/query.py
betolink: I de-duplicated these methods, no need to re-implement them if we are not doing anything different.

betolink:
@weiji14 @rwegener2 I wonder if you guys could take a look at this PR, I suspect you both are busy, so this  should not be necessarily an in-depth review, it would be great if  you could install it on an environment and test it with some notebooks you already use. 

```bash
pip install git+https://github.com/icesat2py/icepyx.git@harmony-take2
```

This is a big refactoring to replace the EGI subsetting backend with Harmony. Most of the code should work as I tried not to break anything. I'll ask in CryoCloud if someone would like to also give this branch a try before we merge it, from the NSIDC perspective the sooner the better because is another check on the on-prem decommissioning.  Thanks!!

cc @asteiker

doc/source/example_notebooks/IS2_data_access2-subsetting.ipynb
```diff
@@ -114,10 +106,8 @@
     "user_expressions": []
    },
    "source": [
-    "By default, spatial and temporal subsetting based on your initial inputs is applied to your order unless you specify `subset=False` to `order_granules()` or `download_granules()` (which calls `order_granules` under the hood if you have not already placed your order) functions.\n",
-    "Additional subsetting options must be specified as keyword arguments to the order/download functions.\n",
-    "\n",
-    "Although some file format conversions and reprojections are possible using the `format`, `projection`,and `projection_parameters` keywords, the rest of this tutorial will focus on variable subsetting, which is provided with the `Coverage` keyword."
+    "By default, spatial and temporal subsetting based on your initial inputs is applied to your order unless you specify `subset=False` to the `order_granules()` function or the `download_granules()` function, which calls `order_granules` under the hood if you have not already placed your order.\n",
```
rwegener2, line 109:
```suggestion
    "By default, spatial and temporal subsetting based on your initial inputs is applied to your order. This will be true no matter if you  use the `order_granules()` function or the `download_granules()` function (which calls `.order_granules under()` the hood if you have not already placed your order). If you don't want your order to be spatially subset, you can use the `subset=False` argument in either `.order_granules()` or `.download_granules()`.\n",
```

rwegener2 (commented):
Hi @betolink 👋🏻 I got pulled into something mid-morning that was semi-urgent and ended up taking much more of the day than I would have liked. So, I didn't do so much of a "code review" as an "external run", but the code I ran worked well!

I left one suggestion on the docs. I'm not totally sure I understood what was written, so be sure to check that for accuracy.

This is an exciting new improvement, @betolink! If you want a more thorough review let me know and I can look at this again next week. But if you're in a rush and just need to keep moving you have a "+1" from my sanity check.

doc/source/HARMONY_MIGRATION_NOTES.md
weiji14, line 1: Is this document meant to be user facing? If so, need to add this as an entry under `doc/source/index.rst`.

weiji14 (commented):
A little squeezed this week (between Easter Monday and another public holiday on Friday), so haven't ran through any code, but just some documentation and user-facing API design related comments for now. I'll try and do a deeper review next week or so.

doc/source/HARMONY_MIGRATION_NOTES.md
betolink: No, this was just internal notes for devs, should we include it as a footnote on the migration?

icepyx/core/harmony.py
```diff
@@ -0,0 +1,334 @@
...
+    stop: dt.datetime
+
+
+class HarmonyApi(EarthdataAuthMixin):
+    """
+    A client for interacting with the NASA Harmony API.
+
+    Attributes
+    ----------
+    harmony_client : harmony.Client
```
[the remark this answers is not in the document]
betolink: Good idea! although the Harmony client should not be a public API for Icepyx users.

icepyx/core/query.py
```diff
@@ -1,445 +1,50 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
 
-        # build list of available CMR parameters if reducing by cycle or RGT
-        # or a list of explicitly named files (full or partial names)
-        # DevGoal: add file name search to optional queries
+        if not hasattr(self, "_temporal"):
+            self._temporal = None  # type: ignore[reportIncompatibleVariableOverride]
+        cycles = kwargs.get("cycles")
+        tracks = kwargs.get("tracks")
```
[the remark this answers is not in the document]
betolink: I think it would be simpler if I consolidate the base_query and query like they were before. Working on it...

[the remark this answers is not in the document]
betolink: Alright, I consolidated the files and now there is one Query like before. Has the same signature.

doc/source/user_guide/documentation/variables.rst
```diff
@@ -13,6 +13,17 @@ Constructor
...
 
+Attributes
+----------
+
+.. autosummary::
+   :toctree: ../../_icepyx/
+
+   Variables.path
+   Variables.product
+   Variables.version
```
weiji14, line 24: Could you add docstrings for these three attributes, so that there's a description of them? Right now it shows up empty under https://icepyx--657.org.readthedocs.build/en/657/user_guide/documentation/variables.html

image

Docstrings can be added here:

https://github.com/icesat2py/icepyx/blob/0512d75e73a19b5a9f2ef79e52494d8e181a0c85/icepyx/core/variables.py#L97-L107

and you can copy the description from here:

https://github.com/icesat2py/icepyx/blob/0512d75e73a19b5a9f2ef79e52494d8e181a0c85/icepyx/core/variables.py#L32-L40

icepyx/core/harmony.py
```diff
@@ -0,0 +1,334 @@
...
+        Checks the status of a submitted Harmony job.
+    place_order(...)
+        Submits an order to Harmony and waits for completion.
+    download_granules(download_dir, overwrite)
+        Downloads all granules associated with past orders.
+    """
+
+    def __init__(self):
+        # initialize authentication properties
+        self.ipx_version = ipx_version
```
weiji14, line 51: Prefer to prefix this with an underscore, so that it's treated as a non-public API implementation detail.
```suggestion
        self._ipx_version = _ipx_version
```

icepyx/core/harmony.py
```diff
@@ -0,0 +1,334 @@
+from concurrent.futures import as_completed
+import datetime as dt
+import json
+from pathlib import Path
+import time
+from typing import Any, Dict, TypedDict, Union
+
+from _icepyx_version import version as ipx_version
```
weiji14, line 8:
```suggestion
from _icepyx_version import version as _ipx_version
```

icepyx/core/harmony.py
```diff
@@ -0,0 +1,334 @@
...
+        request = harmony.Request(**params)
+
+        if not request.is_valid():
+            raise RuntimeError(
+                f"Failed to create valid harmony request: {request.error_messages()}"
+            )
+
+        job_id = self.harmony_client.submit(request)
+        label_req = harmony.AddLabelsRequest(
+            labels=["icepyx", f"icepyx-{self.ipx_version}"], job_ids=[job_id]
```
weiji14, line 138:
```suggestion
            labels=["icepyx", f"icepyx-{self._ipx_version}"], job_ids=[job_id]
```

icepyx/core/harmony.py
```diff
@@ -0,0 +1,334 @@
...
+            The ID of the Harmony job to resume.
+
+        Returns
+        -------
+        dict
+            A dictionary containing the order status and related metadata.
+        """
+        status = self.check_order_status(job_id)
+        if status["status"] == "paused" or status["status"] == "previewing":
+            # we cannot skip after the facvt but pausing and resuming the order does the trick
```
weiji14, line 223:
```suggestion
            # we cannot skip after the fact but pausing and resuming the order does the trick
```

icepyx/core/harmony.py
```diff
@@ -0,0 +1,334 @@
...
+    Methods
+    -------
+    get_capabilities(concept_id)
+        Retrieves the capabilities of a given dataset.
+    check_order_status(job_id)
+        Checks the status of a submitted Harmony job.
+    place_order(...)
+        Submits an order to Harmony and waits for completion.
+    download_granules(download_dir, overwrite)
+        Downloads all granules associated with past orders.
```
weiji14, line 46: Can these be removed? I'm seeing duplicated methods documented on readthedocs, e.g. for `get_capaibilities`, there is https://icepyx--657.org.readthedocs.build/en/657/user_guide/documentation/components.html#icepyx.core.harmony.HarmonyApi.get_capabilities and https://icepyx--657.org.readthedocs.build/en/657/user_guide/documentation/components.html#id2?

```suggestion
```

icepyx/core/is2ref.py
weiji14: @betolink, reminder to remove the duplicated `_get_concept_id` function here in `is2ref.py`.

icepyx/core/is2ref.py
```diff
@@ -89,83 +98,15 @@ def about_product(prod):
 
 # DevGoal: use a mock of this output to test later functions, such as displaying options and widgets, etc.
 # options to get customization options for ICESat-2 data (though could be used generally)
+@deprecated(
+    version="1.4.0", reason="order_vars() is going away, use variables() instead"
+)
 def _get_custom_options(session, product, version):
```
weiji14, line 104: Is this deprecation message correct?

icepyx/core/orders.py
```diff
@@ -0,0 +1,201 @@
...
+    resume()
+        Resumes the order if it has been paused due to exceeding granule limits.
+    pause()
+        Pauses the order.
+    status()
+        Retrieves the status of the order.
+    download_granules(path, overwrite=False)
+        Downloads the granules associated with the order.
+    download(path, overwrite=False)
+        Downloads the granules, waiting until they are ready if necessary.
```
weiji14, line 35: Again, can probably remove these docs, to avoid double documentation on readthedocs at https://icepyx--657.org.readthedocs.build/en/657/user_guide/documentation/components.html#icepyx.core.orders.DataOrder
```suggestion
```

icepyx/core/orders.py
```diff
@@ -0,0 +1,201 @@
...
+        </table>
+        """
+        return html
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+    def id(self) -> str:
+        """
+        Get the job ID of the order.
```
weiji14, line 95: Would prefer if we just use `job_id`, because there is a python built-in function called [`id`](https://docs.python.org/3/library/functions.html#id), and want to avoid confusion.

```suggestion
    def job_id(self) -> str:
        """
        Get the job ID of the order.
```

icepyx/tests/integration/test_quest_fix.py
weiji14, line 1: Is it necessary to rename this file to `test_quest_fix.py`? Can we rename back?

icepyx/core/is2ref.py
betolink: I think we can just remove this altogether, it used to be the custom options for EGI, the API we're replacing but it won't work at all anyways.

weiji14: Ok, then just remove this.

weiji14 (approved):
Blindly ran through the 4 changed notebooks, and appears to have no major issues (though I didn't check closely). Thanks again @betolink for taking on this tremendous effort!

If possible, I'd wait for a second review before merging, but up to you.

icepyx/core/harmony.py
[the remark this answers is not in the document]
weiji14: Started a PR for intersphinx at #680, will wait for this PR to be merged before I do more updates (otherwise the merge conflict will be painful to handle).

rwegener2 (approved):
I can't say I understand every detail, but it all looks good to me! Thanks for the effort to make this really significant change @betolink!

betolink:
Looks like this is ready to be merged! thanks to all the people involved @weiji14 @rwegener2 @mfisher87 @trey-stafford @asteiker et. al. 
I've never done an Icepyx release so I'd appreciate help with that, in the meantime I'll put some notes for the changelog. 
Perhaps we can have Icepyx `v2.0.0` next week. Cheers!

mfisher87:
Wow, awesome job Luis! This was an absolute beast of a change.

PR #661, opened 2025-03-07
Add pytest mark to integration tests requiring download

JessicaS11:
We have several open PRs, issues, and conversations around icepyx testing (e.g. #617). Historically, tests were divided as "requiring an EDL login" and "not requiring an EDL login", which was designed to limit requests to NSIDC and subsequent downloads (per their request), especially by CI. These categories loosely, but not completely, aligned with "integration" and "unit" tests. In #609 (and a few related PRs around that time), the tests were renamed to unit and integration but not refactored to reflect the need for limiting how often download tests were run (i.e. not for every commit to a PR).

This PR creates a custom [pytest mark](https://docs.pytest.org/en/stable/how-to/mark.html) `downloads_data` and uses it to flag test functions that require a data download. This way, integration tests can be run on PRs without downloading any data.

Once this PR is finalized, we should be able to adopt #617 for handling PRs from forks to enable testing that requires credentials without downloading any data.

### DIFF
--- a/.github/workflows/download_test.yml
+++ b/.github/workflows/download_test.yml
@@ -1,0 +1,33 @@
+name: "Integration test"
+# NOTE: This runs only integration tests that download data,
+# as denoted by the pytest downloads_data mark.
+
+on:
+  push:
+    branches:
+      - "main"  # Releases
+      - "development"  # Feature PR merges
+
+jobs:
+  test:
+    name: "Download Data Tests"
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - uses: "actions/checkout@v4"
+
+      - uses: "./.github/actions/install-icepyx"
+        with:
+          python-version: "3.12"
+
+      - name: "Run download tests"
+        env:
+          EARTHDATA_PASSWORD: "${{ secrets.EARTHDATA_PASSWORD }}"
+          NSIDC_LOGIN: "${{ secrets.EARTHDATA_PASSWORD }}"
+        run: |
+          pytest icepyx/tests/integration --verbose --cov app -m "downloads_data"
+
+      - name: "Upload coverage report"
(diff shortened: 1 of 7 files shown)


JessicaS11:
@mfisher87 if you're able to review I would super appreciate it! I tried to limit the changes to the integration test action here so that we can merge this to add the fork approval functionality in #617 with [hopefully] minimal conflicts to resolve.

pytest.ini
```diff
@@ -4,3 +4,5 @@ norecursedirs = .git
 python_files = test*.py
 addopts = --cov=./ --doctest-modules
 doctest_optionflags = NORMALIZE_WHITESPACE NUMBER
+markers =
+    downloads_data: test downloads data from Harmony (deselect with '-m "not downloads-data"')
```
mfisher87, line 8:
```suggestion
    downloads_data: test downloads data from Harmony (deselect with '-m "not downloads_data"')
```

.github/workflows/download_test.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+      - uses: "actions/checkout@v4"
+
+      - uses: "./.github/actions/install-icepyx"
+        with:
+          python-version: "3.12"
+
+      - name: "Run download tests"
+        env:
+          EARTHDATA_LOGIN: "icepyx_devteam"
```
mfisher87, line 25: Is this envvar needed? We only pass the other two into the other integration tests.

mfisher87 (approved):
LGTM :)

.github/workflows/download_test.yml
JessicaS11: I think @betolink added it in to the integration action during his harmony integration work, so I included it here figuring then they'd match...

PR #665, opened 2025-03-14
remove pull_request_target trigger for integration tests

### LINKED ISSUE #663: fix duplicate running of integration tests due to multiple triggers
#617 introduced pull_request_target triggers in the integration action to allow running workflows from forks. However, the if statement that should be keeping duplicate runs from being triggered/run does not appear to be working, and to make things more interesting one run passes and the other fails (see e.g. #657).

@chuckwondo @mfisher87 Any idea if the earthaccess action was updated or changed since we introduced this, or is this somehow an isolated issue?

JessicaS11:
See discussion in #663 around removing duplicate CI runs of integration tests.

Closes #663

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -10,10 +10,6 @@ on:
     branches:
       - "main"         # Release PRs
       - "development"  # Feature PRs
-  pull_request_target:
-    branches:
-      - "main"         # Release PRs
-      - "development"  # Feature PRs
   push:
     branches:
       - "main"         # Releases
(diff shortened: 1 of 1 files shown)


chuckwondo (approved):
Fingers crossed! (btw, not sure when this change will properly take effect, as it is often the case that github workflows don't quite work until landing on the repo's default branch)

chuckwondo:
Both pull_request and pull_request_target still triggered the job, so it still ran twice. I don't know if that's because this is one of those weird cases where things behave differently until the changes land on the repo's "default" branch, or if it's the odd relationship between "pull_request" and "pull_request_trigger".

I would _think_ that it may be the latter because the workflow file used by the pull_request_trigger indeed shows the changes you made to it. We might need to put back a simple `if` condition:

```
if: github.event_name != 'pull_request_target'
```

I wouldn't think that would be necessary since you removed the pull_request_trigger section at the top of the file, but I still can't fully decipher what github is doing here, so that might be necessary to avoid the duplicate run.

At least both runs are passing, so I assume you made an env change?

Looks like you might just need to enhance your test coverage.

JessicaS11:
>At least both runs are passing, so I assume you made an env change?

I don't think so, but given the opacity of exactly what's happening for each trigger...

>Looks like you might just need to enhance your test coverage.

Definitely! I think the transition to harmony (and hopefully ongoing efforts to switch as much as we can use earthaccess under the hood) will help, plus a few open PRs from other folks that started implementing some more tests...

Thanks for looking this over - I'm going to merge and see how things shake out once it's on the "default" branch.

JessicaS11:
@contributor-bot add @chuckwondo for review

JessicaS11:
@all-contributors add @chuckwondo for review

PR #677, opened 2025-04-28
Update fsspec parameters for cloud reads

rwegener2:
## What has been built?

This PR sets up `fsspec` to use the cloud optimized parameters recommended in [this whitepaper](https://nsidc.github.io/cloud-optimized-icesat2/). It does not add the h5py parameters due to benchmarking discussed in https://github.com/icesat2py/icepyx/discussions/675. `fsspec` is used for cloud reading (and only cloud reading) so this update should only affect reads to data in s3.

**Merging Note**: I put this PR up for reference and visibility. I know there are other big changes and a v2 release coming soon, but this PR can wait until the appropriate time for merging.

### Approximate timing results

Even on v006 data adding the added fsspec parameter noticeably speeds up data reads from the cloud (~12 times faster).

| Run | Time (1 group ATL03) |
| --- | --- |
| cloud v006 without optimized fsspec params | ~6 minutes |
| cloud v006 with optimized fsspec params | ~30 seconds |

_Timing was done with the jupyter magic `%%timeit` on the 3 lines below: 1) create Read object 2) append variables 3) load data. This does not account for s3 caching._

## How was it done?

The biggest decision, I think, was whether or not to expose this as an option to the user. This PR is the simplest possible implementation, in which the user is not given a choice. Given what a low level change this is I think this makes the most sense, but others should jump in if they would like.

## How can it be tested?

The code below reads a small v006 file from s3.

```python
import icepyx as ipx

s3url_006 = 's3://nsidc-cumulus-prod-protected/ATLAS/ATL03/006/2019/06/13/ATL03_20190613013940_11570313_006_02.h5'

reader = ipx.Read(s3url_006)
reader.vars.append(beam_list=['gt1l'], var_list=['h_ph'])
ds = reader.load()  # <- This call will now use the new parameters
```


icepyx/core/read.py
```diff
@@ -628,7 +628,11 @@ def load(self):
                 # If path is an s3 path create an s3fs filesystem to reference the file
                 # TODO would it be better to be able to generate an s3fs session from the Mixin?
                 s3 = earthaccess.get_s3fs_session(daac="NSIDC")
-                file = s3.open(file, "rb")
+                fsspec_params = {
+                    "cache_type": "blockcache",
+                    "block_size": 8 * 1024 * 1024,
+                }
+                file = s3.open(file, "rb", **fsspec_params)
```
weiji14, line 635: I'm thinking if there's a way to just call [`earthaccess.open`](https://earthaccess.readthedocs.io/en/latest/user-reference/api/api/#earthaccess.api.open) here? We want to slowly delegate most of the granule reading logic to earthaccess (xref https://github.com/icesat2py/icepyx/issues/575), so would be best if we can have `earthaccess` handle the default cache/block sizes somehow. Cc @betolink for thoughts (since you had some ideas at https://github.com/nsidc/earthaccess/discussions/527).

JessicaS11: >We want to slowly delegate most of the granule reading logic to earthaccess (xref https://github.com/icesat2py/icepyx/issues/575), so would be best if we can have earthaccess handle the default cache/block sizes somehow.

Totally agreed - but I think optimized fsspec parameters (or users setting them in the open call) isn't yet implemented in earthaccess, right? If that's the case, maybe we should open an issue within icepyx that tracks the work in earthaccess so that once it's implemented we can update these lines in the Read module, but in the meantime we can take advantage of the faster read speeds made possible by setting the fsspec parameters. Thoughts?

weiji14: Yep, ok with implementing this in `icepyx` first, with an in-line comment linking to the earthaccess issue at https://github.com/icesat2py/icepyx/issues/575 (or another separate issue).

icepyx/core/read.py
```diff
@@ -591,7 +591,11 @@ def load(self):
                 # If path is an s3 path create an s3fs filesystem to reference the file
                 # TODO would it be better to be able to generate an s3fs session from the Mixin?
                 s3 = earthaccess.get_s3fs_session(daac="NSIDC")
-                file = s3.open(file, "rb")
+                fsspec_params = {
```
JessicaS11, line 596:
```suggestion
                # Goal: delegate most of the granule reading logic to earthaccess (xref https://github.com/icesat2py/icepyx/issues/575)
                # See also: https://github.com/icesat2py/icepyx/pull/677/files#r2083622039
                fsspec_params = {
```

PR #681, opened 2025-05-09
Resolve GeoPandas library warnings

emmanuel-ferdman:
## Summary
This small PR resolves the GeoPandas library deprecation warnings:
```python
icepyx/core/spatial.py:480: DeprecationWarning: The 'unary_union' attribute is deprecated, use the 'union_all()' method instead.
````


weiji14 (approved):
Thanks @emmanuel-ferdman! I won't merge this in yet because we'll want to release v2.0.0 first (#685). But I'll try to get this in soon after in v2.0.1.

JessicaS11:
I've bookmarked looking into why the integration tests are failing - my guess is it's nothing to do with this PR.

JessicaS11:
@mfisher87 @betolink @chuckwondo Any idea why there's an issue with the auth (thus causing the integration tests to fail)? The action matches the one in earthaccess, and the integration tests pass (e.g. [yesterday's run](https://github.com/icesat2py/icepyx/actions/runs/15589510879/job/43904604066)). The only difference is this is coming from a fork (and then I'm approving them to run). I noticed that in this case earthaccess is asking for user input of the username, so I'm guessing it's somehow not picking up on the environmental variables that hold the credentials.

mfisher87:
We're still struggling a bit with getting our integration tests workflow working exactly the way we want it. @chuckwondo has been the rockstar on this, and I definitely struggle with holding the mental model of all the different mechanisms in my head. This is, IMO, a consequence of GitHub's design. Dealing with PRs from forks safely and readably is a **huge pain.**

One thing I can see that's different is that the earthaccess workflow uses the ([dangerous](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) :warning: :warning: :warning:) `pull_request_target` trigger. The difference between `pull_request` and `pull_request_target` (among other things) is that the former uses secrets _and code_ from the source branch (i.e. the fork), , and the latter uses secrets _and code_ from the target branch (i.e. `icesat2py/icepyx`)

The "_and code_" part is an important part of GitHub's security model, but also very restrictive; you can't run integration tests on the code from the source branch with secrets from the target branch. This protects us against malicious forks exfiltrating our secrets by writing code on the fork (e.g. hidden in the tests) that emails our secrets to them, or other malicious things.

We're doing something even more dangerous: we're [manually checking out the source branch's code](https://github.com/nsidc/earthaccess/blob/989afc77b342c13847a4f5695ac818a98c5be297/.github/workflows/integration-test.yml#L119-L133), and running tests against that. Remember, because our trigger is `pull_request_target`, they can access our secrets. The only safety mechanism is that a reviewer must manually trigger the action, and their careful review of the submitted code is the only method of detecting malicious code before doing so.

We intended to have a bot comment in the PR with that important context, but we've struggled to get that portion of the action working correctly for some time now. It's very challenging.

A safe alternative would be to instruct all forks to set their own credentials in their fork repository, and run tests with `pull_request` trigger only. I'm leaning more towards this given how much effort it's taken us to get this working correctly. IMO, GitHub should provide safe scaffolding for this use case instead of expecting us to hack it together. It's wild that this is still unsolved.

betolink:
> A safe alternative would be to instruct all forks to set their own credentials in their fork repository

I'm all for this, I think @jhkennedy created a cookie-cutter project that has some nice instructions on how to setup secrets for them. We should do something similar for both icepyx and earthaccess no?

mfisher87:
> We should do something similar for both icepyx and earthaccess no?

Yeah. I like the _idea_ of taking that burden off of contributors, but in practice it's been _way_ harder than I thought. @chuckwondo how do you feel?

chuckwondo:
> > We should do something similar for both icepyx and earthaccess no?
> 
> Yeah. I like the _idea_ of taking that burden off of contributors, but in practice it's been _way_ harder than I thought. @chuckwondo how do you feel?

IIRC, we took the route we currently use in earthaccess precisely because we were unable to get things working by having users leverage their own credentials via env vars in their forks.

The issue comes down to being able to run integration tests in the primary repository for a PR from a fork, when credentials are required because, as @mfisher87 mentioned, if not carefully orchestrated, allowing "untrusted" contributors to trigger tests requiring secrets can create a security hole. On the flip side, we cannot use the user's secrets either because that would be a security risk to the user.

@betolink, I don't know what you're referring to that @jhkennedy has done that you think would do the trick, but I'm all ears if there's an option that I'm unaware of (which could easily be the case given the intricacies of GitHub).

mfisher87:
> IIRC, we took the route we currently use in earthaccess precisely because we were unable to get things working by having users leverage their own credentials via env vars in their forks.

Do you remember why? I seem to have forgotten :laughing: 

>On the flip side, we cannot use the user's secrets either because that would be a security risk to the user.

It's another GitHub security measure? I was positive that `pull_request` would use the users' secrets... surely there's some way we can do that... there _must_ be! ... right? :sweat_smile:

betolink:
oh I just think he put instructions on how to setup secrets for CI in case a user doesn't know how to do it, but the CI pipeline is meant to run in the forks, after reading what you're saying it makes sense that the other way around doesn't work either, (the target repo can't read the forks secret's). That's why I'm like... what about dummy service account with public password(maybe obfuscated) but not pulled from GITHUB ENVs?

jhkennedy:
This might not be a helpful comment, but I generally am of the opinion that integration tests aren't a contributor concern; they are a maintainer concern, so I prefer not to run integration tests on PRs to a staging branch.

SO, a necessary aside. Icepyx (and earthaccess) uses a trunk-based development model where the trunk is the `development` branch here. Contributions accumulate (are staged) in `development` and then periodically released via tagging a commit. So, trunk-based development without continuous delivery.

Integration tests are a barrier to contributions, both due to the level of testing and due to needing access to secrets. And while integration tests must be working for a release, they don't necessarily all need to be passing when features are staged between releases; they just need to be passing before releasing. 

In a trunk-based development model with continuous delivery, every push to `development` would result in a release, so in that case, you do need to run integration tests for PRs to the trunk. 

However, in a trunk-based development model without continuous delivery, you can let the staging process mitigate the need to run integration tests on PRs and ease the barrier to contributions. So I would not run integration tests on PRs and only run them on pushes to `development` with an `onFailure()` condition to notify the maintainers. 

---

As an aside of an aside, GitHub doesn't really support a trunk-based development model without continuous delivery -- ideally, when you've got staged changed, you'd be able to make a "tag/release request" which shows you all the changes from the last tag to HEAD, and allows you to run workflows like in pull requests, and results in a tag when "merged". You can use a service account to do pseudo trunk-based development without continuous delivery by using a "release" branch that only serves to give you that "release request" view, with pushes to the release branch getting immediately tagged and then fast-forwarded by the service account to the staging branch. This is what we do for HyP3 repositories - feature -> develop with only unit tests, develop -> main with unit and integration tests; develop is the trunk and default branch, main is just to facilitate a tag/release request and release reviews.

mfisher87:
I can see the benefits to that approach, but what about the case where the contribution is new/changed integration tests? In that case, the most important feedback before merge is whether those new/changed tests pass!

jhkennedy:
> I can see the benefits to that approach, but what about the case where the contribution is new/changed integration tests? In that case, the most important feedback before merge is whether those new/changed tests pass!

So for a maintainer/contirbutor from a feature branch (i.e. not a fork) you can run the integration tests -- either as a norm or via a `workflow_dispatch` in those cases:
image

From a fork, I think you're in the same boat either way, so it's either:
1. Do a bunch of work to get the integration tests fork-safe
2. maintainers can make an in-org feature branch and open the PR again from the feature branch, where you can further collaborate
3. Manually run the integration tests as part of the review process
4. Don't accept PRs from forks that change the integration tests
5. something else I haven't thought of... 

I think we're learning (1) isn't super viable, (4) isn't really a good option, so it's (2) or (3), with (2) facilitating better transparency and collaboration (and should be less work for the reviewer than (3)). And I can't speak to (5) :rofl:

jhkennedy:
Anyways, this is a pretty strong diversion for a 4-line PR, so I suggest:
* Moving this conversation to its own issue or discussion
* merging @emmanuel-ferdman 's PR
* circling back to the integration tests

JessicaS11:
@all-contributors add @emmanuel-ferdman for code

PR #685, opened 2025-05-16
v2.0.0 release

betolink:
v2.0.0 release for Icepyx with Harmony integration replacing the deprecated EGI subsetter. Details in the CHANGELOG, documentation and related issues in Github.


weiji14:
Don't have permission to merge into protected branch `main` :sweat_smile: Need to wait for someone with higher permissions.

doc/source/user_guide/changelog/v2.0.0.rst
```diff
@@ -0,0 +1,82 @@
...
+
+.. warning::
+    This is major release and contains a few breaking changes. Full details will be
+    described in the documentation.
+
+    NSIDC (the NASA DAAC that hosts ICESat-2 data) is transitioning their
+    API endpoint from the on-premises EGI system to the cloud-based Harmony services.
+    icepyx has been updated to access these new services; Certain capabilities are temporarily going away
+
+    * Data reformatting: users will only get HDF5 files back from the new API.
```
JessicaS11, line 15: what if the product is gridded (so natively a netcdf)? I assume this is referring to the fact that reformatting is not available?

doc/source/user_guide/changelog/v2.0.0.rst
```diff
@@ -0,0 +1,82 @@
...
+Documentation
+^^^^^^^^^^^^^
+
+- Updated notebooks to reflect the new Harmony API usage (#657)
+
+  - The updated notebooks include instructions for making full and subset orders, and show how to use the new Harmony API.
+  - The top level API works the same, users only need to update to v2.0.0 and the code should work (unless using variable subsetting or reformatting)
+  - Notebooks are in the `doc/source/example_notebooks` folder, and are also available on the Readthedocs documentation page.
+- Include quest tile in index.rst (#497)
+Other
```
JessicaS11, line 73:
```suggestion
- Include quest tile in index.rst (#497)

Other
```

doc/source/user_guide/changelog/v2.0.0.rst
weiji14: @betolink?

doc/source/user_guide/changelog/v2.0.0.rst
weiji14: This shows up ok on https://icepyx--685.org.readthedocs.build/en/685/user_guide/changelog/v2.0.0.html

image

so I think ok to leave it so we have a cleaner merge history. Can always update the release notes later.

doc/source/user_guide/changelog/v2.0.0.rst
betolink: so maybe the comment should say "No file reformatting is available for now."

betolink: And this is in "deprecated" because I had the impression that EGI did some reformatting?

weiji14: How about something like this?

```suggestion
    * Data reformatting: Currently reformatting is not supported for any ICESat-2 products, so users will only get the full HDF5 files back from the new Harmony API.
```

betolink: I think we should stay with something along the lines of "no format transcoding is available" because "full HDF5" sounds like not even subsetting is an option. What do you think?

weiji14:
```suggestion
    * Data reformatting: Currently no format transcoding is available for any ICESat-2 products, so users will only get HDF5 files back from the new Harmony API.
```

weiji14: Ah, we can't push these changes because the `development` branch is protected (and I don't think we have permissions to override it). I guess we'll just leave this as is, and update the changelog in a v2.0.1 release.

weiji14: I've updated this sentence on the GitHub release notes at least - https://github.com/icesat2py/icepyx/releases/tag/v2.0.0. Cheers everyone for making this happen!

weiji14:
Note that upload of sdist/wheel to TestPyPI succeeded, but upload to PyPI [failed](https://github.com/icesat2py/icepyx/actions/runs/15200653146/attempts/1#summary-42754052194) due to an attestation file issue. I've manually downloaded the sdist/wheel from TestPyPI (https://test.pypi.org/project/icepyx/2.0.0/#files) and uploaded it to PyPI at https://pypi.org/project/icepyx/2.0.0/#files (hashsums should match).

Also opened a PR at https://github.com/icesat2py/icepyx/pull/687 to fix the publish-to-pypi.yml GitHub Action workflow so this won't happen again.

Issue #690, opened 2025-05-23
HTML Harmony order details link broken


trey-stafford:
The "View Details" link that gets rendered when a `DataOrder` is rendered in a Jupyter Notebook is broken. Clicking on it results in a URL that responds with a 400 status.

The problem is [in this line](https://github.com/icesat2py/icepyx/blob/da1abd46509615147b1d2a7175d739bc5224a834/icepyx/core/orders.py#L60). 

`self.job_id` is a class method of `DataOrder`, and so the resulting string contains it's repr: 

```
link_html = f'<a target="_blank" href="{self.HARMONY_BASE_URL}{self.job_id}">View Details</a>'
print(link_html)
```
```
<a target="_blank" href="https://harmony.earthdata.nasa.gov/workflow-ui/<bound method DataOrder.job_id of DataOrder(job_id=043660ad-15c2-4bd3-81bd-2ad63405b717, type=subset, granules=EarthdataAuth obj with session initialized)>">View Details</a>
```

I think we should make `job_id` a property instead of a method (we already track `self._job_id` as a property - since this isn't computed data, I think we can just drop the class method and replace `self._job_id` with `self.job_id` at init time. 

Alternatively, replace `self.job_id` in the link_html string with `self._job_id` or `self.job_id()`.

Siddharthh39:
can i take it on

Siddharthh39:
Hay, I tried to give a merge request but it tells me to do something with binder and i am pretty new to Contribution, So i don't exactly know how its done, as binder just keeps on loading for me.

trey-stafford:
Thanks @Siddharthh39! I am not very experienced with Binder myself, but I'll take a look at your associated PR, https://github.com/icesat2py/icepyx/pull/691.

trey-stafford:
I realized while reviewing #691 that I mischaracterized the root cause of the problem. I updated the description to reflect that realization.

Originally, I thought that we were missing an `f` to make the href an f-string, but looking at the PR I realized that the whole `link_html` is already a single f-string. The problem is that `self.job_id` is a class method, not a property.

Siddharthh39:
> I realized while reviewing [#691](https://github.com/icesat2py/icepyx/pull/691) that I mischaracterized the root cause of the problem. I updated the description to reflect that realization.
> 
[...]

PR #691, opened 2025-05-25
fix issue #690

Siddharthh39:
- self.HARMONY_BASE_URL.rstrip("/") removes any trailing slash if it exists (to avoid // in the URL).

- Then you explicitly add a / between the base URL and the job_id.
 
- This ensures the final URL looks like:
 https://harmony.example.com/job/12345
 
 if:
 self.HARMONY_BASE_URL = "https://harmony.example.com/"
self.job_id = "abc123"

output:
<a target="_blank" href="https://harmony.example.com/abc123">View Details</a>

### DIFF
--- a/icepyx/core/orders.py
+++ b/icepyx/core/orders.py
@@ -57,7 +57,7 @@ class DataOrder:
     def _repr_html_(self) -> str:
         # Create a link using the <a> tag
         status = self.status()
-        link_html = f'<a target="_blank" href="{self.HARMONY_BASE_URL}{self.job_id}">View Details</a>'
+        link_html = f'<a target="_blank" href="{self.HARMONY_BASE_URL.rstrip("/")}/{self.job_id}">View Details</a>'
         # Create a self-contained HTML table with a single row
         html = f"""
         <table border="1">


Siddharthh39:
Can anyone give me updates on this?

icepyx/core/orders.py
```diff
@@ -57,7 +57,7 @@ def __str__(self) -> str:
     def _repr_html_(self) -> str:
         # Create a link using the <a> tag
         status = self.status()
-        link_html = f'<a target="_blank" href="{self.HARMONY_BASE_URL}{self.job_id}">View Details</a>'
+        link_html = f'<a target="_blank" href="{self.HARMONY_BASE_URL.rstrip("/")}/{self.job_id}">View Details</a>'
```
trey-stafford, line 60: I can see why this change might be a desirable one (do not assume that there's a trailing `/`/ in the `HARMONY_BASE_URL`), but it does not address the underlying issue reported in #690 .

To be fair, I think my initial characterization of #690 was incorrect. I said that there that there was a problem with the generated `href` link because the string was not being interpolated. I see now that it is (`link_html = f'...'`). 

The problem is actually that `self.job_id` is a class method, not a property. I'll update the issue with this new info.

Here's a real-world example that illustrates the issue with your change:

```
link_html = f'<a target="_blank" href="{self.HARMONY_BASE_URL.rstrip("/")}/{self.job_id}">View Details</a>'
print(link_html)
```

```
<a target="_blank" href="https://harmony.earthdata.nasa.gov/workflow-ui/<bound method DataOrder.job_id of DataOrder(job_id=043660ad-15c2-4bd3-81bd-2ad63405b717, type=subset, granules=EarthdataAuth obj with session initialized)>">View Details</a>
```

Or, more clearly:

```
print(self.job_id)
```
```
<bound method DataOrder.job_id of DataOrder(job_id=043660ad-15c2-4bd3-81bd-2ad63405b717, type=subset, granules=EarthdataAuth obj with session initialized)>
```

The solution is to either:

* call `self.job_id()`: `link_html = f'<a target="_blank" href="{self.HARMONY_BASE_URL}{self.job_id()}">View Details</a>'`

or

* Make `self.job_id` a property of the class. I think this makes the most sense, because it's static data. (and we're already tracking `self._job_id` as a property)...

trey-stafford (changes requested):
Requesting changes to address #690 with clarification of what the problem is.

I think it would also be beneficial for us to add a unit test for this method, to ensure that the resulting HTML link looks the way we want.

betolink:
@trey-stafford should we push some commits to this PR?

trey-stafford:
> @trey-stafford should we push some commits to this PR?

It's @Siddharthh39 's PR, if they would like to proceed with addressing the change, that would be great! Since this is a fork, I'm not sure we could push directly to this ourselves? But if we need to pick up the thread, the effort would be minimal, as described above.

Siddharthh39:
@trey-stafford if you guys need any help let me know. I will provide as much as possible 'cuz i am really busy with my office work as well.

bbuzzanga:
Hi, I wanted to check on the status of this. It seems to me that icepyx is currently unusable without it.

And/or if someone could provide instructions on how to order before downloading

JessicaS11:
@Siddharthh39 Just checking in to see if you are going to be able to make the changes requested:
> Make self.job_id a property of the class. 

If not, please let us know so we can address #690 in a separate PR.

@bbuzz31 Thanks for checking in on this. Once we hear from @Siddharthh39, if they are unable to make the changes perhaps @trey-stafford or @betolink may be able to help us get a fix through quickly.

Siddharthh39:
> @Siddharthh39 Just checking in to see if you are going to be able to make the changes requested:
> 
> > Make self.job_id a property of the class.
> 
> If not, please let us know so we can address #690 in a separate PR.
> 
> @bbuzz31 Thanks for checking in on this. Once we hear from @Siddharthh39, if they are unable to make the changes perhaps @trey-stafford or @betolink may be able to help us get a fix through quickly.

Sorry i won't be able to work.

JessicaS11:
@trey-stafford Are you able to address #690 with the solution you suggested here?

JessicaS11:
Closing this PR from a fork; #690 still needs addressing though.

PR #712, opened 2025-10-10
[bug fix] remove incorrect passing of "no orbital parameters set" string to "readable_granule_name"

### LINKED ISSUE #704: Query with Icepyx
This does not work because 'tracks' is causing nothing to return:
```
# Build a query that points to the specific granule (no spatial filter needed)
q = ipx.Query(
    product='ATL06',
    spatial_extent=[(lon, lat) for lon, lat in zip(lons, lats)],
    date_range=['2019-12-02', '2019-12-02'],
    tracks = ['1022'],
    version='006',
) 
```

JessicaS11:
#704 identified a bug introduced during the transition to Harmony (#657) that caused the string "no orbital parameters set" to be passed to CMR as part of the "readable_granule_name", resulting in queries with no results returned. This mitigates that issue and adds a test for using query with the tracks parameter set.


JessicaS11:
@tsnow03 @rohaizharis Are either of you available to review this PR?

PR #713, opened 2025-10-10
get doctests running as part of test actions

JessicaS11:
While working on #712, I noticed that the codecov report showed some lines as untested that should have been tested by the docstring (and also that some of those same doctests should be failing!). I discovered that despite `--doctest-modules` being set in `pytest.ini`, it wasn't running the docstring tests. This PR modifies the pytest.ini file to help it find all the docstring tests and run them (local testing) and the unit_test.yml action to include them in CI testing. It also starts fixing some of the failing docstring tests.

This lack of docstring tests running explains a not-yet-documented mismatch in what `product_summary_info` provides in v2 after #685 versus the short summary it previously showed.


JessicaS11:
The unit tests will continue to fail due to #723. Going to merge so we can get that fix through without the other version update distractions.

PR #715, opened 2025-10-10
Updating earthaccess and fixing a few minor bugs.

### LINKED ISSUE #703: [v2 Harmony] read tutorial needs updating
The [data read in tutorial](https://icepyx.readthedocs.io/en/latest/example_notebooks/IS2_data_read-in.html) still uses `Read.vars` instead of `Read.variables` and needs updating.

betolink:


doc/source/example_notebooks/IS2_cloud_data_access.ipynb
JessicaS11, line 1: second change has a typo (second s): `variabless`

doc/source/tracking/pypistats/downloads.svg
JessicaS11, line 1: Any idea what precommit is changing here (and why it wasn't done in the PR that updated these figures, #716?)

betolink: Seems like the same chart but I don't know why it got regenerated.

doc/source/example_notebooks/IS2_cloud_data_access.ipynb
betolink: good catch, fixed it.

icepyx/core/read.py
```diff
@@ -78,6 +78,8 @@ def _get_track_type_str(grp_path) -> (str, str, str):
 
     import re
 
+    track_str = "Unknown"
```
JessicaS11, line 81: Is this addressing a particular issue, or just a good practice so we have something to return (in which case, should we also add similar lines for `spot_dim_name` and `spot_var_name`?

pyproject.toml
```diff
@@ -1,20 +1,18 @@
 [project]
 name = "icepyx"
 description = "Python tools for obtaining and working with ICESat-2 data"
-license = {file = "LICENSE"}
+license = { file = "LICENSE" }
```
JessicaS11: any idea what's causing the reformatting? I don't know enough about how local settings can cause changes (vs what's in the icepyx precommit config file) to figure out if somehow this file hasn't been linted before or if it was linted on your fork with different settings.

icepyx/core/read.py
betolink: `spot_dim_name` and `spot_var_name` have values for all conditions but `track_str` gets it from a regex expression, which may not find matches.

JessicaS11, line 81:
```suggestion
    track_str = "Unknown"
    spot_dim_name = "Unknown"
    spot_var_name = "Unknown"
```

JessicaS11 (approved):
Thanks for the updates, Luis!

JessicaS11:
@betolink any idea why the integration tests can't authenticate? the permissions check is passing, but then it looks like the authentication itself is failing.

betolink:
I think it's the same as before, the PR is coming from a fork and the action (from my fork) can't read the secrets to run the tests. Last time I copied the work from my fork to a local branch. I think once merged this should work.

JessicaS11:
> I think it's the same as before, the PR is coming from a fork and the action (from my fork) can't read the secrets to run the tests. Last time I copied the work from my fork to a local branch. I think once merged this should work.

Yeah, I wondered about that. But (a) you have write permissions; (b) I was triggering the run manually and the action run log shows that I have permission:
`[Action Query] The user: JessicaS11 permission is admin.`
`[Action Require] The JessicaS11 permission check is true.`

So I would expect that there's access to the secrets. It makes it so confusing.

Issue #720, opened 2025-11-18
Download granule not working


bidhya:
I am trying to run a simple Quickstart as instructed on this first page:
`import icepyx as ipx
query = ipx.Query(
    # Collection short name
    "ATL06",
    # Bounding box
    [-55, 68, -48, 71],
    # Time bounds
    ['2019-02-20','2019-02-28'],
)
query.download_granules('/tmp/icepyx')
`

However, I get the following Attribute error:
`AttributeError                            Traceback (most recent call last)
Cell In[1], line 10
      1 import icepyx as ipx
      2 query = ipx.Query(
      3     # Collection short name
      4     "ATL06",
   (...)      8     ['2019-02-20','2019-02-28'],
      9 )
---> 10 query.download_granules('/tmp/icepyx')

File [~/Github/icepyx/icepyx/core/query.py, in Query.download_granules(self, path, overwrite)
   1022 if hasattr(self, "last_order") is None:
   1023     raise ValueError("No order has been placed yet.")
-> 1024 status = self.last_order.status()
   1025 if status["status"] == "running" or status["status"] == "accepted":
   1026     print(
   1027         (
   1028             "Your harmony job status is still "
   1029             f"{status['status']}. Please continue waiting... this may take a few moments."
   1030         )
   1031     )

AttributeError: 'Query' object has no attribute 'last_order'
`
I followed the installation instruction like this:
`mamba env create --name icepyx-env --channel conda-forge icepyx
mamba activate icepyx-env
mamba update icepyx
`
Current version installed is 2.0.0

I know there is message about harmony job status, but I do not know how to fix this?

If icepyx is not working for a while, do you have alternative suggestions to download icesat-2 data?

JessicaS11:
Hello @bidhya! Great to see you here again, and thanks for the excellent issue post.

There was a bug in v2.0.0 that only recently got fixed (#715), and I've been sick so I only today got it into a released version. v2.0.1 is now available from PyPI (you could try `pip install icepyx` in your virtual environment, and then restart your kernel if you're using JupyterNotebook), but it usually takes a day or two for conda-forge (and thus mamba) to pick up the new release (which is where the installation instructions get the library from). Alternatively, if you re-create your environment after [this page](https://anaconda.org/channels/conda-forge/packages/icepyx/overview) shows v2.0.1 as the latest version, you should be all set (but please let us know if not!)

bidhya:
[...]

Issue #699, opened 2025-06-20
[v2 harmony] download requires explicitly running order function first


JessicaS11:
It seems data cannot be downloaded unless it has explicitly been ordered first (i.e., the download function will not implicitly order the data if the user has not already done so). This is a change in behavior from v1. Options to address this (presuming this is intended behavior - @betolink?) are:
- update the data access tutorial quick start to reflect the change
- modify the download_granules function to order the data if an order does not already exist

betolink:
I think this is related to #688  I was going to fix it along with #691 and we got into an stalled state.

JessicaS11:
realizing this is a duplicate of #688

Issue #666, opened 2025-03-15
Can't download ATL19 / 23 `south_polar` ocean


bbuzz31:
icepyx doesn't seem to let me download for `south_polar` region of ATL19 and ATL23 grids, although it works fine for `mid_latitude` and `north_polar`. 

I'm constructing the query essentially like this:
```
region_a = ipx.Query(ATL23, [-180,-90,180,90], date_range=['2018-10-01', '2025-04-01'])
region_a.subsetparams(Coverage=region_a.order_vars.wanted)

![Image](https://github.com/user-attachments/assets/e35b93c1-507d-45ff-ab56-e73987baf39b)

region_a.download_granules(self.path_reg)


atl19 = 'longitude latitude delta_time_beg delta_time_end gridcntr_lat gridcntr_lon landmask '\
    'dot_avg_albm dot_avg_uncrtn_albm dot_avgcntr dot_avgcntr_uncrtn ice_conc_albm n_segs_albm depth_avg_albm ' \
    'dot_dfw_albm dot_dfw_uncrtn_albm dot_dfwcntr ssb_avg_albm ssb_avgcntr ssb_dfw_albm ssb_dfwcntr'.split()

dct_prod_var_list = {
                'ATL19':
                {'mid_latitude': atl19,
                'north_polar': atl19[4:] + 'ds_grid_y ds_grid_x'.split(),
                'south_polar': atl19[4:] + 'ds_grid_y ds_grid_x'.split(),
                },
}
dct_prod_var_list['ATL23'] = dct_prod_var_list['ATL19'] # i think they're the same

dct_var_list = dct_prod_var_list[self.prod]
for k, v in dct_var_list.items():
    region_a.order_vars.append(var_list=v, keyword_list=[k])
```

I've attached a screen shot of a resulting product that's downloaded.

I've also played around with the parameters to subset, especially since there is some weird behavior... if i pass `delta_time_***` to `north_polar` i get an empty group in the resulting hdf5. the same thing happens for `south_polar` (but at least the group exists then)

Thanks!

JessicaS11:
Hello @bbuzz31! Thanks for sharing your problem and some code. Could you share exactly what error you're getting? Then I can try and recreate the issue later this week.

bbuzz31:
Hi @JessicaS11, sorry i forgot to attach the screen shot. But there is no error; the screen shot just shows the south_polar group is entirely missing in the downloaded product

[...]

PR #726, opened 2025-12-04
fix download function to order granules if needed

### LINKED ISSUE #688: Unintended `AttributeError` when attempting to download granules before an order
Prior to icepyx 2, it was possible to search for data and then immediately call `download_granules`, without calling `order_granules` first.

Now, it is expected that a harmony order be placed first. However, an `AttributeError` is raised instead of a `ValueError` with an informative error message:

```
import datetime as dt

import icepyx as ipx


date_range = (dt.date(2007, 1, 1), dt.date(2024, 10, 28))

lower_left_lon=-49.149
lower_left_lat=69.186
upper_right_lon=-48.949
upper_right_lat=69.238

result = ipx.Query(
    "ATL06",
    [
        lower_left_lon,
        lower_left_lat,
        upper_right_lon,
        upper_right_lat,
    ],
    date_range,
)

result.download_granules("downloaded-data/ATL06/")
```
```
  File "/home/trst2284/code/icepyx/scratch.py", line 24, in <module>
    result.download_granules("downloaded-data/ATL06/")
  File "/home/trst2284/code/icepyx/icepyx/core/query.py", line 1028, in download_granules
    status = self.last_order.status()
             ^^^^^^^^^^^^^^^
AttributeError: 'Query' object has no attribute 'last_order'
```

I think we just need to fix [this line](https://github.com/icesat2py/icepyx/blob/14a1372e049703337f83c56a3553aed6378d97af/icepyx/core/query.py#L1026) to:

```
if not hasattr(self, "last_order")
```

JessicaS11:
Several users (#666, #720) have reported issues with receiving errors when attempting to download data without first running `order_granules()`, a known issue documented in #688 and #699. This PR updates the download function to first order data if the user has not explicitly run the order function, returning to the intended behavior of icepyx in v1.

### DIFF
--- a/icepyx/core/query.py
+++ b/icepyx/core/query.py
@@ -1001,6 +1001,8 @@ class Query(GenQuery, EarthdataAuthMixin):
         self,
         path: Path,
         overwrite: bool = False,
+        subset: bool = True,
+        skip_preview: bool = False,
     ) -> Union[list[str], None]:
         """
         Download the granules for the order, blocking until they are ready if necessary.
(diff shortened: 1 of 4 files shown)


icepyx/core/query.py
```diff
@@ -1019,8 +1030,8 @@ def download_granules(
         """
         # Order granules based on user selections if restart is False and there
         # are no job IDs registered by the harmony API
-        if hasattr(self, "last_order") is None:
-            raise ValueError("No order has been placed yet.")
+        if not hasattr(self, "last_order"):
```
betolink, line 1033: makes sense, I probably used the exception due my lack of familiarity with what most users were expecting here.

betolink (approved):
Looks good to me!

JessicaS11:
Thanks @betolink! The fact that for some reason the status isn't being reported properly aside, the [unit tests are failing](https://github.com/icesat2py/icepyx/actions/runs/19939005125/job/57175406972). It looks like, despite the version number being explicitly passed as "6", version 7 granules are being returned. Any idea why?

PR #736, opened 2025-12-23
Fix version query

betolink:
By default we should use the most recent version for ATL datasets, this PR fixes a bug that will use the first version returned by CMR (which most times is V006).

PR #741, opened 2026-02-11
fix version query v2

### LINKED ISSUE #723: Unable to query ATL13 version='007' granules despite "old version" warning and availability on Earthdata
I am trying to query ATL13 data. When I use the default query (which defaults to V006), I receive a UserWarning stating that I am using an old version of the product. However, when I explicitly set version='007' in the Query object, avail_granules() returns an V006  list, even though V007 data is visible and available on the NSIDC/Earthdata Search website for the same region and time.

To Reproduce Here is a minimal code snippet to reproduce the behavior:

```Python
import icepyx as ipx

# 1. Query with default version (likely V006) -> Triggers Warning
region_v6 = ipx.Query(
    product='ATL13',
    spatial_extent=[-128.0, 41.0, -121.0, 45.0],
    date_range=['2023-12-30', '2023-12-31'], # Dates where V007 definitely exists
)
# Output: UserWarning: You are using an old version of this product

# 2. Query with explicit V007 -> Returns v006 granules
region_v7 = ipx.Query(
    product='ATL13',
    spatial_extent=[-128.0, 41.0, -121.0, 45.0],
    date_range=['2023-12-30', '2023-12-31'],
    version='007'
)

print(f"V007 Available Granules: {region_v7.avail_granules()}")
```
# Output: V007 Available Granules: [V006 list]
Expected behavior Since icepyx warns that V006 is old, and V007 granules exist on the NSIDC website for this timeframe, avail_granules() should return the list of V007 granules when version='007' is specified.

Actual behavior

Running with version='006' triggers: UserWarning: You are using an old version of this product.

Running with version='007' returns an v006 list.

I checked region_v7.CMRparams and noticed that the specific version information might not be correctly resolved or passed to the CMR query parameters.

Environment :

Operating System: Windows 11 (Anaconda)

Python Version: 3.13.9

icepyx Version: 2.0.1

Additional context I have verified on [Earthdata Search](https://search.earthdata.nasa.gov/) that ATL13 V007 granules (e.g., ATL13_20231231044153_01892201_007_01.h5) exist for the specified query parameters. It seems like icepyx might not yet have the updated Collection Concept ID for ATL13 V007 in its lookup tables.

JessicaS11:
#736 resulted in some failed unit tests because an old version was being used to show what happens when no results are returned (predating CMR's modern, more helpful error messages that more explicitly state the problem). By filling in `None` if a valid `concept_id` was not found (e.g. due to an old version being submitted), that value was passed on to CMR, causing it to balk (because an invalid concept_id was submitted). This PR tries to address several interelated issues brought to the surface by this issue (#723) and the fix implemented in #736:
1. Don't use a no-results-found query to determine if too old a version is submitted. Instead, surface an error when the concept_id cannot be found.
2. Update the test stating no results were found, letting CMR provide the error message.
3. Streamline several duplicative lines of code all dedicated to getting the concept id for a given product and version

### DIFF
--- a/icepyx/core/query.py
+++ b/icepyx/core/query.py
@@ -351,7 +351,7 @@ class Query(GenQuery, EarthdataAuthMixin):
         Available data products can be found at: https://nsidc.org/data/icesat-2/data-sets
     version : str, default most recent version
         Product version, given as a 3 digit string.
-        If no version is given, the current version is used. Example: "006"
+        If no version is given, the current version is used. Example: "007"
     cycles : str or list[str], default all available orbital cycles
         Product cycle, given as a 2 digit string, or a list of 2-digit strings.
         If no cycle is given, all available cycles are used. Example: "04"
@@ -376,7 +376,7 @@ class Query(GenQuery, EarthdataAuthMixin):
     >>> reg_a_dates = ['2019-02-20','2019-02-28']
     >>> reg_a = Query('ATL06', reg_a_bbox, reg_a_dates)
     >>> print(reg_a)
-    Product ATL06 v006
+    Product ATL06 v007
     ('bounding_box', [-55.0, 68.0, -48.0, 71.0])
     Date range ['2019-02-20', '2019-02-28']
 
@@ -395,7 +395,7 @@ class Query(GenQuery, EarthdataAuthMixin):
     >>> reg_a_dates = ['2019-02-22','2019-02-28']
     >>> reg_a = Query('ATL06', str(aoi), reg_a_dates)
     >>> print(reg_a)
-    Product ATL06 v006
+    Product ATL06 v007
     ('polygon', [-55.0, 68.0, -55.0, 71.0, -48.0, 71.0, -48.0, 68.0, -55.0, 68.0])
     Date range ['2019-02-22', '2019-02-28']
 
@@ -493,7 +493,8 @@ class Query(GenQuery, EarthdataAuthMixin):
         --------
         >>> reg_a = ipx.Query('ATL06',[-55, 68, -48, 71],['2019-02-20','2019-02-28'])
         >>> reg_a.CMRparams
-        {'temporal': '2019-02-20T00:00:00Z,2019-02-28T23:59:59Z',
+        {'concept_id': 'C2670138092-NSIDC_CPRD',
+        'temporal': '2019-02-20T00:00:00Z,2019-02-28T23:59:59Z',
         'bounding_box': '-55.0,68.0,-48.0,71.0'}
         """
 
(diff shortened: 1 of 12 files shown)


JessicaS11:
@betolink I took an alternate approach to fixing the version submission issue that cleans up some duplicate code and also dealt with the failing test in #736 (that was looking for a no results error from CMR due to submitting an old version). Can you review here, and then update your PR so we can get the test and typing included?

icepyx/core/query.py
```diff
@@ -885,7 +880,7 @@ def _order_subset_granules(self, skip_preview: bool = False) -> str:
                 raise ValueError("No temporal or spatial parameters provided.")
 
         job_id = self.harmony_api.place_order(
-            concept_id=concept_id,
+            concept_id=self.concept_id,
```
betolink, line 883: nice!

icepyx/tests/unit/test_granules.py
```diff
@@ -600,7 +600,7 @@ def test_no_granules_in_search_results():
     ermsg = "Your search returned no results; try different search parameters"
     with pytest.raises(AssertionError, match=ermsg):
         ipx.Query(
-            "ATL06", [-55, 68, -48, 71], ["2019-02-20", "2019-02-20"], version="2"
+            "ATL06", [-55, 68, -54.999, 68.001], ["2019-02-20", "2019-02-20"]
```
betolink, line 603: Why the change in coords?

betolink (commented):
I think this is good to go, before merging, how should we test it? I think I can push the tests from my fork here by hand or you could also copy them over.

icepyx/tests/unit/test_granules.py
JessicaS11: This test was using the `version="2"` to get a "No Results" response from CMR (you may remember we talked about this at our meeting back in January - fixing passing the version correctly was causing `None` to be submitted to CMR and it didn't like it)... so I changed the coordinates to get the "no results" response more legitimately.

JessicaS11:
> I think this is good to go, before merging, how should we test it? I think I can push the tests from my fork here by hand or you could also copy them over.

~~If you could update your fork or push directly to this branch, that would be super! I changed the base of your PR to point to this PR as one route.~~

I generalized your tests to make it easier to update them with new version releases and merged them into this branch. Just need to approve this PR now!

betolink (approved):
I think this is good to go, I don't know why that granule ATL06_20190225121032_09020203_006_02.h5 was not showing up also CMR is having issues today. If this is not urgent we could wait until tomorrow to make sure is not on us.

PR #755, opened 2026-04-21
add _is_gridded attribute to read class

JessicaS11:
Knowing if the ICESat-2 data product is gridded will be used to inform the appropriate read pathways (gridded products can easily and quickly be read into raster-based formats like Xarray DataArrays, whereas non-gridded products are more suited to table-based formats like Pandas DataFrames). This PR makes that information an attribute during the init so it is available throughout processing.


rwegener2 (approved):
Looks great! 🚀

JessicaS11:
After the UML update the Unit and Integration tests are not reporting (see [this comment](https://github.com/icesat2py/icepyx/pull/756#discussion_r3174792964)). Going to merge this since they were passing before the approving review.

PR #754, opened 2026-04-15
add zizmor security GH action

JessicaS11:
Thanks to @mfisher87 for creating this workflow for earthaccess and pointing me towards it to enhance our GitHub action security.

Issue #618, opened 2024-09-26
branch protection tests not triggered after PR review auto-update


JessicaS11:
In #609, a review triggered a UML update (as intended) which added a commit to the PR. This commit did not trigger another round of unit tests (which were previously passing), thus keeping the PR from meeting all required checks (and thus enabling merging). Retriggering the tests on the branch did not result in an updated status because they're still pointing to the commit before the auto-update. We'll have to tweak the unit test triggers to handle this.

mfisher87:
Weird! That should trigger a `pull_request` `synchronize` event and then trigger the unit tests. When this happens, until we figure out why and fix it, we can close and re-open the PR to force a run on the latest commit (that's a `pull_request` `reopen` event).

JessicaS11:
Turns out this isn't a new problem: https://github.com/orgs/community/discussions/25702

From [this post](https://github.com/orgs/community/discussions/25702#discussioncomment-8650283), it sounds like commits coming from actions do not trigger further CI workflows based on tokening (GH is trying to prevent infinitely recursive workflows to the detriment of the situations where one or two recursions are needed!).

From [the add-and-commit action docs](https://github.com/EndBug/add-and-commit):
> When pushing, the action uses the token that the local git repository has been configured with: that means that if you want to change it you'll need to do it in the steps that run before this action. For example: if you set up your repo with [actions/checkout](https://github.com/actions/checkout/) then you have to add the token there.
Changing the token with which the repo is configured can be useful if you want to run CI checks on the commit pushed by this action; anyway, it has to be set up outside of this action.

And [specifically](https://github.com/EndBug/add-and-commit?tab=readme-ov-file#the-commit-from-the-action-is-not-triggering-ci):
> ## The commit from the action is not triggering CI!
> That's because you're checking out the repo using the built-in [GITHUB_TOKEN secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication): GitHub sees that the push event has been triggered by a commit generated by CI, and doesn't run any further checks to avoid unintentional check loops.
> If you're sure that you want the commits generated during CI to trigger other workflow runs, you can checkout the repo using a [Personal Access Token (PAT)](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token): this will make the resulting commit the same as if you made it yourself.
[...]

PR #756, opened 2026-04-22
Zizmor fixes

weiji14:
Follow up to #754 to apply more security related fixes.

Will need to check https://github.com/icesat2py/icepyx/security/code-scanning?query=is%3Aopen+branch%3Adevelopment+tool%3Azizmor (after this PR is merged into the `development` branch) to ensure all issues have been resolved

TODO:
- [x] Fix [dependabot-cooldown](https://docs.zizmor.sh/audits/#dependabot-cooldown)
- [x] Fix [artipacked](https://docs.zizmor.sh/audits/#artipacked)
- [x] Fix [unpinned-uses](https://docs.zizmor.sh/audits/#unpinned-uses)
- [x] Fix [excessive-permissions](https://docs.zizmor.sh/audits/#excessive-permissions)
- [ ] etc

### DIFF
--- a/.github/workflows/binder-badge.yml
+++ b/.github/workflows/binder-badge.yml
@@ -6,6 +6,6 @@ jobs:
   badge:
     runs-on: ubuntu-latest
     steps:
-      - uses: manics/action-binderbadge@main
+      - uses: manics/action-binderbadge@99560b89939071823821e105afc83aaa87927e03 # v3.0.0
         with:
           githubToken: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/binder-badge.yml
+++ b/.github/workflows/binder-badge.yml
@@ -2,9 +2,14 @@ name: AddBinderBadge
 on:
   pull_request_target:
 
+permissions: {}
+
 jobs:
   badge:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+
     steps:
       - uses: manics/action-binderbadge@99560b89939071823821e105afc83aaa87927e03 # v3.0.0
         with:
--- a/.github/workflows/binder-badge.yml
+++ b/.github/workflows/binder-badge.yml
@@ -1,16 +1,40 @@
+# Add Binder Link for testing a branch when a PR comment with `/binder` is made
+# https://mybinder.readthedocs.io/en/latest/howto/gh-actions-badges.html#example-2-comment-with-a-binder-badge-in-response-to-a-comment
 name: AddBinderBadge
-on:
-  pull_request_target:
+on: [issue_comment] # issues and PRs are equivalent in terms of comments for the GitHub API
 
 permissions: {}
 
 jobs:
-  badge:
-    runs-on: ubuntu-latest
+  trigger-chatops:
+    # Make sure the comment is on a PR, and contains the command "/binder"
+    if: (github.event.issue.pull_request != null) && contains(github.event.comment.body, '/binder')
+    runs-on: ubuntu-slim
     permissions:
       pull-requests: write
 
     steps:
-      - uses: manics/action-binderbadge@99560b89939071823821e105afc83aaa87927e03 # v3.0.0
+      # Use the GitHub API to:
+      #  (1) Get the branch name of the PR that has been commented on with "/binder"
+      #  (2) make a comment on the PR with the binder badge
+      - name: comment on PR with Binder link
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
         with:
-          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            // Get the branch name
+            github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.payload.issue.number
+            }).then( (pr) => {
+
+              // use the branch name to make a comment on the PR with a Binder badge
+              var BRANCH_NAME = pr.data.head.ref
+              github.rest.issues.createComment({
+                issue_number: context.payload.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: `[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/${context.repo.owner}/${context.repo.repo}/${BRANCH_NAME}) :point_left: Launch a binder notebook on this branch`
+              })
+            })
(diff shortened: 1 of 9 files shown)


.github/workflows/binder-badge.yml
```diff
@@ -2,10 +2,15 @@ name: AddBinderBadge
 on:
   pull_request_target:
```
weiji14: Secure way would be to use `pull_request` instead of `pull_request_target` trigger. E.g. following https://mybinder.readthedocs.io/en/latest/howto/gh-actions-badges.html#example-2-comment-with-a-binder-badge-in-response-to-a-comment

Yes, slightly more complicated, but I've done it before at https://github.com/pangeo-data/pangeo-docker-images/pull/631. This will require a change in behaviour, in that someone (with the proper permissions) has to write a comment with `/binder` to have the Binder button show up. I can go ahead with this if that's ok?

.github/workflows/unit_test.yml
```diff
@@ -10,16 +10,15 @@ on:
       - "development"
   workflow_run:
     workflows: [Update UML diagrams]
```
weiji14, line 12: This `workflow_run` trigger is apparently [dangerous](https://docs.zizmor.sh/audits/#dangerous-triggers), but also I'm not too sure how to avoid it 🙃. I think the key is that we want the unit tests to be re-ran after the UML diagram commit, need to think of how it can be done more safely.

weiji14 (commented):
Ok, have fixed most issues [except for one](https://docs.zizmor.sh/audits/#dangerous-triggers) that is more complicated. See below.

.github/workflows/binder-badge.yml
JessicaS11: This sounds like a good approach to me. I'm not sure how often folks use the binder button anyway, so I don't think explicitly needing to comment to ask for it is a bad idea.

.github/workflows/unit_test.yml
JessicaS11: The reason we need to run the unit tests after the UML diagram commit is because the unit tests passing are required branch protection checks for merging. Is there a better approach to getting the UML diagrams updated that makes it so we don't need this workflow trigger to rerun the unit tests? Otherwise I'm also not sure how to avoid this trigger.

.github/workflows/binder-badge.yml
weiji14: Ok, have modified the workflow in commit 4dcca1cbd46ed3202b0726f239bcd4b3a3f076e5, so someone will need to write `/binder` to place the Binder badge. Let me know if you want this to be documented somewhere (I was thinking of putting it in the Pull Request template, but there isn't any yet).

.github/workflows/unit_test.yml
weiji14: Ok, so reading your previous comment at https://github.com/icesat2py/icepyx/issues/618#issuecomment-2379684365 which links to https://github.com/EndBug/add-and-commit/tree/v10.0.0#the-commit-from-the-action-is-not-triggering-ci, they say:

> If you're sure that you want the commits generated during CI to trigger other workflow runs, you can checkout the repo using a [Personal Access Token (PAT)](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token): this will make the resulting commit the same as if you made it yourself.

So I could make sure the commit happens with a Personal Access Token (PAT) instead of the default GITHUB_TOKEN, and then we should be able to remove the `workflow_run` trigger. I had a look around and there seems to be one already used by the traffic action here:

https://github.com/icesat2py/icepyx/blob/31e208a51672327f8c51733d354d036f4344561a/.github/workflows/traffic_action.yml#L27-L28

So we could probably just re-use that token (if that's ok, or we can create another one). How does that sound?

weiji14: > So we could probably just re-use that token (if that's ok, or we can create another one). How does that sound?

By the way, if deciding to create a new token, I'd recommend adding the secret PAT to a dedicated environment (see: [creating an enviroment](https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments#creating-an-environment)) so that it can only be accessed in a restricted way. Downside is that a maintainer will need to manually approve the UML workflow run (in addition to approving the PR itself). Up to you on what your risk tolerance is (we'll actually be fairly safe already after this PR's changes).

JessicaS11: > Downside is that a maintainer will need to manually approve the UML workflow run (in addition to approving the PR itself)

I'm wondering if a better solution is to rethink when/how we update the UML diagrams, particularly if we're going to be adding some type of manual approval anyway. It seems to be a repeat pain point (for instance, @rwegener2 approved a PR earlier today, it updated the UML diagrams, and the unit tests ran on [31e208a](https://github.com/icesat2py/icepyx/commit/31e208a51672327f8c51733d354d036f4344561a) (#751) instead of #755, so the latter is just hanging waiting for results from the action. I'm thinking we could update the UML diagrams as part of the release process instead (but the option is always available to run it sooner if we want). Maybe make it a hook like binder is now?

.github/workflows/binder-badge.yml
JessicaS11: testing /binder

> Let me know if you want this to be documented somewhere (I was thinking of putting it in the Pull Request template, but there isn't any yet).

Agreed. I will try to add a PR template next week, since I think you're right that's the best spot for it.

.github/workflows/unit_test.yml
weiji14: > it updated the UML diagrams, and the unit tests ran on [31e208a](https://github.com/icesat2py/icepyx/commit/31e208a51672327f8c51733d354d036f4344561a) ([#751](https://github.com/icesat2py/icepyx/pull/751)) instead of [#755](https://github.com/icesat2py/icepyx/pull/755), so the latter is just hanging waiting for results from the action.

Got a little confused by this. I checked and the UML commit was made (https://github.com/icesat2py/icepyx/pull/755/commits/21da432e4f9f9e893ceaa26f03d72ece8da2f268), but you're saying it just retriggered the unit tests on the `development` branch rather than the Pull Request branch? That doesn't sound correct...

> I'm wondering if a better solution is to rethink when/how we update the UML diagrams, particularly if we're going to be adding some type of manual approval anyway ... I'm thinking we could update the UML diagrams as part of the release process instead (but the option is always available to run it sooner if we want). Maybe make it a hook like binder is now?

Making it part of the release process sounds reasonable. I can certainly turn it into a manual hook, i.e. someone writes `/update-uml` to trigger the diagram refresh. Shall we go ahead with that?

weiji14:
@JessicaS11, let's maybe get most of these fixes in, and we can handle the UML workflow related changes (https://github.com/icesat2py/icepyx/pull/756#discussion_r3175379720) separately?

JessicaS11 (approved):
Thanks so much for making this security improvements, @weiji14!

